PR #1, opened 2017-03-20
Aec connect

pwdonh:
Hi Francois, this is implementing amplitude envelope correlation (AEC) as an additional connectivity metric in BST. It's very similar to the method used in e.g. 

Hipp, Joerg F., et al. "Large-scale cortical correlation structure of spontaneous oscillatory activity." Nature neuroscience 15.6 (2012): 884-890.

It computes the correlation between the band-limited envelopes, and the user has the option to orthogonalize the source time-series before computing the envelopes. This is a way of excluding zero-phase interactions that could be due to fieldspread.

I thought it would be useful for the main distribution, let me know what you think. I have some example scripts here: https://github.com/pwdonh/bst_scripts

### DIFF
--- a/toolbox/connectivity/bst_connectivity.m
+++ b/toolbox/connectivity/bst_connectivity.m
@@ -428,27 +428,27 @@ for iFile = 1:length(FilesA)
             nFreqBands = size(OPTIONS.Freqs, 1);
             BandBounds = process_tf_bands('GetBounds', OPTIONS.Freqs);
 
-            % Intitialize returned matrix
+            % Initialize returned matrix
             R = zeros(size(sInputA.Data,1), size(sInputB.Data,1), nFreqBands);
             % Loop on each frequency band
             for iBand = 1:nFreqBands
                 % Band-pass filter in one frequency band + Apply Hilbert transform
+                DataAband = process_bandpass('Compute', sInputA.Data, sfreq, BandBounds(iBand,1), BandBounds(iBand,2), 'bst-fft-fir', OPTIONS.isMirror);
+                HA = hilbert_fcn(DataAband')';                
                 if isConnNN
-                    DataAband = process_bandpass('Compute', sInputA.Data, sfreq, BandBounds(iBand,1), BandBounds(iBand,2), 'bst-fft-fir', OPTIONS.isMirror);
-                    HA = hilbert_fcn(DataAband')';
                     HB = HA;
                 else
-                    DataAband = process_bandpass('Compute', sInputA.Data, sfreq, BandBounds(iBand,1), BandBounds(iBand,2), 'bst-fft-fir', OPTIONS.isMirror);
                     DataBband = process_bandpass('Compute', sInputB.Data, sfreq, BandBounds(iBand,1), BandBounds(iBand,2), 'bst-fft-fir', OPTIONS.isMirror);
-                    HA = hilbert_fcn(DataAband')';
                     HB = hilbert_fcn(DataBband')';
                 end
                 if OPTIONS.isOrth
                     if isConnNN
                         for iSeed = 1:size(HA,1)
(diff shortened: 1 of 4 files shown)


toolbox/connectivity/bst_connectivity.m
```diff
@@ -420,6 +420,64 @@
...
             
+        % ==== AEC ====
+        case 'aec'
+            bst_progress('text', sprintf('Calculating: AEC [%dx%d]...', size(sInputA.Data,1), size(sInputB.Data,1)));
+            Comment = 'AEC: ';
+            % Get frequency bands
+            nFreqBands = size(OPTIONS.Freqs, 1);
+            BandBounds = process_tf_bands('GetBounds', OPTIONS.Freqs);
+
+            % Intitialize returned matrix
```
martcous: Typo: Initialize*

toolbox/connectivity/bst_connectivity.m
```diff
@@ -420,6 +420,64 @@
...
+            R = zeros(size(sInputA.Data,1), size(sInputB.Data,1), nFreqBands);
+            % Loop on each frequency band
+            for iBand = 1:nFreqBands
+                % Band-pass filter in one frequency band + Apply Hilbert transform
+                if isConnNN
+                    DataAband = process_bandpass('Compute', sInputA.Data, sfreq, BandBounds(iBand,1), BandBounds(iBand,2), 'bst-fft-fir', OPTIONS.isMirror);
+                    HA = hilbert_fcn(DataAband')';
+                    HB = HA;
+                else
+                    DataAband = process_bandpass('Compute', sInputA.Data, sfreq, BandBounds(iBand,1), BandBounds(iBand,2), 'bst-fft-fir', OPTIONS.isMirror);
```
martcous: It seems this same line is executed in both cases your condition, so it should can be executed first outside the condition to avoid repetition.

toolbox/connectivity/bst_connectivity.m
```diff
@@ -420,6 +420,64 @@
...
+                    DataBband = process_bandpass('Compute', sInputB.Data, sfreq, BandBounds(iBand,1), BandBounds(iBand,2), 'bst-fft-fir', OPTIONS.isMirror);
+                    HA = hilbert_fcn(DataAband')';
```
martcous: (Same comment for this line)

toolbox/connectivity/bst_connectivity.m
```diff
@@ -420,6 +420,64 @@
...
+                    HB = hilbert_fcn(DataBband')';
+                end
+                if OPTIONS.isOrth
+                    if isConnNN
+                        for iSeed = 1:size(HA,1)
+                            % Orthogonalize complex coefficients, based on Hipp et al. 2012
+                            HBo = imag(bsxfun(@times, HB, conj(HA(iSeed,:))./abs(HA(iSeed,:))));
+%                             HBos = real(HBo .* ((1i*HA)./abs(HA))); % for generating the time-series
```
martcous: Why is this line commented?

pwdonh: This is generating the orthogonalized time-series. It's not necessary to compute it explicitly, since only the amplitude component (HBo) is needed. I left it, since it might be helpful for understanding, in case someone looks at the code.

ftadel:
Looks good, thanks!
I haven't tested it  because I don't have much time now, but it doesn't impact much the rest of the software anyway.  

@peter you can put your scripts here if you want:
https://github.com/brainstorm-tools/bst-users

ftadel:
@pwdonh 
Actually, I just tried to run it and realized that "corr" is not a standard Matlab function.
Is it in a toolbox or in a function you have on your computer?
Can you post an alternative that would work on a regular Matlab install (with no additional toolbox)

Thanks
Francois

PR #2, opened 2017-03-27
use bst_corrn() instead of corr()

pwdonh:
true, corr() part of the statistics toolbox. This gives the same results.


ftadel:
Perfect, thanks

PR #3, opened 2017-04-11
Edf+ Export with Annotations Support

martcous:
So this has been a pretty cool side project that I think is now ready to review. It works fine on my end with a few test files, and the created files can be opened with EDFbrowser. Please test on your end and also don't hesitate to comment on the coding style, I'm still getting used to serious development in Matlab.

PS: Not sure why it reports changes in the copyright header of each modified files, I did not touch them.


ftadel:
Looks perfect!

About the copyright notices, it has to do with the termination of the lines (\n vs. \r\n). 
I have to fix my deployment scripts...

ftadel:
Well, actually I tested with a few more files, and the way you determine reclen in out_fopen_edf doesn't work well on many combinations of file duration/sampling frequency.
I think you could simply enforce blocks of 1s, padding the last record with zeros.

For out_fopen_edf and out_fwrite_edf: you can commit directly on the master branch if you want, I won't be touching these files. All the other changes are correct.

martcous:
The problem is that we have a maximum record size of 61440 bytes, otherwise strict viewers such as a EDFbrowser won't read the file, so if we have a crazy high sampling rate it might cause issues. Can you describe the issue you're having?

ftadel:
Download the file sample_epilepsy.zip, corresponding to the EEG/epilepsy tutorial:
http://neuroimage.usc.edu/brainstorm/Tutorials/Epilepsy

Two bugs:
1) Create the link to the raw file, and try to export it as an EDF file, you get an error in out_fwrite_edf.m line 55.
2) Import 10s of it (right-click > Import), then try to export this imported file, you get an error out_fwrite.m line 42

PR #4, opened 2017-04-13
Bugfix for EDF+ export (multi-epoched events + fractional records)

martcous:
I do not have access to the master so I'm creating a new PR. I get now what you meant with the bug on record length, I now allow records not completely filled and pad the last record with zero like you suggested. Multi-epoched events were also not supported so I fixed that.

It seems that for some situations, out_fwrite_edf() is called multiple times for the same file. That causes issues with my code because I need to save the nextEvent (annotation) but that information is lost at the next call. I cannot save it in the file header because a struct is passed by value in Matlab. I didn't feel like making major changes to out_fwrite() by adding a return value or something, so I just used a Matlab global variable. Maybe there's a way to set a "Brainstorm" global variable. If so, let me know, I'll do it this way.

### DIFF
--- a/toolbox/io/out_fopen_edf.m
+++ b/toolbox/io/out_fopen_edf.m
@@ -41,16 +41,14 @@ header.starttime  = datestr(date, 'HH.MM.SS');
 header.nsignal    = length(ChannelMat.Channel);
 header.nrec       = (sFileIn.prop.samples(2) - sFileIn.prop.samples(1) + 1) / EpochSize;
 
-% We need to choose a record length that produces a whole number of records
-% and a record size less than 61440 bytes (at 2 bytes per sample).
+% We need to choose a record length that produces a  record size less than
+% 61440 bytes (at 2 bytes per sample).
 header.reclen     = 1.0;
-epsilon           = 1e-8;
 for i = 1:10
-    remainder     = mod(header.nrec / header.reclen, 1);
     recordSize    = header.reclen * EpochSize * header.nsignal * 2;
     
-    if (remainder > epsilon && abs(remainder - 1) > epsilon) || recordSize > 61440
-        header.reclen = header.reclen / 10;
+    if recordSize > 61440
+        header.reclen = header.reclen / 2;
     else
         break;
     end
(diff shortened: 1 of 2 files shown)


ftadel:
The function export_data.m loops on short file blocks to be able to process very large files without memory issues. This is why out_fwrite is called multiple times.

Global variables should be avoided whenever possible. A more elegant way of dealing with this issue would be to make the duration of the blocks processed in this loop is equal to the duration of the EDF record. This way out_fwrite would be called only once per EDF record, and you should not need any additional information (global variable or returned value) at each call to out_fwrite_edf.

You would need to edit the function GetDefaultEpochSize in bst_process.m, called at line 210 of export_data.m.

What do you think?

martcous:
If you don't think that will cause memory problems, fine with me. That does seem like the easiest solution without using global variables. Otherwise, I could precompute the annotations in out_fopen_edf() but that is complicated because I need a way to know how many blocks the file will be split in ahead of time.

ftadel:
You can generate the full Annotations channel in out_fopen_edf (and save it as a field of sFile.header), or the one for the current record in out_fopen_edf. 
But in both cases, I think you need to edit the function GetDefaultEpochSize, so that you can define exactly how large you want each processing block to be. You can move your code that defined the record length in this function, in a new case for the EDF format.

martcous:
OK cool yes it's doable, maybe even without touching GetDefaultEpochSize(). In any case, I'll try some things and get back to you.

martcous:
That should do it -- removed the need for a global variable by tying annotations to a specific sample. If the sample is inside the block to be written, the annotation is also written.

ftadel:
Are you ready to try merging this?
To test your new code: try converting various formats to EDF+ and opening them in EDFBrowser and Cartool. You can get some examples in the .zip file I indicated previously.
When you're ready, I think you should be able to merge it by yourself now (I sent you an invitation to be a collaborator on the project, let me know if you need more privileges than this)

martcous:
So I've been testing this more thoroughly. One problem is that we need to choose a unit in the EDF header (V, uV, etc) and the samples need to be in int16. Right now I'm forcing uV but when you have values lower than 0.5 uV, we lose information. I'd like to look at the range of values to choose an appropriate unit (e.g. nV when uV doesn't cut it), but it doesn't seem we have access to the data (F) matrix in out_fopen_edf(). Is there a way to access it? Or do we leave it like that with some information loss?

ftadel:
The values are not directly the int16 values saved in the file.
These integer values are multiplied by a gain value defined in the file with 4 parameters: physical_min, physical_max, digital_min, digital_max. See the corresponding code:
- in_fopen_edf, line 128
- in_fread_edf, line 206

I think most files use default values, that correspond to a dynamic range that fits most EEG recordings. Check on the few EDF files in the examples on Jupiter.
You would have to change these values if you save MEG recordings in EDF, because the units are completely different. Most file formats contain directly gain values that you can reuse. To see where they are in the header structure of a given format, you can check the corresponding in_fread_... function

Does it make sense?

martcous:
Yes it makes sense, but does not really answer my question. How can I figure out the right unit and physical/digital min/max to use in out_fopen_edf() if I don't have access to the data (F) at this point? The gain will only be there in the header if the file was originally imported as an EDF (or another format with a gain as you mentioned, which is not all of them). We can always talk about it in person next week.

ftadel:
Most file formats have a gain parameter, that can be found in the original header. When this is available, it could be better to reuse it. 
But this is not a trivial question. We'll discuss this next week.

For now, you can try using standard values that work well for most EEG recordings. You could already pull this modification to the master branch, it would be better than the current version.

martcous:
Agreed, thanks!

PR #6, opened 2017-04-21
(WIP) Added option to export figure as SVG

martcous:
It was requested to add an option to export Brainstorm figures in SVG. By just saving the plot as SVG in Matlab, it somehow saves the whole thing as a bitmap so all vector information is lost. Here I used a slightly modified version of the plot2svg open library in order to implement this. This is still work in progress and very preliminary.


martcous:
@pwdonh Can you help me test this?

martcous:
This does not seem to work very well on time-frequency maps.

pwdonh:
How do I pull those changes to my local copy in an easy way? I'm not so git-savvy:)

martcous:
git fetch origin pull/6/head:save_figure_svg

pwdonh:
It works in general.

Yes, for time-frequency maps it's a bit problematic, since it seems to save every tile separately. When I load it in inkscape I can select and move each one individually. 

When I plot an image in Matlab and save it, that's not the case. As in:

imagesc(rand(100));
saveas(gcf, './image.svg')

There the image seems to be saved as one 'object', whatever that means in a vector file.

PR #7, opened 2017-05-04
Changed default font to a list of default fonts to support OS without Arial

martcous:
Since Linux doesn't come with Arial (proprietary font), all menu items with the default font would show up as Sans Serif, which would not properly show the arrow keys (^ and v). I added a list of default fonts, and Brainstorm tries to load one after the other, and if everything else fails, loads Sans Serif. Apparently, "Liberation Sans" is a good open-source Arial alternative. If you have other ideas we can add it to the list.

### DIFF
--- a/toolbox/core/bst_get.m
+++ b/toolbox/core/bst_get.m
@@ -3201,16 +3201,43 @@ switch contextName
         
     case 'Font'
         fontSize = varargin{2};
+        fontTypes = {};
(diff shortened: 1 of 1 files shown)


ftadel:
Another way to fix this issue was to create icons for the [^] and [v] buttons.
But it was good to fix the default font selection as well.

PR #8, opened 2017-05-05
Added log XScale, XGrid and YGrid buttons to PSD plots

martcous:
Added log XScale, XGrid and YGrid toggle buttons to PSD plots. All that's missing is a proper icon for log scale, as discussed by email.

### DIFF
--- a/toolbox/gui/figure_spectrum.m
+++ b/toolbox/gui/figure_spectrum.m
@@ -731,6 +731,17 @@ function SetShowLegend(iDS, iFig, ShowLegend)
     % Redraw figure
     UpdateFigurePlot(hFig, 1);
 end
+function ToggleGrid(hAxes, hFig, xy)
+    ToggleAxesProperty(hAxes, [xy 'Grid']);
+    ToggleAxesProperty(hAxes, [xy 'MinorGrid']);
+    
+    % Toggle selection of associated button if possible
+    buttonContainer = findobj(hFig, '-depth', 1, 'Tag', ['ButtonShowGrid' xy]);
+    if length(buttonContainer)
+        button = get(buttonContainer, 'UserData');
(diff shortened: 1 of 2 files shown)


ftadel:
To make things even simpler, I think we talked about having only one button for both X and Y grids.
Unless you think it's really not a good idea, could you make it this way? (just trying to number of buttons as low as possible)
Thanks

ftadel:
And your ICON_LOG is available in the new brainstorm.jar.

martcous:
OK, I merged the X and Y grids button and added the new LOG icon. Ready for review!

ftadel:
Ok for the modifications to figure_spectrum.m, but not for figure_timeseries.m
I don't see any reason why we would want to display the time in log scale. 
Please remove this - Thanks

martcous:
The added buttons only affect the spectrum plots, thanks to this line:
```
if ~strcmpi(GlobalData.DataSet(iDS).Figure(iFig).Id.Type, 'Spectrum')
        set([h8 h9], 'Visible', 'off');
    end
```
They were added to figure_timeseries.m because that is where the function CreateScaleButtons() is defined and this function is called by figure_spectrum.m to create its buttons as well.

ftadel:
Oops sorry, I read that too fast. Will process this soon.

ftadel:
The LOG button works well, but the Grid button still requires some work:
- I added it also for the recordings (this is where it's mostly needed)
- When switching between the two types of display modes (butterfly / columns), the selected properties should be maintained, both for the PSD and the time series plot
- Use the fields TsInfo.ShowXGrid and TsInfo.ShowYGrid to record the expected status of this property for the plot - this way, it will be also compatible with the menus in the popup menu Figure > Show XGrid/YGrid
- In "columns" mode: we only want to display the X grid, the Y grid is too dense and makes no sense.

PR #11, opened 2017-05-18
Made Grid and Log buttons consistent when switching display modes.

martcous:
1. Fixed memory leak in PSD figures that would create new button elements every time you would change display modes.
2. Made both Grid and Log buttons to keep their selection & properties when switching between display modes.
3. Made Show Grid button to only display the X grid of plots in columns mode.

### DIFF
--- a/toolbox/gui/figure_spectrum.m
+++ b/toolbox/gui/figure_spectrum.m
@@ -713,12 +713,14 @@ end
 
 
 %% ===== HIDE/SHOW LEGENDS =====
-function ToggleAxesProperty(hAxes, propName)
+function newPropVal = ToggleAxesProperty(hAxes, propName)
     switch get(hAxes(1), propName)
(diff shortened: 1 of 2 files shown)


ftadel:
Looks good, but a few comments:
- figure_spectrum.m line 1040: I don't think this is a good idea to clear the entire figure, it deletes the axes and then has to create it. Can you work it out with more targeted deletions of the objects? Have you checked how it works for figure_timeseries?
- figure_timeseries.m line 2611: You don't need to update the field TsInfo.YGrid when switching to column more, just hide the YGrid from the plot. This way, it doesn't change the configuration of the figure when switching back to butterfly

(I read that quickly, maybe I'm wrong, this was all written a long time ago)

martcous:
I've addressed your second point. For the first point, this is exactly how it works in figure_timeseries.

PR #13, opened 2017-05-29
Chosen measure display option of timeseries is now saved

martcous:
As requested by Sylvain via email:

> I am looking at a lot of power spectra these days with BST and it is annoying that the display options (log(power), magnitude or power) are not consistent across displays (cortical maps or spectra) or remembered by BST when opening a new file. For instance, the default is always log(power). Can you please look into how to make it the last selected display option by the user?

### DIFF
--- a/toolbox/core/bst_get.m
+++ b/toolbox/core/bst_get.m
@@ -181,6 +181,7 @@ function [argout1, argout2, argout3, argout4, argout5] = bst_get( varargin )
 %    - bst_get('ResizeFunction')          : Get the appropriate resize function
 %    - bst_get('groot')                   : Get the root graphic object
 %    - bst_get('JFrame', hFig)            : Get the underlying java frame for a Matlab figure
+%    - bst_get('DisplayMeasure')          : Display option of measure for time series (log, power, magnitude, etc.)
 %
 % SEE ALSO bst_set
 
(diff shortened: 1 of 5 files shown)


ftadel:
OK, except for modification on figure_timefreq > GetFigureData.
This function is supposed to return what's available in this figure, not to interfere with the way it is supposed to be rendered.

The properties of the figures are defined at their creation, in the view_*.m functions. In this case, it will be in view_spectrum.m (see lines 112-120). 

You have to be very careful with this display function for the "time-frequency" figures, as the available options are not the same for all the files, and it would not work to simply save the previous function and reuse it at the next plot.

This problem doesn't exist if you stick to PSD plots, as you will always have the options "log", "magnitude" and "power" available. I'd recommend you stick to the PSD/FFT case (figure_spectrum), to avoid any random side effects of these modications.

In panel_display, add a test to make sure you are looking at the appropriate type of figure.
Change the name of the option you added to "LastPsdDisplayFunction" or something like this.

martcous:
That's unfortunate, I liked that it was also compatible with timefreq figures. Thanks for the comments -- done!

ftadel:
We have to be careful not to break anything else. This last version looks much less harmful than the previous one.

Plus, I'm not sure it would be a good idea to save the last option selected as the default value for the time-frequency figures. The average user plays a lot with all the buttons in the interface without understanding much to what they do, then they close the figures, and the week after they cannot figure out why it is not opening in the way they are used to (ie. displaying the power for the time-frequency maps) or worse, they don't even realize it is not the power that is displayed.
Not all the options are good to save in the user preferences. Keeping some values to good stable defaults is a good way to make the software more resistant to non-expert users.

PR #15, opened 2017-05-30
Added HeadModel info to connectivity outputs

martcous:
RE: Beth issue. I also reverted the temporary change when results type files did not have a HeadModelType, since this is not a normal state, as discussed.

I'm assuming here the following. Correct me if that is not ok.
1. there is a 1-to-1 relation between HeadModelFile and HeadModelType fields (when one is populated, so is the other one).
2. Populating the HeadModelFile and HeadModelType fields when possible in the LoadInputFile() bst_process has no unforeseen consequences.


ftadel:
Looks good!

PR #16, opened 2017-06-02
Process Extract Time now properly works when extracting a single time point

martcous:
Bug reported here: http://neuroimage.usc.edu/forums/t/extract-time-file-does-not-have-correct-time-definition/2996/2

I copied the behavior from the Average Time process.


ftadel:
Corresponding bug was taken care of in commit: https://github.com/brainstorm-tools/brainstorm3/commit/35a9cd478dee946614ad2dc48321c7b61b2f6fe3

The two time values must be strictly increasing, otherwise it causes lots of other problems. I fixed this in the "average time" and "extract time" processes.

PR #17, opened 2017-06-15
Added downsampling option to ICA process.

martcous:
Needs testing.

Let me know if you want some sort of validation on the entered frequency (e.g. error if it is higher than current or too low).

### DIFF
--- a/toolbox/process/functions/process_ica.m
+++ b/toolbox/process/functions/process_ica.m
@@ -55,6 +55,10 @@ function sProcess = GetDescription() %#ok<DEFNU>
     sProcess.options.eventtime.Comment = 'Event window (ignore if no event): ';
     sProcess.options.eventtime.Type    = 'range';
     sProcess.options.eventtime.Value   = {[-.200, .200], 'ms', []};
+    % Downsampling
+    sProcess.options.resample.Comment = 'Downsample (0=disable): ';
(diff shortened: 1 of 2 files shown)


pwdonh:
Hi Martin, thanks for this. I tried it out and it works.

When the ICA is run on downsampled data, it often complains about having too little samples (as a warning):

_There is probably not enough data for a correct ICA decomposition. Number time samples in input: ~ Recommend number of time samples:729000 (18225s)_

The recommend number of samples is computed in `process_ssp2.m, line 552`, from the number of channels as: `10*nChan^2 `. I was thinking it would make sense to use the requested number of ICA components instead (I use usually 40 components). Since the data is reduced to 40 dimensions with a PCA first, the ICA is computed on a 40-dimensional data matrix.

martcous:
@ftadel What do you think about Peter's suggestion of changing the equation for the warning to use the number of ICA components as opposed to the number of channels (process_ssp2.m, line 550)?

ftadel:
Peter's comment about number of requested data points: yes
https://github.com/brainstorm-tools/brainstorm3/commit/32c7a18741b6ca2379efb966282a990c2d84f728

Also added resample for process_ica2:
https://github.com/brainstorm-tools/brainstorm3/commit/6b7cd16e881ac02791c5bdae5554597498653007

Issue #18, opened 2017-06-19
Remember "Resample recordings" value in "Import in database..." menu


jeremymoreau:
In the "Pre-processing" box of the  "Import in database..." menu it would be nice if the "Resample recordings" value defaulted to the most recently set value instead of resetting to the sampling frequency (e.g. 2400 Hz) every time.

martcous:
Done in PR #19.

PR #19, opened 2017-06-19
Resample rate option is now saved when importing in database

martcous:
Per issue #18.
The only downside is that the sampling rate of the file is now no longer shown in that menu. Would you like to add it as a text field?

### DIFF
--- a/toolbox/gui/panel_import_data.m
+++ b/toolbox/gui/panel_import_data.m
@@ -340,7 +340,7 @@ function [bstPanelNew, panelName] = CreatePanel(sFile, ChannelMat) %#ok<DEFNU>
 
         % === RESAMPLE ===
         % Resample checkbox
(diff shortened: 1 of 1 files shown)


jeremymoreau:
Thanks! If it's not too much work, then yes perhaps displaying the sampling frequency next to the text box might be useful, e.g. Resample recordings (sampling freq = 2400 Hz): ______
I wouldn't bother if it's complicate however, since it can be viewed elsewhere and I wouldn't expect most users to change this very frequently (personally, I always sample at 2400 Hz).

martcous:
Added current sampling frequency in resample text with Jeremy's approval ;)

ftadel:
I realized after merging that you were not using the correct option name. I fixed your edits to use the existing field ResampleFreq instead of the non-existing ResampleRate.
https://github.com/brainstorm-tools/brainstorm3/commit/c0b1f3c30a5eabd81e1f9f27f90e83a288570292

martcous:
Thanks Francois!

PR #22, opened 2017-06-27
Fixed OpenMEEG download links

martcous:
The OpenMEEG download links were broken due to a new version coming out. It seems the new release doesn't have all the different platforms we were supporting so I simply changed the links to the new path of the older version.

### DIFF
--- a/toolbox/forward/bst_openmeeg.m
+++ b/toolbox/forward/bst_openmeeg.m
@@ -74,13 +74,13 @@ else
 end
 % Get default url
 switch(osType)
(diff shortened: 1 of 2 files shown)


agramfort:
sorry @martcous we moved stuff without being careful.

I moved again the files to have shorter paths:

http://openmeeg.gforge.inria.fr/download/?C=M;O=D

cc @eolivi

martcous:
Thanks for the heads up @agramfort, fixed again.

agramfort:
thanks. All good on my side

PR #23, opened 2017-07-17
Added new option to volume head model for atlas from scouts

martcous:
Added an option to MRI volume types of Head Model computation where you can choose an atlas (scouts) from a dropdown list. Brainstorm will then compute a centroid for each of your scout and extract a source at that region.


martcous:
As discussed, made this an external process instead: https://github.com/brainstorm-tools/bst-users/pull/3/files

PR #26, opened 2017-07-28
added code in the tess_detect_region for compatibility with the USCBrain

ajoshiusc:
The labels are checked if they have postfix and removed before sending to case statement. This change is to make the BrainStorm work with USCBrain atlas descriptions.


ajoshiusc:
bst_get.m updated with the correct name USCBrain_BrainSuite_2017.

PR #27, opened 2017-08-15
Added 2D rotation to 3D figures with control click

martcous:
Here's something that's been bothering me. It is a common behavior in 3D viewers that Shift + click gives you free 2D translation and Control + click gives you free 2D rotation. However, the Control + click rotation is not implemented in Brainstorm. I understand that Control + click is used for some other functionalities (with MRI slices for example), and I did not want to change that. So I essentially added the Control + click free 2D rotation to the 3D viewer when that key combination is not already used.

### DIFF
--- a/toolbox/gui/figure_3d.m
+++ b/toolbox/gui/figure_3d.m
@@ -35,7 +35,7 @@ function varargout = figure_3d( varargin )
 % For more information type "brainstorm license" at command prompt.
 % =============================================================================@
 %
-% Authors: Francois Tadel, 2008-2016
+% Authors: Francois Tadel, 2008-2016; Martin Cousineau, 2017
 
 eval(macro_method);
 end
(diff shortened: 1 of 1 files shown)


ftadel:
I don't think this works, it blocks the popup in some cases. 
Matlab is annoying because it processes the CTRL+click and the right-click as the same event ("alt").
If you want to be able to handle the control + click correctly, you need to track the status of the CTRL key. But because not all the events are processed correctly, all my attempts to do this were quite unsuccessful (sometimes, the CTRL key release is simply ignored, or processed before the key-down event...)

To rotate laterally (camroll), you can simply click+move on the sides of the figures:
http://neuroimage.usc.edu/brainstorm/Tutorials/ExploreAnatomy#MRI_in_3D
Don't you think this is enough?
Which programs were you referring to?

martcous:
Fair enough, I wasn't aware of the limitation. Simply Click + move doesn't work properly for me, it's a free 3D rotation rather than restricted to the plane of the window. Definitely not a priority, just something that bugged me. I've been used to it with viewers such as MI-Viewer, VTK viewer and Brainvisa's Anatomist (if I recall correctly, haven't used them since my master's). If there is an another available <Key> + click combination, we could use it.

ftadel:
Click on the left or right of the figure and move the mouse up/down.
Or click on the bottom or top of the figure and move the mouse left/right.
This does what I understood you describe.

martcous:
Sure, this works if I move the mouse in a perfect straight line haha.

ftadel:
Any particular reason why you need this rotation to be perfectly in the plane of the image?

martcous:
Just used to it. Very low priority like I said. You may close this.

PR #28, opened 2017-08-20
Add export of channels to brainsight coordinate file (for NIRS)

thomas-vincent:
It uses the subject MRI to convert coordinates and define the origin of exported coordinates.
Tested under SPM8 and brainsight: exported coordinates are consistent with anatomy exported as nifti.
If MRI is exported in another format than nifti, coordinates may not be consistent.


ftadel:
I added recently a better support for the sform/qform matrices from the .nii files:
https://github.com/brainstorm-tools/brainstorm3/commit/d3360fa7a323e7a4e861dba2e29596c8eef419ca


The export function out_mri_nii includes this additional part, which doesn't seem to be included in your code:
https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/io/out_mri_nii.m#L107

ftadel:
I don't know if you need to fix anything here.
Let me know when you want me to merge this PR

thomas-vincent:
Indeed, I missed sMRI.Header.nifti. It should be more accurate now.

On another matter, sometimes the sform field exists but contain only zeros, but qform contains a valid transformation. It is a concern that should be addressed in out_mri_nii.m? (there the qform is not used in the export).

ftadel:
If the volume was coming from a .nii file, the qform is unchanged in the exported volume.

If it is not, there is no way we can rebuild this information. It is supposed to contain the transformation between the voxels and the scanner coordinates, and this is not stored anywhere in Brainstorm.
https://nifti.nimh.nih.gov/nifti-1/documentation/nifti1fields/nifti1fields_pages/qsform_brief_usage

thomas-vincent:
Ok, thanks for the clarification.
My bad, I didn't see any line actually saving sMRI.Header.qform but in fact it is saved via:
    fwrite(fid, nifti.quatern_b,  'float');
    fwrite(fid, nifti.quatern_c,  'float');
    fwrite(fid, nifti.quatern_d,  'float');
    fwrite(fid, nifti.qoffset_x,  'float');
    fwrite(fid, nifti.qoffset_y,  'float');
    fwrite(fid, nifti.qoffset_z,  'float');

Anyway, this is quite unrelated to the channel export. I think we can close this PR.

PR #32, opened 2017-08-28
Added new connectivity process "Phase Transfer Entropy"

martcous:
As suggested by Sylvain Baillet and Benjamin Morillon:

> This method is a directed measure of connectivity (akin to Granger in the frequency domain), which has had some success lately, incl. on Benjamin’s study.
> The output values are normalized between -0.5 and 0.5, with 0 meaning no preferred directionality. So >0 values indicate inflow (B->A) and <0 ones indicate outflow (A->B).


ftadel:
Your code is online and pushed to /meg/meg1.
It would be good to include some form of online documentation with this method (and with all the other connectivity measures...)

PR #33, opened 2017-08-30
Bug fix: Multiple input files for "process_pac_fpMap"

SoheilaSamiee:


ftadel:
After merging this PR, I realized that some of your new files mix the camelCase and "_" conventions in the naming. In general, I'm trying to keep the same convention for everything (even though it's not always clear inside the code). 
Could you please remove the capital letters from the file names (and replace them with "_" if you think it's necessary)?

SoheilaSamiee:
OK, sure, I will modify them,

PR #43, opened 2017-10-24
EDF Reader: added case where channel names have spaces around dash.

martcous:
Here was the issue: the channels were named "EEG XX - Pz". The current code worked fine without the spaces around the dash ("EEG XX-Pz") and properly used Pz as ref and extracted XX. I added a case where " - " would be present and first replace it by "-". This is the simplest fix in my opinion, but once again if you think this is too specific to this user feel free to reject.


ftadel:
Perfect

PR #44, opened 2017-10-26
Integrated Plotly: Spectrum and Timeseries figures can be exported to a Plotly server

martcous:
- Added Plotly icon (thanks for your help)
- Added Plotly's Matlab library in external (untouched)
- Added Plotly username, API key and domain (optional) to the Brainstorm preferences window.
- Spectrum and timeseries figures can now be exported to Plotly from the right click -> Snapshot -> Export to Plotly menu.
- In order to make this work properly but keep the Plotly's external library untouched, I had to do minor hacks to the figures. I basically clone the figure (otherwise it becomes not functional and I would be forced to close it after sending it to Plotly), and remove / do minor changes to some elements such that it is compatible with Plotly. This is performed in out_figure_plotly(). This avoids having to do any changes to the Plotly external library so that we can easily update it if needed in the future.
- Future work: integration with the histogram figure. We'll see what else could be useful (3D figures are out of the question for now).

plotly1
plotly2


ftadel:
This is too big to add for just a few users, we don't want to include all this plotly library in /external. The main package is downloaded twice (script and compiled) every time users update Brainstorm, every extra Mb is a pain for most users outside of North America.

Please make this as something that is downloaded in $HOME/.brainstorm/ when needed.
You can find some inspiration in bst_openmeeg.m

As this will remain a relatively confidential feature outside of the MNI, I would also prefer to have the options set separately from the main options window. Whenever needed you can ask for the configuration options with java_dialog('Input', {...}, ...), and save the input in the preferences (extra entry in bst_get/bst_set).

martcous:
Fair comment, I'll do that, thanks!

martcous:
To avoid having all the plotly library in our Git history, I created a new PR (#45).

Issue #51, opened 2017-12-07
Cursor stuck at start of page after opening a second figure/window


jeremymoreau:
I am using Brainstorm 06-Dec-2017 on Ubuntu 16.04.3 with Matlab 2016b. My desktop environment is MATE 1.12.1 and I use the mutter 3.18.3 window manager. I have noticed an annoying bug where the red cursor is reset and stays stuck at the start of a page after opening a second window/figure and clicking anywhere one the first window. This does not affect only the latest version of Brainstorm (unfortunately, I do not know in which version this appeared), but seems to be platform dependent (I could not reproduce the problem on another workstation running an older version of Ubuntu MATE with a different window manager).

Steps to reproduce:

1. Open a MEG recording time series view.
2. Move the red cursor by clicking somewhere (e.g. at 5s)
3. Open a second time series (e.g. EOG) or figure (e.g. topography)
4. Click somewhere on the first time series as if to move the red cursor. The red cursor is reset to the start of the page, and won't move.
5. If the second window that was opened is a time series (e.g. EOG), then it is still possible to move the red cursor by clicking on that window, but clicking anywhere on the first window resets the cursor position again. Moving the cursor with arrow keys or page up/down is also still possible.

martcous:
I cannot reproduce the issue on my Ubuntu 16 with the same version of Matlab. Could it be specific to certain MEG time series?

jeremymoreau:
Yes, as we discussed earlier, this only seems to affect raw time series. Imported time series do not appear to be affected.

jeremymoreau:
I've figured out a couple more things.

- I believe this bug is related to the implementation of NVIDIA's TwinView feature on Linux, which treats multiple monitors a single extended x-screen. This is an old feature (c. 2000) and is on by default. The alternative mode (using separate x-screens for separate displays) is not attractive because it prevents drag-dropping windows from one screen to the other.
-  In my setup I have two 2560x1440 monitors (display 0 on the left at (0,0) , display 1 on the right at (2560, 0)). Display 1 is the primary display. I want the Brainstorm main interface window to be on disp0 and the figures to open on disp1.
    - In the "NVIDIA X Server Settings". Setting the position of display 1 to "relative (Right of)" instead of the default "Absolute +2560+0" makes no difference.
[...]

PR #54, opened 2017-12-12
Bugfix of strange cursor issue in time series figures.

martcous:
See issue #51. I could only reproduce it on Jeremy's computer. Essentially, get(hAxes, 'CurrentPoint') sometimes returns negative X coordinates, which get remapped to 0. I think it's related to having multiple screens. However, I could not find any documentation about it online. In that state, 0 is the rightest position on the figure, and everything to its left is negative. Since it includes the right "border" of the figure, I had to hardcore its seemingly fixed value. See screenshot below for what I'm referring for.
time_series_right_border

If you think of a better way to fix this, or feel this issue is too specific, feel free to comment.

### DIFF
--- a/toolbox/gui/figure_timeseries.m
+++ b/toolbox/gui/figure_timeseries.m
@@ -635,16 +635,9 @@ end
 function X = GetMouseTime(hFig, hAxes)
     % Get current point in axes
     X = get(hAxes, 'CurrentPoint');
-    X = X(1,1);
     XLim = get(hAxes, 'XLim');
-    % Weird bugfix where sometimes the cursor coordinates are flipped in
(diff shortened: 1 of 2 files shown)


ftadel:
You fix cannot work in a general context. This value 0.68 you used is following the units displayed inside in the axes. Change the page duration (0.1s or 100s), and these border pixels represent 10s or 0.001 seconds...
If you want to fix this, it has to be done in a way that includes both the figure size in pixels and the axes size within this figure. You need to estimate both the correspondence axis scaling/pixels and the distance of the axes to the right edge of the figure.

ftadel:
No update?

PR #55, opened 2018-01-12
Update in_fopen_eeglab.m

svetlanapinet:
Some events might have a duration above one sample in EEGLAB. I added a couple lines that should take care of it and import it into Brainstorm. "EEG.event(ii).duration" is always a field in the EEGLAB structure, but will be equal to 1 if event doesn't have a specific duration.


ftadel:
@svetlanapinet I added additional checks in the code because it was crashing on some examples I had.
https://github.com/brainstorm-tools/brainstorm3/commit/10b55e5d0098807f9d06ce5bd48bf50370b47403

I cannot test that it still does what you want because I didn't have any files with non-zero numerical duration fields. Try again that it still works for you.

svetlanapinet:
Tried it with a couple of my files and it works perfectly! Thanks!

PR #59, opened 2018-02-03
recompute bandpass filter after downsampling

pwdonh:
I started using ICA with the downsampling option again and noticed that the filters are not adjusted to the new sampling rate, so they could be way off. Before loading the data in process_ssp2, the filters are precomputed (in FiltSpec) at the original sampling rate (line 257). Those filters are then used on the downsampled data (line 519), which results in much lower frequencies. I didn't notice it when I tested before (PR#17), maybe I used a wide enough frequency band, so that eye and heart artifacts were still captured.

I added a piece to recompute the filters after downsampling, using the new sampling frequency. This means the length of the transient (nTransient) at original sampling frequency is still used during loading of the data, but changed to its new value after downsampling. Is there anything I'm missing?

### DIFF
--- a/toolbox/process/functions/process_ssp2.m
+++ b/toolbox/process/functions/process_ssp2.m
@@ -484,6 +484,11 @@ function OutputFiles = Run(sProcess, sInputsA, sInputsB)
         end
         sfreq = resample;
         TimeVector = TimeResample;
+        % Recompute filters at new sampling frequency
+        if ~isempty(BandPass) && ~all(BandPass == 0)
+            [tmp, FiltSpec] = process_bandpass('Compute', [], sfreq, BandPass(1), BandPass(2), 'bst-hfilter', isMirror);
+            nTransient = round(FiltSpec.transient * sfreq);
+        end
     end
     
     % ===== COMPUTE AVERAGE =====
--- a/toolbox/process/functions/process_ssp2.m
+++ b/toolbox/process/functions/process_ssp2.m
@@ -24,7 +24,7 @@ function varargout = process_ssp2( varargin )
 % For more information type "brainstorm license" at command prompt.
 % =============================================================================@
 %
-% Authors: Francois Tadel, Elizabeth Bock, 2011-2016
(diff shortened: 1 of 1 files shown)


ftadel:
Good catch. But what about we compute the filter correctly in the first place instead of recomputing a second one?
Would this work for you if instead of the change you posted in this PR, we replace the block "Compute bandpass filter" at line 250 with the following:
```
     % ===== COMPUTE BANDPASS FILTER =====
    % Get the time vector for the first file
    DataMat = in_bst_data(sInputsA(1).FileName, 'Time');
    sfreq = 1 ./ (DataMat.Time(2) - DataMat.Time(1));
    % Design band-pass filter
    if ~isempty(BandPass) && ~all(BandPass == 0)
        % If we need to resample the recordings
        if (resample > 0)
            filterFreq = resample;
        % Use the original sampling frequency
        else
            filterFreq = sfreq;
        end
        % Design the filter
        isMirror = 0;
        [tmp, FiltSpec] = process_bandpass('Compute', [], filterFreq, BandPass(1), BandPass(2), 'bst-hfilter', isMirror);
        nTransient = round(FiltSpec.transient * sfreq);
        % Show warning when computing epoched files
        if ~isRawA
            bst_report('Warning', sProcess, sInputsA, sprintf('Removing %1.3fs at the beginning and the end of each input for filtering.', FiltSpec.transient));
        end
    else
        nTransient = [];
    end
```

If it works for you, please update the PR with the corrected code.
Thanks

pwdonh:
The problem is that we need two different values of _nTransient_. During loading of the data _nTransient_ samples are loaded before and after the selected time window (at the original _sfreq_). And after downsampling and filtering _nTransient_ samples are discarded (at the new _sfreq_).

Maybe defining two _nTransient_ variables. Adding to your code example:
```
nTransientLoad = round(FiltSpec.transient * sfreq);
nTransientDiscard = round(FiltSpec.transient * filterFreq);
```
and replace the corresponding variables in the following? This would avoid another if statement

ftadel:
Something like this?

pwdonh:
Yes, that works like expected

Issue #53, opened 2017-12-08
Custom keyboard shortcut definition for EEG/MEG montages


jeremymoreau:
Switching montages via keyboard shortcuts is currently possible via shift + A, B, C etc. There are a couple of issues with this setup, however:

1. Deleting a bad EEG channel often results in all the keyboard shortcuts changing (because some montages become unavailable). This also means keyboard shortcuts are frequently different across runs...
2. The only way to change the keyboard shortcut for a montage via the GUI is by deleting multiple montages, and adding them back in a different order. This also makes it annoying to get the same keyboard shortcuts on different computers.
3. No keyboard shortcuts for montages >26 (which means only 8 custom montages on top of the defaults for CTF MEG). 

It would be nice if it were possible to set the keyboard shortcut in the montage editor (even if they all still follow the "shift + letter" pattern).

ftadel:
It would possible to set shortcuts the way we configure shortcuts for the event marking. But I won't have time to work on this before next year.
@martcous?

ftadel:
PR merged, now we need to update the documentation (text and screen captures):
https://neuroimage.usc.edu/brainstorm/Tutorials/ReviewRaw#Montage_selection
https://neuroimage.usc.edu/brainstorm/Tutorials/MontageEditor

There are probably many other places where some updates are needed, an exhaustive review of the tutorials is needed.

martcous:
Change documented here: https://neuroimage.usc.edu/brainstorm/Tutorials/MontageEditor#Assigning_a_keyboard_shortcut

ftadel:
Thanks!

PR #73, opened 2018-04-12
Added montage keyboard shortcut preferences (SHIFT + B-Z)

martcous:
RE: Issue #53.

Added new menu option in Montage Editor to edit montage keyboard shortcuts (SHIFT + A to SHIFT + Z).
montage1

 I kept SHIFT + A to be unchangeable and open all channels. The rest can be configured to a specific montage of the user's choice.
montage2

Note that this means by default montages will NOT have a keyboard shortcut. If you don't agree with this, we'll need to agree on what's the best way to implement this (I don't think we want half a state where some montages have the default value and others a custom one).
montage3


ftadel:
It's probably not a big problem if there are no shortcuts defined at the beginning.
I think people interested in this feature are people who are reviewing a lot of data, especially in clinical EEG or SEEG. And most of them would be ok to spend a minute defining their shortcuts correctly to later save a lot of time in their daily practice. It's worse in SEEG than in MEG or EEG, there are too manu montages available and the proposed shortcuts are currently not usable (as mentioned by Jeremy). 

However, asking clinicians to copy the exact name of the montage is not practical. If I understand well, people would have to enter manually the name in the text boxes. The problem is that montages in SEEG are called "Subject01: d' (bipolar 1)[tmp]", this is impossible to copy without making mistakes. For each new patient, clinicians would have to type about 10 of them. This is too complicated, no one would use this. (Have you discussed with Jeremy about this interface you posted here?)

I would see two possible options to improve this:
1) Embedded in the "Montage editor" window:
 - we could have a button "Set shortcut" that would define the shortcut for each montage individually (opening a window, asking to enter one letter for the selected montage)
- the name of the shortcut could be displayed on the right in the JList, similarly to the way it is done with the name of the function in the scout list of the Scout tab
- if the shortcut letter is already attributed somewhere else, too bad, we lose the association.
2) In the interface you wrote, replace the text boxes with JComboBox objects with the list of all the available montages. This way, selecting a montage with a complicated name becomes easier. (strip the "[tmp]" in the elements displayed in the list, to make it more readable)

Option #1 seems easier to code to me (a lot less interface). But having the two would be great.

In both cases:
- the subject name should be stripped (discard everything before ": ") from the association shortcut-montage. This way, the same shortcut would be reusable for multiple subject. Leading to something interesting: typing "A" could always bring the electrode "A", for all the subjects, avoiding the permanent redefinition of the shortcuts.
- it would be great if the shortcuts were displayed in the list of montages in the window "Montage editor"

ftadel:
(I edited a bit this comment, read it from github, not from the email :)

ftadel:
Why have you pushed another brainstorm.jar in this branch? Is it just a mistake?

martcous:
New icon and ListRenderer to fit new requirements, I will let you know when it's ready to review again.

martcous:
This is now ready to review. You can set the keyboard shortcut (B to Z) from the Montage editor using a keyboard icon:
image

image

The subject prefix and [tmp] suffix are stripped if present, so your selection will be saved between subjects. I kept Shift+A reserved for All channels.

martcous:
Refer to this commit for the Java modifications: https://github.com/brainstorm-tools/bst-java/commit/e4dbb789a4842fe3846454d0b9d1acf166be2b98

ftadel:
As we discussed during one of our development meetings, it would be nice to offer at least one modifier, so that users could have at the same time A and CTRL+A as shortcuts for instance.
This would match the need for SEEG of displaying electrodes B and B' (left and right depth electrodes implanted in similar brain regions).
For this specific purpose, it could be nice to offer a menu that would define by default all the standard SEEG shortcuts (eg. key A= montage A, key CRTL+A=montage A') 

Have you explored the technical options for doing this?
Once again, I encourage you to discuss this with Jeremy about the best possible approach for this.

martcous:
Yes, I did discuss with Jeremy about this... He did not think it was necessary.

ftadel:
Ok, let me test it then.

ftadel:
While I'm testing it, can you work on updating the documentation too? (text and screen captures)
https://neuroimage.usc.edu/brainstorm/Tutorials/ReviewRaw#Montage_selection
https://neuroimage.usc.edu/brainstorm/Tutorials/MontageEditor

There are probably many other places where some updates are needed. Can you please look in all the tutorials where there are other references to the montages, and possibly other screen captures to update (only if this is the montage is point of focus of the screen capture - otherwise it doesn't matter much)

ftadel:
It works great, but there is a major fix needed (on top of the doc):

On my win10 laptop, with my positions of windows, when clicking on the keyboard button, the window to select the shortcuts appears behind the montage editor window, making it impossible to use or close. You may have to remove the always on top or modal flags before showing your other dialog, and restore them again. I do this for several things in the code (consider you can have only one modal/alwaysontop window at a time).

This is heavily system dependent, it depends on the way Java/Swing is implemented and supported. Not all the systems will behave similarly, because some OS or window managers simply ignore the "always on top" flags of the windows. Operations that require superimposing multiple java dialogs should always be be tested on Linux, Windows and MacOS.

You also need to make the text of your input window scalable, like the rest of the interface (preference "Interface scaling").

martcous:
Thanks Francois, I'm on it for the improvements. Can you post these sort of commits in the PR next time? Otherwise I would not be notified unless you had not mentioned it.

ftadel:
Ok sure. 
(though I posted it after merging - I though about testing with an old Matlab too late...)

If it helps, you can be notified of all the commits on the repository. You need to add an email service to your github prefs, google for it if you're interested. This way I get notices of everything happening on the repo.

martcous:
AlwaysOnTop and interface scaling issues addressed: https://github.com/brainstorm-tools/brainstorm3/commit/e7e20ef658cf83afeed87d297a890bff9f94f479 and https://github.com/brainstorm-tools/bst-java/commit/495d600b7ce833e4427f240c4c34e83bfa06a9e5

ftadel:
Works great!

ftadel:
Are you going to take care of the doc as well?

martcous:
Updated the doc!

PR #74, opened 2018-04-13
Added pure-csv reader for MontageIn files

gabri470:
Brainstorm Native MON file format might be a bit cumbersome to write from external programs. Here I added a in_montage_csv file based on in_montage_mon that reads pure comma separated values (CSV) file following MON-specifications.


ftadel:
Hi Gabriele,

Thanks for the contribution.
Because tools are not used unless they are documented: could you add a short description of the file format? 
It would be great if you could write a paragraph describing the format (and if you want, the two others you don't like), I would add it at the end of this tutorial page: http://neuroimage.usc.edu/brainstorm/Tutorials/MontageEditor
(or let me know if you'd like to edit directly the website, I'd send you the instructions to create an account - it's an old wiki but still relatively easy to edit)

Additionally, it would be interesting to have the format described in the comments in the header of the function in_montage_csv. But I won't enforce that I rarely do myself :)

Cheers
Francois

PR #77, opened 2018-04-25
ICA component selection bug

pwdonh:
This fix is needed when only one reference channel is selected. See http://neuroimage.usc.edu/forums/t/ica-sort-components-based-on-correlation-results-in-only-one-component/6497


ftadel:
Thanks!

PR #88, opened 2018-06-06
Added STD/SEM option to Cluster time series

martcous:
It was requested in person to add Standard error to the cluster time series, so I added Mean+Std and Mean+StdErr as two new cluster functions. However, I am not sure if this would break something else, I've only ever used cluster functions from the Cluster tab.

image
image


martcous:
Thanks Francois, I added support for Standard Deviations to process_extract_cluster.

PR #91, opened 2018-06-22
Real-time head alignment bug fixes

Moo-Marc:
First bug would not load the second helmet (from a second dataset) into the same figure, but would reuse the first helmet/dataset.
Second bug would fail to properly find the helmet edge when the helmet points were in dewar coordinates.  That fix might affect other figures, maybe we should only adjust z, like in the 2dlayout method?

### DIFF
--- a/toolbox/gui/view_channels.m
+++ b/toolbox/gui/view_channels.m
@@ -113,7 +113,7 @@ else
 end
 
 % ===== LOAD CHANNEL FILE IF NOT DONE =====
-if isempty(GlobalData.DataSet(iDS).ChannelFile) || isempty(GlobalData.DataSet(iDS).Channel)
+if isempty(GlobalData.DataSet(iDS).ChannelFile) || isempty(GlobalData.DataSet(iDS).Channel) || ~strcmpi(GlobalData.DataSet(iDS).ChannelFile, ChannelFile)
     % Load channel file
     ChannelMat = in_bst_channel(ChannelFile);
     % Copy information in memory
--- a/toolbox/gui/view_channels.m
+++ b/toolbox/gui/view_channels.m
@@ -113,7 +113,11 @@ else
 end
 
 % ===== LOAD CHANNEL FILE IF NOT DONE =====
-if isempty(GlobalData.DataSet(iDS).ChannelFile) || isempty(GlobalData.DataSet(iDS).Channel) || ~strcmpi(GlobalData.DataSet(iDS).ChannelFile, ChannelFile)
+% Also check if we are processing the same channel file, which can be
+% different when doing real-time head position alignment (2 helmets from
+% different datasets in the same figure).
+if isempty(GlobalData.DataSet(iDS).ChannelFile) || isempty(GlobalData.DataSet(iDS).Channel) || ...
+    ~strcmpi(GlobalData.DataSet(iDS).ChannelFile, ChannelFile)
     % Load channel file
     ChannelMat = in_bst_channel(ChannelFile);
     % Copy information in memory
--- a/toolbox/math/bst_project_2d.m
+++ b/toolbox/math/bst_project_2d.m
@@ -31,7 +31,13 @@ end
 % Different projection methods
 switch (Method)
     case '2dcap'
-        % Spheric coordinates
+        % Subtract mean in case points are not in SCS coordinates, which is
+        % the case e.g. when doing real-time head position display (dewar
+        % coordinates).
+        x = x - mean(x);
+        y = y - mean(y);
+        z = z - mean(z);
+        % Spherical coordinates
         [TH,PHI,R] = cart2sph(x, y, z);
         % Flat projection
         R = 1 - PHI ./ pi*2;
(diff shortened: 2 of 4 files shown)


Moo-Marc:
Thanks Francois.  I've restricted the "centering" to the channel_tesselate case.  It was actually already being done in that file, so I moved the 2d projection after.  Tested on the realtime PC.

Moo-Marc:
While I understand the general sentiment, I think maybe you dismissed this fix a bit too quickly.  channel_tesselate used to work for the helmet in both SCS and dewar coordinates until the change to bst_project_2d.  I think restoring that robustness is the way to go, rather than working around the new limitation by drastically altering how realtime head tracking works.  
The change itself is rather simple: explicitly use centered coordinates to find the helmet edge instead of assuming the points are in are in SCS.  Of course, we could instead simply go back to the 2dlayout projection method instead of 2dcap, which is what introduced the problem, but I thought my fix was better.

ftadel:
Ok, I will give it a closer look when I’m back at my desk next week.  I’m travelling at the moment.
François

--
François

Le 10 juil. 2018 à 15:44, Marc Lalancette <notifications@github.com<mailto:notifications@github.com>> a écrit :


While I understand the general sentiment, I think maybe you dismissed this fix a bit too quickly. channel_tesselate used to work for the helmet in both SCS and dewar coordinates until the change to bst_project_2d. I think restoring that robustness is the way to go, rather than working around the new limitation by drastically altering how realtime head tracking works.
The change itself is rather simple: explicitly use centered coordinates to find the helmet edge instead of assuming the points are in are in SCS. Of course, we could instead simply go back to the 2dlayout projection method instead of 2dcap, which is what introduced the problem, but I thought my fix was better.

ftadel:
This function channel_tesselate is called by many functions. It is difficult to estimate what is the impact of modifying it, as some calls only have very indirect outputs (eg. computation of interpolation matrices).

The modification you propose alters the nature of what the function does: bst_project_2d does completely different things depending on where the center Z=0 is, therefore calling it before or after the centering of the sensors has a major impact on the output.

I haven't found any obvious case where the modification you propose causes a major failure, but it could exist. As I am not really clear on what the implications of modifying this function are,  there is a risk of appearance of nasty hidden bugs in specific configurations. I am very reluctant to bringing modifications we don't completely control to "stable" functions (functions that have been unchanged for a while a do not seem to create any immediate trouble). 
I'm sorry, it is mainly my fault, this is poorly coded in the beginning: some parts of the Brainstorm interface are not modular enough and are now big stacks of successive experiments that were not properly cleaned after. But I don't have much time to clean everything now...

The real-time head tracking is used only at the MNI, while the rest of the interface is used in thousands of labs. At the project level, it is safer to invest time in recoding pieces of the this local real-time head tracking tool than making general modifications that might break something in one of the thousands of other sites.

Can you please describe exactly what you current the issue is?
Which lines of bst_headtracking are causing problems?
Can you share an example to reproduce the unwanted behavior of channel_tesselate? (eg. right-click on the subject > File > Export subject, and share this zip file)
I'll try to think about ways to solve your issue without breaking anything anywhere.

Moo-Marc:
The heart of the issue is that channel_tesselate no longer works for a helmet in dewar coordinates, i.e. somewhat far from the origin.  (Are you sure only realtime tracking calls that function with non-SCS coordinates?)  Here's how this conflicts with head tracking:
bst_headtracking calls view_helmet twice, and [one time](https://github.com/brainstorm-tools/brainstorm3/blob/d85205ac809e5c0ce5855a3b44c96119d048d0c8/toolbox/realtime/bst_headtracking.m#L345) it's with the channel file created from the real-time buffer.  That one only contains the sensor locations in dewar coordinates.  You can view the problem (bad helmet triangulation, due to "bad" 2d-projection) by displaying the helmet of the attached subject's SensorPositions condition.
[HeadTracking_HeadMaster_RTpc_2018-07-16.zip](https://github.com/brainstorm-tools/brainstorm3/files/2203301/HeadTracking_HeadMaster_RTpc_2018-07-16.zip)

Then the list of calls to the source of the problem is: view_helmet, view_channels, figure_3d, channel_tesselate, bst_project_2d.

To get around the new limitation, we'd have to alter the sensor coordinates before saving the channel file on [line 311](https://github.com/brainstorm-tools/brainstorm3/blob/d85205ac809e5c0ce5855a3b44c96119d048d0c8/toolbox/realtime/bst_headtracking.m#L311), say by subtracting the helmet center from all coil locations, then doing the inverse translation on the helmet patch vertices right after [line 353](https://github.com/brainstorm-tools/brainstorm3/blob/d85205ac809e5c0ce5855a3b44c96119d048d0c8/toolbox/realtime/bst_headtracking.m#L353).  It's a bit messy since it saves a channel file in a somewhat undefined coordinate system, but that's the simplest solution I've thought of within bst_headtracking only.

Let me know what you think.

Moo-Marc:
Hi François.  I implemented your suggestion.  I also tested it with the latest Fieldtrip and it works as before.

Issue #52, opened 2017-12-07
Add a new default average referential EEG Montage with all left channels on top


jeremymoreau:
Creating new bipolar montages in Brainstorm is easy, but average montages are not so obvious. I would like an additional default (automatically generated regardless of number of EEG electrodes) montage with all the left channels on top, followed by midline channels, and right channels (e.g. Fp1, F7, ..., Fz, Cz, ..., Fp2, F8, ...). This has many use cases, but lateralisation of epileptiform activity is one. A spacer between the three groups (left, midline, right) would be nice also.

ftadel:
Good idea.
Can you prepare the montage(s) you want and send me corresponding text?
I'll add then as default options.

jeremymoreau:
The "text montage editor" isn't available for the average reference montage (the montage editor just shows a read-only identity matrix) unlike for the other bipolar or referential montage. The montage would also differ depending on the number of EEG channels recorded or bad channels set. I think this needs to be automatically generated like the current interleaved average reference montage, just in a different order.

This "Average reference (L -> R)" montage would go something like this:
Left channels
 :
Midline channels
 :
Right channels

The order within each group should be anterior -> posterior (e.g. Fp1, F7, F3, T7, C3, P7, P3, O1; this should just be the index order in the channel file since channels are typically labelled anterior-posterior by convention).

ftadel:
Ok, I see... I won't have time to work on this before next year.
Maybe @martcous is intersted in working on that?

martcous:
Done in PR https://github.com/brainstorm-tools/brainstorm3/pull/94

PR #94, opened 2018-06-27
Added new default sorted L->R avg ref EEG montage

martcous:
RE: issue #52.
Added a new default montage called "Average reference (L -> R)" or "Avg Ref LR" shortened that serves the same purpose as "Average reference" but sorts channels from left to right based on their EEG position in their name.
montage_lr
montage_lr2
After meeting with Jeremy, it was clarified that the current "Average reference" default montage should be kept as is, and that sorting using the original index is sufficient once sorted by left, mid and right.

### DIFF
--- a/toolbox/core/bst_figures.m
+++ b/toolbox/core/bst_figures.m
@@ -403,7 +403,7 @@ function UpdateFigureName(hFig)
             TsInfo = getappdata(hFig, 'TsInfo');
             if isempty(TsInfo) || isempty(TsInfo.MontageName) || ~isempty(TsInfo.RowNames)
                 strMontage = 'All';
-            elseif strcmpi(TsInfo.MontageName, 'Average reference') || ~isempty(strfind(TsInfo.MontageName, '(local average ref)'))
+            elseif ~isempty(strfind(TsInfo.MontageName, 'Average reference')) || ~isempty(strfind(TsInfo.MontageName, '(local average ref)'))
                 strMontage = 'AvgRef';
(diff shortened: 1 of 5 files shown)


ftadel:
Looks good!

However we do not want to offer this montage for all the EEG/SEEG/ECOG files. It relies on a 10-20 naming convention (z=center, even=right, odd=left), and would give random results if the channel naming is not following this 10-20 convention. Many of the research EEG systems (EGI, BrainProducts...) do not follow this naming convention. Plus the name of the SEEG and ECOG channels will never fit. 

Could you add some tests to make sure this menu is only offered when we are in presence of a 10-20 or 10-10 setup? For instance, maybe you can check that at least 50% of the standard 10-20 names are present in the channel list? Or any other heuristic that Jeremy thinks is a good solution.

Have you made sure your logic also works for the full 10-05 naming set? (example available in the default EEG caps in Brainstorm)

Thanks

martcous:
Good point, try that! I added a 50% check with a regexp.

ftadel:
Your current update with the regexp does not solve the issue, I think it would accept most SEEG montages and most EGI and BrainProduct caps, which we want to exclude. 
The simplest option would be to test the actual list of 10-10 channels (from the default channel file in brainstorm), with intersect(lower(data_channels),lower(10-10_channels)), or something like that.

martcous:
My regexp was working fine for SEEG and EGI during my tests  (and I thought it made sense for BrainProduct caps), but I changed it to use your method :)

PR #98, opened 2018-06-28
Added SVD warning for process_inverse_2016/2018 with Matlab 2018a

martcous:
Let me know if that is that you had in mind (only shown when interactive, not from a script).

### DIFF
--- a/toolbox/process/functions/process_inverse_2016.m
+++ b/toolbox/process/functions/process_inverse_2016.m
@@ -152,7 +152,7 @@ function [OutputFiles, errMessage] = Compute(iStudies, iDatas, OPTIONS)
         OPTIONS = struct_copy_fields(OPTIONS, Def_OPTIONS, 0);
     end
     
-    if bst_get('MatlabVersion') == 904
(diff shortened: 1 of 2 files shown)


ftadel:
This would be extremely annoying to have this message every single time for nothing... In crashes in very few cases.
A try-catch block around the calls to bst_inverse_linear_* would be more adapted. You could just rethrow the error with the original message, but adding a note about the SVD changes if Matlab==2018a .

It would be interesting to check if this was fixed with 2018b. Could you do that?

martcous:
Thanks Francois, is this better?

ftadel:
Have you tested if this syntax MException/throw/rethrow works with older Matlab (2008a)?
(With older Matlab versions, I'm not even sure the syntax "try e" works...)
Otherwise, you can probably simply create a new error by calling error()

martcous:
Yes, MException() and rethrow() were introduced in Matlab 2007a. I tested in 2008.
https://www.mathworks.com/help/matlab/ref/mexception.rethrow.html
https://www.mathworks.com/help/matlab/ref/mexception-class.html

PR #100, opened 2018-07-09
Adding a duration threshold for statistical maps

annesodub:
This modification adds a duration field in the panel_stat to filter out periods of significance that does not exceed a duration threshold (in sec). 

For instance in Kirchner et al. 2009 "Ultra-Rapid Sensory Responses in the Human Frontal Eye Field Region." (J. of Neurosc.) : 

"Only periods significant for at least 5 ms consecutively were considered in all analyses"
This can only be achieved by having the possibility to filter out period of significance that does not exceed 5 ms in duration. 

Please let me know if anything is wrong with the PR, I am a newbie.

Cheers,
AnneSo


martcous:
Please change back the HTTP USC links to HTTPS.

annesodub:
How can I do that?

martcous:
Change the http://neuroimage.usc.edu/ links in process_extract_pthresh.m to https://neuroimage.usc.edu/ links.

ftadel:
Wouldn't this be more readable and useful in milliseconds?
Are you ever going to be thresholding above 1s?

For aligning the two text boxes: add the constrain 'tab' when creating jTextDurThresh and jTextPThresh with gui_component.  You could also display the units after the text box, as it is done in most of the interface.

ftadel:
Many details fixed in: https://github.com/brainstorm-tools/brainstorm3/commit/0a3fa6ed15c47413f6a8ea34a967de1de34e76d8
(allowing the processing of source files, switching to ms instead of seconds, setting the default to 0s...)
The online documentation was updated as well.

@annesodub There is still the support for the time-frequency files to add. It is not working at the moment.

ftadel:
One more fix: https://github.com/brainstorm-tools/brainstorm3/commit/0a3fa6ed15c47413f6a8ea34a967de1de34e76d8

PR #104, opened 2018-07-25
Added support to import multiple event files at once

martcous:
This was requested by a user. It allows you to select import multiple event files at once instead of selecting them one by one in the GUI. I implemented it recursively because it required less line changes and the overhead was low, but if you insist I can make it iterative instead.

### DIFF
--- a/toolbox/io/import_events.m
+++ b/toolbox/io/import_events.m
@@ -58,18 +58,30 @@ if isempty(EventFile) && isempty(newEvents)
     % Get default directories and formats
     DefaultFormats = bst_get('DefaultFormats');
     % Get file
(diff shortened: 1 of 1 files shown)


ftadel:
I haven't tested it but it looks ok. Except that "newEvents(end+1:end+length(events)) = events;" won't work as expected if you are importing multiple files with the same event name. You would have multiple entries in newEvents instead of merged list of occurrences.

If you can think of a quick fix for this, I'll wait before merging. Otherwise, it doesn't matter much, this "newEvents" variable is not used for anything interesting (only for reporting).
Let me know when you want me to merge it.

martcous:
Sorry, I forgot about this. I've looked into it and I'd say it's fine to merge it like this. Otherwise, major refactoring would be needed.

PR #108, opened 2018-08-19
[MRG] Use bin matrices rather than mat to call openmeeg rather than 

massich:
temporal fix to https://neuroimage.usc.edu/forums/t/error-while-computing-sources-input-to-svd-must-not-contain-nan-of-inf/7806 while https://github.com/openmeeg/openmeeg/issues/341 gets fixed.


ftadel:
bst_openmeeg_bin has not been updated for a long time, I don't think this is a good idea to recommend using it as is. It would automatically replace openmeeg 2.4 with 2.2...

I think you'd rather submit a new bst_openmeeg_bin.m by copying bst_openmeeg.m and replacing all the .mat with .bin (what you suggested initially). Or adding the file extension as an input parameter to bst_openmeeg.

Is there anything else to change beside the extension change, to have it working with the .bin format?

massich:
Closing. see https://github.com/openmeeg/openmeeg/issues/341#issuecomment-415107121

PR #116, opened 2018-09-11
Added support for reading MFF files

martcous:
- Added a trimmed out version of the [mffmatlabio](https://github.com/arnodelorme/mffmatlabio) library to /external/.
- Added an automatic downloader of the MFF JAR file to HOME/.brainstorm/ from the [NeuroImage link](https://neuroimage.usc.edu/bst/getupdate.php?d=MFF-1.2.2-jar-with-dependencies.jar).
- Brainstorm checks if the JAR file is present during startup and adds it to the Java class path.
- The automatic downloader prompts the user to restart Matlab since you cannot add to the Java class path while the interface is running.
- in_fopen_mff() calls the importer of the mffmatlabio library which returns an EEGLAB object. It then imports the object in Brainstorm using in_fopen_eeglab().
- in_fread_mff() either loads the already loaded data from the header or uses mff_importsignal() to load it all on the fly.

Let me know if "EEG: Philips MFF (*.mff)" is not the best file type description and I can rename it to something better, I wasn't sure how to name it.


martcous:
1 is done but I disagree with 2 and 3. Like I mentioned by email, this would require essentially rewriting the mffmatlabio library. Most individual functions return an EEGLAB formatted output (e.g. mff_importcoordinates, mff_importevents). I don't think this is development time well spent (and maintenance time as well). If the size of the /external/ folder is the issue we can make the users download the whole repo at once. I already trimmed it by more than half its size excluding the JAR.

ftadel:
Ok for keeping the entire library unchanged.

I'm not worried about the size of the .m files, which are small and compress well.
But then maybe it makes more sense to download the entire library from the same source (jar + .m), rather than having only half of it on the Brainstorm github repo. It would make it easier to upgrade the library. Does it make sense?

I checked a while ago how to download a github repo from a single URL as a .zip (similarly to the download button from the github website) but could not find an easy solution. 
I put the full github clone here:
https://neuroimage.usc.edu/bst/getupdate.php?d=mffmatlabio-1.2.2.zip

martcous:
We can use the GitHub URL for downloading the repo (I do it for Plotly), but it won't work for our Java 6 users (pre Matlab 2013b). I can do a check on the java version to either use the updated GitHub repo for Java 7+ or the legacy neuroimage zip for Java 6.

ftadel:
Maybe it's easier to use only one version of the library copied to the neuroimage server then.

martcous:
Done!

ftadel:
Works great!
You just need to add something that does a better initial alignment between the sensors and our template anatomy, so that it looks a bit more like the screen capture I sent to Arnaud.
I think there was something like this in the the EEGLAB reader already...

PR #117, opened 2018-09-12
Added check that temporary directory is different from and does not contain the database directory

martcous:
This has crept up before on the forums where people would lose their brainstorm_db after emptying their temporary directory. I added a warning when they are the same or the brainstorm_db is inside the temp.


ftadel:
This is great!
I never thought about this as a possible cause for all these people complaining about losing all their databases :)
Thanks

PR #121, opened 2018-10-22
Added possibility to toggle a group (class) of process options

martcous:
As suggested by @Moo-Marc, it would be nice to be able to toggle a group of process options with a checkbox (see screenshot below).

process_options_class

I added two new optional process option fields:
1. **Class** = "ClassName", add this to options you would like to group together;
2. **Controller** = "ClassName", add this to a checkbox that toggles all options where Class =  "ClassName".

I am open to rename them to something more appropriate. Do note that "Group" was taken, which is why I settled with "Class".

Quick test: add this to a process (e.g. process_absolue.m)

>     sProcess.options.enabletest.Comment = 'Enable test?';
>     sProcess.options.enabletest.Type    = 'checkbox';
>     sProcess.options.enabletest.Value   = 0;
>     sProcess.options.enabletest.Controller = 'testClass';
>     sProcess.options.test1.Comment = 'Test field: ';
>     sProcess.options.test1.Type    = 'text';
>     sProcess.options.test1.Value   = 'test value';
>     sProcess.options.test1.Class   = 'testClass';

### DIFF
--- a/toolbox/process/panel_process_select.m
+++ b/toolbox/process/panel_process_select.m
@@ -792,6 +792,8 @@ function [bstPanel, panelName] = CreatePanel(sFiles, sFiles2, FileTimeVector)
         end
         % Set list tooltip
         jListProcess.setToolTipText(pathProcess);
+        % Initialize classes to be toggled off
+        ClassesToToggleOff = {};
         
         % === PROTOCOL OPTIONS ===
         for iOpt = 1:length(optNames)
@@ -808,6 +810,10 @@ function [bstPanel, panelName] = CreatePanel(sFiles, sFiles2, FileTimeVector)
             end
             % Enclose option line in a River panel
             jPanelOpt = gui_river([2,2], [2,4,2,4]);
+            % Add class name to panel
+            if isfield(option, 'Class')
+                jPanelOpt.setName(option.Class);
+            end
             % Define to which panel it should be added
             if isfield(option, 'Group') && strcmpi(option.Group, 'input')
                 jPanelInput.add(jPanelOpt);
@@ -995,7 +1001,12 @@ function [bstPanel, panelName] = CreatePanel(sFiles, sFiles2, FileTimeVector)
 
                 case 'checkbox'
                     jCheck = gui_component('checkbox', jPanelOpt, [], ['<HTML>', option.Comment], [], [], @(h,ev)SetOptionValue(iProcess, optNames{iOpt}, double(ev.getSource().isSelected())));
-                    jCheck.setSelected(logical(option.Value));
+                    isSelected = logical(option.Value);
+                    jCheck.setSelected(isSelected);
+                    % If class controller not selected, toggle off class
+                    if ~isSelected && isfield(option, 'Controller') && ~isempty(option.Controller)
+                        ClassesToToggleOff{end + 1} = option.Controller;
+                    end
                 case 'radio'
                     jButtonGroup = javax.swing.ButtonGroup();
                     constr = [];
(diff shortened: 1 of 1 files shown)


martcous:
Note to self: update the doc once this is approved: https://neuroimage.usc.edu/brainstorm/Tutorials/TutUserProcess

ftadel:
I understand why in terms of interface, it would look better to have some menus grayed out. 
However, I am not in favor of adding new controls to the pipeline editor if they don't correspond to anything useful. For instance, it is not necessarily useful to use one checkbox to enable/disable only one text box, when this text box can be emptied to disable an option (like the selection of the sensors). Be careful with not overloading the interface with buttons that do not bring any additional feature.

With what I understand in your code, in your example if I check the box "Enable test?", it enables the text box "Test field", but it does not change the value of option "test1". Therefore, whether your check the box "Enable test" or not, the values of all the other options would stay the same. 
I guess what you were expecting was to have to deal with the separate options in the Run() function of the process: If enabletest==1 then I get the actual value of option test1, otherwise I do not get/need this option?
This has to be explained in the documentation, as someone might expect that if option "test1" is not available (not enabled), it doesn't have a value.

I've linked explicitly checkboxes and controls only for two types of options (scout_confirm and cluster_confirm). In these cases the checkbox and scout selection get combined into one single option value. This was done many years ago, when there were only 10 or 20 processes, but we're not going to recode as new special input types all the options we want to group under the dependency of a checkbox.

Do you have examples of existing processes you think would look better with combined options like this? Or examples of new processes? (I guess you're not working on this without a precise objective?)

You can merge this pull request when you think it's ready but I won't test it before: be very careful that it does not break anything in the pipeline editor, and that all existing processes will keep on working exactly like they do now.

martcous:
> I understand why in terms of interface, it would look better to have some menus grayed out.
> However, I am not in favor of adding new controls to the pipeline editor if they don't correspond to anything useful. For instance, it is not necessarily useful to use one checkbox to enable/disable only one text box, when this text box can be emptied to disable an option (like the selection of the sensors). Be careful with not overloading the interface with buttons that do not bring any additional feature.

Of course a checkbox to enable a single text box is not useful, this was just a quick mock-up to show you how it works ;) In real use, you can expect multiple options to be linked to a single toggle checkbox.

> With what I understand in your code, in your example if I check the box "Enable test?", it enables the text box "Test field", but it does not change the value of option "test1". Therefore, whether your check the box "Enable test" or not, the values of all the other options would stay the same.
> I guess what you were expecting was to have to deal with the separate options in the Run() function of the process: If enabletest==1 then I get the actual value of option test1, otherwise I do not get/need this option?
> This has to be explained in the documentation, as someone might expect that if option "test1" is not available (not enabled), it doesn't have a value.

You understand correctly. Of course, this will be properly explained in the documentation once approved.

> I've linked explicitly checkboxes and controls only for two types of options (scout_confirm and cluster_confirm). In these cases the checkbox and scout selection get combined into one single option value. This was done many years ago, when there were only 10 or 20 processes, but we're not going to recode as new special input types all the options we want to group under the dependency of a checkbox.

I am not suggesting to change existing processes. This change does not break any existing process. This is just a new useful tool for new processes that have multiple optional parameters that are linked together.

> Do you have examples of existing processes you think would look better with combined options like this? Or examples of new processes? (I guess you're not working on this without a precise objective?)

A perfect example would be the Depth weighting parameters in the advanced options of Compute sources. They are working exactly in the same way, but are only possible right now because this is a custom additional panel. This PR will bring this feature to the main process panel.

> You can merge this pull request when you think it's ready but I won't test it before: be very careful that it does not break anything in the pipeline editor, and that all existing processes will keep on working exactly like they do now.

Of course. But I think looking at the code, it's pretty unlikely it breaks existing processes. Maybe some odd use cases where people do try to use it in their own processes on specific option types, but that can be fixed once reported and doesn't break existing features.

Thanks!
Martin

ftadel:
My review didn't change: "You can merge this pull request when you think it's ready but I won't test it before". Be careful that it doesn't break anything for the training in Chile, it's difficult to deal with a blocking bug happening on 40 computers at the same time :)

martcous:
Okay, I went through a bunch of processes using the online tutorials and everything seemed functional to me so I'll go ahead and merge.

martcous:
Updated the [User Process](https://neuroimage.usc.edu/brainstorm/Tutorials/TutUserProcess#Definition_of_the_options) documentation.

PR #122, opened 2018-10-26
Merge of basic electrophysiology tools

martcous:
This is a huge merge of basic electrophysiology tools mostly developed by @mpompolas, with Brainstorm functionality and refactoring done by myself.

The processes and features added here are documented in a series of tutorial pages about electrophysiology located here: https://neuroimage.usc.edu/brainstorm/e-phys/Introduction

A very small Java change is also needed, see PR here: https://github.com/brainstorm-tools/bst-java/pull/2

Here is a non-exhaustive list of the changes:

- Added support for reading file formats of Intan and Plexon files, expanded support for Blackrock files
  - Added Precision field to ImportOptions of in_fread() to support reading data as single rather than double to save memory in specific cases (only supported by formats listed above)
- Added processes spike_field_coherence (Electrode x Electrode), noise_correlation (Neuron x Neuron) and rasterplot_per_neuron (Time x Trial) that produce a time-frequency like figure. Instead of creating a new figure function for each, small hacks were made to the timefreq figure:
  - Custom proper axis labels instead of Time x Frequency are shown on the figure in these special cases;
  - The "Hide edge effects" option in the Display panel is disabled;
  - The frequency slider is disabled;
  - The measure is set to 'power' and the option is not shown in the Display panel.
- The 'MaxBlockSize' process option for raw files (default: 100Mb) can now be edited in the Brainstorm preferences.
- A new version of the .BST format that supports event labels of length longer than 20 characters was added.
- Three spike sorters ('WaveClus', 'UltraMegaSort2000' and 'KiloSort') were implemented in Brainstorm. They are downloaded and installed on the fly when the user requests them (except for the supervised part of KiloSort that requires an external program, Klusters).
- Unsupervised spike sorters require a data file per electrode rather than a single file containing all the data. Function in_spikesorting_rawelectrodes() "demultiplexes" data files to create a file per electrode in two passes (multiple implementations were tested and this was the fastest).
  - First, the data is read in blocks and binary .BIN files that can be appended are written with the raw data, one per electrode. This makes use of the new  precision option of in_fread() to save memory and disk space;
  - Then, these binary files are converted to .MAT files iteratively.
  - These huge demultiplexed files are saved in the Brainstorm temporary directory. Processes that require them checks if they are in the temp file, otherwise re-generates them on the fly.
- Unsupervised spike sorting processes use these demultiplexed files to create spike-sorter-specific "spike sorted" output files that are needed by each program for the supervised spike sorting step.
  - These are saved in the brainstorm_db next to the raw file under a new folder.
  - They are shown as a single "raw ephys" object in the brainstorm database next to the raw file, with the spike icon.
  - Double clicking on them has the same effect as running the supervised spike sorting process, and opens up the appropriate spike sorter and the Spikes panel.
  - Deleting this node in the brainstorm database deletes the corresponding folder in the brainstorm_db.
image
- A new 'Spikes' panel was added so that users can interact with their spike sorter of choice.
  - It has a dropdown with neurons. Selecting a neuron from the dropdown updates the loaded spike sorting software.
  - Clicking 'Save and Next' will convert the spike sorter specific event to a Brainstorm event, add this spike event to the raw file, add a star (*) at the end of the label in the dropdown to indicate to the user that this neuron was saved, and load the next neuron in the list.
  - Special case: KiloSort will list montages instead of neurons.
- Two new process option types were added (can be seen in action in process_tuning_curves.m):
  - 'event' will allow you to select one or more events from the selected data file, similarly to scouts selection.
  - 'event_ordered' is similar to 'event' but lets the user enforce a specific order in their selection of events.
  - The two types have an optional parameter 'Spikes' that allows process creators to enforce to include only or exclude spike events from the list.

That's all I can think of for now for an overview. The tutorial pages go in more details. Let us know if you have any question!


martcous:
Thanks for the first batch of the comments! They were all addressed.

mpompolas:
Thanks for the initialisation of the merge guys, I will try to follow the procedure as far as my time permits.
Make Electrophysiology Great Again!!!

ftadel:
https://github.com/brainstorm-tools/brainstorm3/pull/122/files/3216f7ccd1f4a1d1b739532cad28317f99346b36..61b171ad1891cf5e25d029ec678d6a20dc53a00e
You don't even need to check for the existence of the field. It's in the program data structure, that's it.

Everything else looks good!
Merge it whenever you want. I will read your tutorial and test all this practically next week.

ftadel:
Last thing to do: add a short paragraph with a some screen captures in the News page, which links to your tutorials: https://neuroimage.usc.edu/brainstorm/News
Something in the format of the November 2017 section (SEEG tutorial)

PR #130, opened 2018-11-15
dspm --> dspm2018

guiomar:
Update in the process_inverse_2018 for source reconstruction


ftadel:
Already done in
https://github.com/brainstorm-tools/brainstorm3/commit/a4ef49642d231df7475a701515938c0bd42c1146

PR #133, opened 2018-11-16
Added channel importer of MFF files (coordinates.xml)

martcous:
RE: issue #116 

> TODO: Add an entry to import directly the mff .xml files as channel files in the database

Let me know if this is what you meant. I made sure that it does not load the entire data in memory (it required adding an EEGLAB file to the repository because it is private in the MFF library). I kept it all in in_fopen_mff() to avoid code duplication.


ftadel:
I'm not sure I remember why reading independently the positions .xml files was needed, it seems like a lot of work for not many use cases. Anyway, it's good if it's done.
If it works for you, go ahead and merge.

My main concern with these .mff files was the registration in Brainstorm coordinates, and that I reported in the last comment of the issue you mention here. This is something I already fixed in commit  https://github.com/brainstorm-tools/brainstorm3/commit/838128ff38a26457ab46e71faecdd909a0df2c48#diff-114b3644fa80c9daaadcd90d7192e4a2

PR #137, opened 2018-11-28
Minor modification on line 1020 to address forum post reporting memory issues with large source grids. 

sbaillet:
Greater memory efficiency; works better for denser source grids and faster execution.

### DIFF
--- a/toolbox/inverse/bst_inverse_linear_2018.m
+++ b/toolbox/inverse/bst_inverse_linear_2018.m
@@ -1017,7 +1017,7 @@ if strcmpi(OPTIONS.InverseMeasure, 'amplitude')
         if NumDipoleComponents(kk) == 3 % we are in three-d,
             Kernel(Ndx,:) = WQ{kk} * Kernel(Ndx,:); % put the source prior back into the solution
         elseif NumDipoleComponents(kk) == 1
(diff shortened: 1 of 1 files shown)


sbaillet:
Done.

ftadel:
If the idea is just to scale the rows of matrix "Kernel", we'd rather use bsxfun instead:

```
tic; 
diag(Alpha{kk}) * Kernel(Ndx,:); 
toc
Elapsed time is 1.535137 seconds.

tic; 
bst_bsxfun(@times, Kernel(Ndx,:), Alpha{kk}(:)); 
toc
Elapsed time is 0.031496 seconds.
```

@jcmosher Can you please check that this does what you expect from this function?

ftadel:
@jcmosher ?

ftadel:
@jcmosher Can you please just confirm you receive copies of these messages by email?
Thanks

ftadel:
@jcmosher ?............

ftadel:
Anybody still cares about this PR?

jcmosher:
I read the code that Sylvain proposed. Looks good. As he points out, will be far more efficient for dense source grids. What I don't recall is why I would have programmed such a sloppy "diag" in the first place. "spdiags" is the correct approach here.

ftadel:
@jcmosher Thanks

PR #147, opened 2018-12-06
Updated RICOH/Yokogawa MEG reader toolbox

tomitomi3:
Updated RICOH/Yokogawa MEG reader toolbox.

- Topic
  - Supported KIT 24 bit MEG data.
  - Added 2nd Axial gradiometer structure.

There is no change API with this update.


tomitomi3:
Thank you!

PR #148, opened 2018-12-10
Fix line coloring for NIRS topography

thomas-vincent:
There is an issue with line colors that are not updated in topography when viewing NIRS data.

[This previous commit](https://github.com/brainstorm-tools/brainstorm3/commit/e24ff6c83deab434c8569612711a408f98e5bad0#diff-953276eb98a4ae071e2464b4033e98bcR178) changed  TopoHandles.hLines which seems to be a cell array now. However, for NIRS data, it still is an array of Line.

This commit preserves the new behavior and reenables the previous way of parsing TopoHandles.hLines.

### DIFF
--- a/toolbox/gui/figure_topo.m
+++ b/toolbox/gui/figure_topo.m
@@ -175,9 +175,13 @@ function UpdateTopoPlot(iDS, iFig)
         iDataCmap(iDataCmap > size(sColormap.CMap,1)) = size(sColormap.CMap,1);
         dataRGB = sColormap.CMap(iDataCmap, :);
         % Set lines colors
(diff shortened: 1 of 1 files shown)


ftadel:
Sorry for not updating correctly the NIRS display when I updated the 2DLayout.

I preferred to fix the "3DOptodes" mode at a higher level: 
https://github.com/brainstorm-tools/brainstorm3/commit/8fda432e7a5a9d6494b2793e373ab61c5ff75190

Does it work for you?

thomas-vincent:
Sorry for the late reply.
It works fine! Thanks.

PR #150, opened 2018-12-14
Fix OutChannelFlag variable assignment

annesodub:
Fixed OutChannelFlag variable assignment in order to be able to run a paired ttest between two datasets. Before this modification, the Stat.ChannelFlag structure  was set to empty, resulting with a "Index exceed matrix dimension" error.

### DIFF
--- a/toolbox/process/bst_avg_files.m
+++ b/toolbox/process/bst_avg_files.m
@@ -485,7 +485,7 @@ if strcmpi(matName, 'F')
     OutChannelFlag  = ones(NbChannels, 1);
     OutChannelFlag(MeanBadChannels) = -1;
 else
-    OutChannelFlag = [];
+    OutChannelFlag = ChannelFlag;
 end
 
 % === FINALIZE COMPUTATION ===


ftadel:
Nope, this fix doesn't work. The ChannelFlag you used here is the one from a random file in the input list.
If OutChannelFlag is empty, there is probably a good reason for it, and we need to fix the error that you got further down the stream.

Please post the full error message you get.

ftadel:
Any update on this topic?

ftadel:
?

annesodub:
Here is the error message I get when I try to perform a paired ttest. 
image

Input files (dummy example) : 
image

The ttest seems to perform OK with the proposed solution. Please let me know if I can try anything else.

ftadel:
I fixed this problem in a different way, using the correct list of bad channels across your samples. See this commit:
https://github.com/brainstorm-tools/brainstorm3/commit/16b64ffc0c22af1e15ed2b711f3b93a68ada033f#diff-529a6969d8600036c25e164573812e83L484

Update Brainstorm and let me know if it works for you.
Bises

annesodub:
Works like a charm! Thank you.
Bises

PR #154, opened 2019-01-03
Hossein shahabi

HosseinShahabi:
https://github.com/brainstorm-tools/brainstorm3/commit/168fc96b7e62a38e0ef189ad707d3d75a16432e7


ftadel:
Please merge your two pull requests (and rename this one...)
And I'll wait until you push all the commits that modify the calls to process_bandpass as well.

You said that you wanted to modify the notch filter: could you include these modifications as well in this PR? This way we would have only one single step for upgrading all the filters.

PR #158, opened 2019-01-09
Hossein bandstop/notch

HosseinShahabi:
Since I need to revisit the bandpass filter, you can apply bandstop/notch at this time.
These two codes are set with calls in other functions. 
I will put another PR for bandpass (and subsequent changes in other files) later.

### DIFF
--- a/toolbox/process/functions/process_bandstop.m
+++ b/toolbox/process/functions/process_bandstop.m
@@ -23,7 +23,7 @@ function varargout = process_bandstop( varargin )
 % For more information type "brainstorm license" at command prompt.
 % =============================================================================@
 %
-% Authors: Francois Tadel, 2014
+% Authors: Hossein Shahabi, Francois Tadel, 2014-2019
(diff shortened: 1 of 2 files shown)


ftadel:
It would be good to factorize the code that displays the filters specs, because it represents a large share of the code volume you posted here. The separation between the computation of the filter specifications and their display would be clearer.
Create a single function "DisplayFilterResponse" in process_bandpass for instance, and call it in all three filter functions (process_bandpass('DisplayFilterResponse', ...)) - It makes process functinos less independent, but it will make the code maintenance easier in the future.
Thanks

HosseinShahabi:
It won't be very efficient I guess.  
The "DisplayFilterResponse" is completely tailored for each function. It uses specific parameters, functions, and information regarding that function. So probably a small portion of the "DisplayFilterResponse" will be shared among all three. 
I am happy with the current one. So Please let it go! 
Thanks

PR #159, opened 2019-01-14
Speed improve

thomas-vincent:
2 proposal to improve speed:
- more efficient cell array modification in file_compare.m
- make bst_process/CallProcess also return sInputs to avoid calling bst_get('GetInputStruct') afterwards.

### DIFF
--- a/toolbox/anatomy/mri_isosurface.m
+++ b/toolbox/anatomy/mri_isosurface.m
@@ -28,10 +28,10 @@ if (bst_get('MatlabVersion') == 903)
     % Share the vertices manually
     relative_tolerance = 1e-12;
     sz = size(Vertices);
-    [C, ~, IC] = uniquetol(Vertices, relative_tolerance);
+    [C, ignore, IC] = uniquetol(Vertices, relative_tolerance);
     Vertices = reshape(C(IC),sz);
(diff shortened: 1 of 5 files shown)


martcous:
Were the file_compare modifications tested on older Matlab versions? We support up to Matlab 2008a.

thomas-vincent:
Hello,

I tested on matlab 2008b which is the oldest I could get. I had to make some fixes (rnd generetor init and removal of tilda in outputs)
 
Running test_all.m worked fine (except for an error which is unrelated, see [this issue](https://github.com/brainstorm-tools/brainstorm3/issues/160).
Running tutorial_introduction worked fine too.

What would you advise to further test things?

martcous:
Looks fine to me. @ftadel ?

ftadel:
I haven't tested for speed differences, but it can only better.
Thanks for the improvements!

And thanks for fixing the backward compatibility of the other functions! 
These "~" escaped my screening, they're usually part of the things I look for when reviewing new code :)

thomas-vincent:
When I started to do some tests, I thought it would save more time...

Now that I've redone the tests, it doesn't save that much, 8% only. But still improving!
 
image

image

PR #162, opened 2019-01-21
Threshold colormap according to t-test for better color dynamic

thomas-vincent:
This PR add the possibility to threshold the colormap to better represent t-stats.
Currently, non-significant values are masked prior to rendering the data but the colormap still assigns colors to these values:

image
The full color dynamic is not used and the colormap does not reflect stat thresholding.

To enable colormap thresholding, a new field was added in the colormap bst object: UseStatThreshold, 
and a new menu item:
image

The result shows more contrast and a grey area of the colormap indicates non-significant values :
image

In terms of implementation, I see a small issue by having to guess the t-test side from the stat and p-values (see process_extract_pthresh.m). But I think it's safe this way since I did not see anything stored in the data file to indicate the t-test side (except for the comment which is not a reliable source of information). The worst that could happen is that it falls back to the default unthresholded colormap anyway. 
It would be more troublesome to change what we expect from the saved data. 

Let me know what you think.


thomas-vincent:
Tested on matlab 2008b and 2017a.

sbaillet:
Great suggestion, Thomas: thanks!
And timely too as we were about to get into such changes shortly!

thomas-vincent:
I'm glad it will be useful!

I added a new colormap (ovun for over/under) that is more suited for 2-sided t-test. It gives consistency between close-to-threshold values: values just below the lower thresh have the same limuninosity as the values just above the upper thresh.  
I tweaked Brainstorm.jar to do so. I compiled with the same original target java version (release 6, major version 50) and tested with matlab 2008 and 2017.

ftadel:
Can you please open another PR to poste your Java modifications?
https://github.com/brainstorm-tools/bst-java/tree/master/brainstorm

I'll review the two at the same time (Thursday or Friday)

ftadel:
And please remove your modified brainstorm.jar from this PR
I'll recompile it and update it after reviewing the code
Thanks for this nice contribution!

thomas-vincent:
I wasn't aware of the bst-java repo... it seemed indeed strange to work only inside the .jar
I made the modification there, see https://github.com/brainstorm-tools/bst-java/pull/3

I reverted brainstorm.jar.

ftadel:
Wow! You dived really deep into the code for this! Very impressive :)

I pushed various fixes and modifications to your PR here:
https://github.com/brainstorm-tools/brainstorm3/commit/eab1962a49b79e1fa6822ca054a94f4adca2270e

- I moved your function cmapThreshold to bst_colormaps (as StatThreshold), because it might called from other figure types
- In process_extract_pthresh: it looks like you considered there was only one time point? Your functions crashed with multiple time points
- The option  "Use stat threshold" didn't seem to work well in combination with the option "Absolute values", so made them exclusive - if you fix it, feel free to disable the force reset of the other option (SetColormapAbsolute and SetUseStatThreshold in bst_colormaps)
- It would be nice to have the same display available for sensor-level topographies and time-frequency maps. I prepared everything for it, but stopped when I realized that there was something different in the case of the topo plots: the values are re-interpolated, therefore your t-thresholds should not be applied in the same way, otherwise it cuts all the values around the sensors. Left: original plot, Right: with the "Use stat threshold" option
image  image
Do you have any suggestion for fixing this? When it's done, we could make the ovun colormap + the option "Use stat threshold" as the defaults for the "stat2" data types.
- I made this menu visible only for the colormap type "Stat2", I thought it was not relevant for the others.

I hope I didn't damage anything...
Let's keep on working on this, this will really improve the display of the stat results!

ftadel:
Created related issue: 
https://github.com/brainstorm-tools/brainstorm3/issues/165

PR #163, opened 2019-01-22
Added support to read PL2 Plexon files

martcous:
We decided to keep the current external files that read PLX files, and let the users install the SDK themselves if they want to read PL2 files. I documented the simple process of installing the SDK in an advanced section of the electrophysiology tutorial and link it from Brainstorm if the user doesn't have the SDK installed: https://neuroimage.usc.edu/brainstorm/e-phys/Introduction#Importing_PL2_Plexon_files


martcous:
@mpompolas Please review :)

PR #167, opened 2019-01-29
added two ez fingerprint processes

silencer1127:


silencer1127:
In order to call EZF from process 2 without inputs, we need to allow users to click run to open the process panel. Therefore, instead of return when either input is empty, I generate dummy inputs using the template similar to that in process 1. The result is that users can open the panel but currently all processes except for EZF are greyed out. Hope this does not conflict with other stuff.

ftadel:
I still recommend two things I mentioned in the corresponding issue:
- Try to get rid of the image processing toolbox dependency in your program - dependencies to expensive external toolboxes are a major issue for code distribution
- Create a page on the Brainstorm wiki that presents your program and add a link "SEEG seizure analysis: EZ Fingerprint" in the section "Advanced processing" in https://neuroimage.usc.edu/brainstorm/Tutorials.

PR #168, opened 2019-01-30
exchanged size(x,2) with previously defined Norig

sdovgin:
Simplified by replacing code with previously defined variable


ftadel:
Any update on this topic?

PR #173, opened 2019-02-08
Raise error if some channels are duplicated

thomas-vincent:


thomas-vincent:
This happens in some faulty nirs data files.

The duplicated channels are actually renamed automatically by brainstorm by appending "_01", "_02" , ..., eg "S1D1WL830_02".
Maybe brainstorm should raise an error instead?

ftadel:
Some clinical EEG setups actually do save multiple channel names with the same name... So far, I've only seen this for technical tracks, so it's not a big problem.
It is not possible to handle this in Brainstorm, because there are many functions that identify the channels by their names. But there is no fundamental problem with having an acquisition devices that identify channels by their indices only, and allow multiple identical labels.

It's better to leave more flexibility when importing the file into Brainstorm, these issues can maybe fixed manually once imported in the software (with maybe a warning mentioning that there is something that needs attention). But maybe this doesn't make sense in the context of fNIRS.

Let me know when you want me to merge this pull request.

thomas-vincent:
Thanks for the clarification about EEG sensor naming.

For NIRS, duplicates should not be allowed since the channel naming has implication in interpreting the signal. For instance, all recorded optode pairs must have the same number of measures.

So we can leave this check only for NIRS reading and go with this PR.

Thanks

Issue #144, opened 2018-11-29
Filter specifications: Notch and and band-stop


ftadel:
Hossein, Richard:
- Add a button "View filter response" to the notch and band-stop processes 
- Create "transient" events in the filtered files, as it was done for the band-pass filter
- Update the section "Filter specifications" of the tutorial: http://neuroimage.usc.edu/brainstorm/Tutorials/ArtifactsFilter#Filters_specifications_.5BTODO.5D

Francois: 
- Update section "Apply a notch filter" to add the transients in screen capture and text (after the documentation of the notch filter)

HosseinShahabi:
I want to draw your attention to the current band-stop filter in Brainstorm.

This filter is an 8 order Butterworth filter which mostly likes a wideband notch filter than a bandstop (1-bandpass). I've attached the magnitude and impulse response of the filter (designed for a bandstop [.2 .4]). 
![image](https://user-images.githubusercontent.com/45434165/50183542-48860180-02c7-11e9-84ed-875a0dd15e2a.png)
![image](https://user-images.githubusercontent.com/45434165/50183556-53409680-02c7-11e9-9a54-5a8acb2468d9.png)

The edge effect based on 99% energy is around 20 samples for this case. Please note that the edge effect will be doubled since we use "filtfilt". Here is my question: 
Do we want to keep this configuration or change it?

HosseinShahabi:
Dear @ftadel 

I edited the process_bandstop by adding necessary codes and filter response panel. 
I will PR the code (Single function in this case). Once you approved the code, I'll edit the Tutorial. 
![image](https://user-images.githubusercontent.com/45434165/50622490-6441f800-0ec2-11e9-9f09-720bd5e33e7f.png)

The process_notch needs further work since we are modifying the algorithm.

HosseinShahabi:
Both are set. 
https://github.com/brainstorm-tools/brainstorm3/pull/158

ftadel:
Have you modified the code for the notch and bandstop filters?
If so, it would be good to discuss it with @sbaillet @richardmleahy @jcmosher and @dimitriospantazis before pushing them to the Brainstorm distribution.
To be added to the agenda of a biweekly meeting?

sbaillet:
that was indeed the plan.

Best wishes,

Sylvain.
-
Sylvain Baillet, PhD
Professor, Montreal Neurological Institute
McGill University
> more<https://goo.gl/591JdQ>

[...]

PR #177, opened 2019-02-20
Filtering

HosseinShahabi:
I Updated the process functions so there is a single Display function in process_bandpass and it is called in two others. 
Hope you can merge it this time.

### DIFF
--- a/toolbox/connectivity/bst_connectivity.m
+++ b/toolbox/connectivity/bst_connectivity.m
@@ -493,12 +493,12 @@ for iFile = 1:length(FilesA)
             for iBand = 1:nFreqBands
                 % Band-pass filter in one frequency band + Apply Hilbert transform
                 if isConnNN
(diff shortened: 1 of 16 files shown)


ftadel:
It seems to work beautifully! Thanks for this great contribution!

I just renamed the older options, and made it possible to call process_bandpass with the old parameter 'bst-hfilter': https://github.com/brainstorm-tools/brainstorm3/commit/7c1c47d1431fa485430ed3128c5a0d5cd489bd2f

Since your last edits there is something wrong with the tutorial page - this editor is not really stable and can be very annoying sometimes... I'll try to fix it without losing your changes.

ftadel:
Ok, I fixed the tutorial page. I think the issue is that you were trying to include the contents of a github page in a <DIV> block: this doesn't work.

Can you please also update the bandpass and bandstop sections of the specifications?
Please close the related issue when done: https://github.com/brainstorm-tools/brainstorm3/issues/144

Thanks!

PR #178, opened 2019-02-20
Tucker davis support

mpompolas:


mpompolas:
If there is a git ignore again for mex files, just remove it.
This pull request is just for Tucker Davis

PR #180, opened 2019-03-09
NWB continuous/epoched

mpompolas:
**Added trial support for the NWB file format.**

Importing / navigation of trials
Events assignment
Transition between continuous and epoched signals


mpompolas:
Dataset in case you need to play around with:
https://drive.google.com/file/d/1ceoaieEpn7NVmU36o-yRTWwJFYjjWdJQ/view

ftadel:
You keep on posting new commits to this PR, so I guess this is still work in progress...
Let me know when this is ready for review and merging.

mpompolas:
The trials is ready for merging.
The .nwb file format is a work in progress so they update things daily and I just follow it. Since you haven't merged it yet, I kept working on this fork.

Go ahead with the merge of this continuous/epoch commit (as far as I can tell everything works at this point), and I'll make another fork for keeping the importer up-to-date with their newest changes.

ftadel:
Some comments:

1) out_figure_movie: Your proposed fixed does not work when there is only one figure.
I fixed it in a different way: https://github.com/brainstorm-tools/brainstorm3/commit/9fa8d699c98216dfe72c5470805daf15875e5ab1
Please include my change in your branch.

2) You created a new type of "device" (based on the channel file name). I understand that you did this for offering the popup menu "Switch epoched/continous", but I don't think this is a good solution. This "device" tag is intended to define different types of acquisition systems, not different file formats. 

For the CTF file format, there is a one-to-one match between the "device" and the "file format". I test improperly for the device instead of the file format in tree_callbacks.m, but it never caused any problem because of this equivalence between the two. 
The NWB format does not correspond to any particular device, so we should not mix the two. The channel file is named "NWB channels" so you could test for the Comment of the channel file instead, to define if you should display the menu "Switch epoched/continous".
Files to modify: bst_get.m, channel_detect_device.m, tree_callbacks.m

Thanks!

mpompolas:
Thanks Francois,
I will look into them and keep you posted

mpompolas:
Done. Let me know if you need anything else

ftadel:
Right now I guess it is not possible to install the toolbox.
This issue needs to be fixed first: https://github.com/brainstorm-tools/brainstorm3/issues/181

ftadel:
After restarting Matlab, the path to the NWB library was not added to the Matlab path anymore.
I fixed this in: https://github.com/brainstorm-tools/brainstorm3/commit/280817699704698d659468d6174bcd9f81832c12
I also cleaned up a bit the code (the other edits have no impact on the way the code works)

PR #183, opened 2019-03-14
FIX: make OpenMEEG 2.4 work with squid files containing multiple integration points

agramfort:
this should fix the pb with openmeeg 2.4

cc @massich @papadop


ftadel:
I pushed something very similar more or less simultaneously: https://github.com/brainstorm-tools/brainstorm3/commit/7e463aa1eff95a86c0f3e6208a364ed2ebdf99f3#diff-9e590703ce3a61a47c365d821b1164e6

Thanks for the quick solution!

PR #186, opened 2019-03-21
Added option to paste list of files in process box

martcous:
This was requested at our workshop last week.
Similarly to how we can copy the list of files from the process box (or boxes), you can now paste the list of files you have from the clipboard. This is for people processing different groups of files often, they can quickly copy and paste to change the group they are now processing. They can also reuse scripts by just copying the top of their scripts where the sFiles and sFiles2 variables are defined.

processbox_paste

There might be better ways to implement this but I know it would be already very useful for some users in this form.

### DIFF
--- a/toolbox/gui/panel_nodelist.m
+++ b/toolbox/gui/panel_nodelist.m
@@ -33,7 +33,7 @@ function varargout = panel_nodelist( varargin )
 % For more information type "brainstorm license" at command prompt.
 % =============================================================================@
 %
-% Authors: Francois Tadel, 2010-2017
+% Authors: Francois Tadel, 2010-2017; Martin Cousineau, 2017-2019
 
 eval(macro_method);
 end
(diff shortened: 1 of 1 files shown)


ftadel:
Additional suggestions (quite useless, only if you want to take a break from working on something else :)
- Add the CTRL+C / CTRL+V keyboard shortcuts
- Also add the possibility to input plain lists of files in various formats
  - {'Subject01_copy/SZ1/data_block001.mat', 'Subject01_copy/SZ1/data_block001.mat'}
  - Subject01_copy/SZ1/data_block001.mat
Subject01_copy/SZ1/data_block002.mat
  - 'Subject01_copy/SZ1/data_block001.mat'
'Subject01_copy/SZ1/data_block002.mat'

martcous:
I implemented your second suggestion here (had some time in a meeting): https://github.com/brainstorm-tools/brainstorm3/commit/bfbb7a0259c89f9f893c266c87b0233918dfa728

PR #187, opened 2019-03-26
Improved channel_average

Moo-Marc:
Improvements:
1. Used to give an error if number of channels didn't match.  It now deals properly with varying channels (matches them by name), with the options of keeping either all channels (default), only common channels, or those from the first file (similarly to bst_avg_files).
2. Averaging locations caused "shrinkage" towards origin.  Now MEG channels are averaged based on the dewar=>native transformations, separately averaging the "brain origins" (based on the cortex surface if available) and the head orientations.  EEG channel locations are still simply averaged, but with a correction to avoid "shrinkage".
3. There was a bug in averaging orientations: they were not normalized.  Fixed.

Following this, process_average should also be modified to always match channels by name.  This is only an option (why?) and only for files that contain RowNames (tf and matrix types).  I think this can easily lead to errors in group averaging.  There should at the very least be a warning or error if channels don't match and the code can't deal with it.  Now it just averages by channel index.

### DIFF
--- a/toolbox/db/db_combine_channel.m
+++ b/toolbox/db/db_combine_channel.m
@@ -1,5 +1,5 @@
 function [isNewFile, Message] = db_combine_channel( iSrcStudies, iDestStudy, UserConfirm, NoWarning )
-% DB_COMBINE_CHANNEL: Create a channel file for a study that do not have one yet. 
+% DB_COMBINE_CHANNEL: Create a channel file for a study that does not have one yet. 
 %
 % USAGE:  db_combine_channel( iSrcStudies, iDestStudy, UserConfirm, NoWarning )
 %         db_combine_channel( iSrcStudies, iDestStudy, UserConfirm )
@@ -19,7 +19,7 @@ function [isNewFile, Message] = db_combine_channel( iSrcStudies, iDestStudy, Use
 % DESCRIPTION: 
 %     - Create a channel coherent with other input studies.
 %     - Do not replace any previous channel file.
-%     - If more than one source channel file, the channels locations are averaged.
+%     - If more than one source channel file, the channel locations are averaged.
 
 % @=============================================================================
 % This function is part of the Brainstorm software:
@@ -61,6 +61,7 @@ end
 % Get protocol directories
 ProtocolInfo = bst_get('ProtocolInfo');
 ChannelMats = {};
+iAvgStudies = [];
 
 %% ===== LOAD CHANNEL FILES =====
 for i = 1:length(iSrcStudies)
@@ -75,6 +76,7 @@ for i = 1:length(iSrcStudies)
     ChannelFile = bst_fullfile(ProtocolInfo.STUDIES, sStudy.Channel.FileName);
     % Load channel file
     ChannelMats{end + 1} = in_bst_channel(ChannelFile);
+    iAvgStudies(end + 1) = iStudy;
 end
 
 % If no valid source channel files
@@ -83,7 +85,7 @@ if isempty(ChannelMats)
 end
     
 %% ===== COMPUTE CHANNEL MEAN =====
-[MeanChannelMat, Message] = channel_average(ChannelMats);
+[MeanChannelMat, Message] = channel_average(ChannelMats, iAvgStudies, 'all');
 % An error occurred 
 if isempty(MeanChannelMat) 
     if ~NoWarning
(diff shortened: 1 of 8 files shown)


ftadel:
I'm not sure how to validate what you've done here. I guess you've already tested this thoroughly for data types (MEG, EEG, SEEG, ECOG, NIRS)?
Should I just trust what you've done here and merge the PR?

> Following this, process_average should also be modified to always match channels by name. This is only an option (why?) and only for files that contain RowNames (tf and matrix types). I think this can easily lead to errors in group averaging. There should at the very least be a warning or error if channels don't match and the code can't deal with it. Now it just averages by channel index.

Historically, there was no checks, no naming of the channels. When the file types timefreq and matrix appeared, there was the new need to average by "RowName" because TF files have a typically different lists of channels (bad channels are excluded). 
This is an option and should not be enforced because you might have different names for matching signals, for example when averaging the same scouts time series coming from different files ("scoutname @ filename").

I don't recommend going through the editing of the complicated bst_avg_files for matching signals by names, since we already have a solution that does this in a more permanent way: the process "Standardize > Uniform list of channels" (process_stdchan.m).  

But indeed, what is missing is a test for the list of channel names in process_average and maybe channel_average.m. Something that would return a warning if the list of channels do not match (different ordre or different selection), suggesting the use of process_stdchan.

In practice, it won't happen in regular studies with recordings from one center only to have data files with the same number of channels but a different order or a different selection. But it might happen in long-term longitudinal or multi-centric studies.

Moo-Marc:
Hi François

Thanks for your comments.  After thinking a bit more, this is not ready for merging.  Not until channel_average has at least a channel check for data inputs.  I'll have a look.

> we already have a solution that does this in a more permanent way: the process "Standardize > Uniform list of channels" (process_stdchan.m)

Too bad I didn't know about that.  It seems to do similar things that I did, though it affects all the input studies.  I'll have a closer look to compare with what I did.

> I guess you've already tested this thoroughly for data types (MEG, EEG, SEEG, ECOG, NIRS)?

I tested CTF MEG + EEG with process_average imported data files, and another user tested with timefreq on channels, from multiple users from two sites.  Regarding MEG, as long as all data formats have a "dewar=>native" transformation, then it should work.  Is that always the case?  Regarding all other types, they are treated the same way.  I'm happy to do a little more testing if data in other formats are available.  Is there anything specific you'd like me to try?

> for example when averaging the same scouts time series coming from different files ("scoutname @ filename")

I see.  If that's the only case though, we could match the scoutnames (and ignore the rest).

> In practice, it won't happen in regular studies with recordings from one center only to have data files with the same number of channels but a different order or a different selection.

It may not be common but can definitely happen in regular studies.  Bad channels can change during MEG system servicing.  One sensor could be fixed and another one fail.  But even if the numbers change, I think from a user perspective it would be much nicer not to have to standardize all the inputs, and deal with varying channels on the fly when averaging.

Moo-Marc:
This last commit...
1. makes sure there is an error either if files are skipped due to different channel numbers (was only a warning before), or if the numbers match but there were different channels;
2. fixes nAvg to match the actual number of files averaged (included skipped files before which was misleading);
3. only uses averaged files to generate a new channel file;
4. keeps the average channel file intact if only one input study, otherwise removes projectors and a few history tweaks.

I still want to make a few tweaks to process_stdchan:
1. I think we should remove projectors if some channels are removed.  Simply removing the channel rows doesn't preserve the "effect" of the projector.  They'd have to be recomputed with the reduced channel set if still needed. 
2. MegRefCoef can be kept if references didn't change (even if other channels change).
3. (minor) Some errors should probably be 'info', when there's nothing to do because the files are already "standardized".

Final note: GetUniformRows actually only keeps the scoutnames (removes "@ filename").  Are there other cases where the same signals might have different row names?

Moo-Marc:
Ok, I've finished what I wanted to do.  Probably needs a bit more testing though.  @Martin, can you help me with that?  Do we have data from other modalities or systems that we can try averaging?

ftadel:
> Do we have data from other modalities or systems that we can try averaging?

You should test at least: MEG (Elekta, CTF), EEG, SEEG, ECOG, NIRS.
You can find examples for everything in the the tutorials.  
(Except for ECOG, I haven't prepared the dataset yet, I will hopefully before the end of next week, but not sure - Otherwise if you want this faster, take some EEG data, change the type of the channels to ECOG, move the electrodes closer to the brain, and here is your ECOG test dataset :)

Moo-Marc:
I'm going to close this.  I don't have time to test more right now and looks like I had made other changes that are not here.  So I'll sort through it later and open a new pull request.

PR #191, opened 2019-04-05
new process swap_headcoils

Moo-Marc:
New little process for fixing CTF MEG data when the head coils were swapped by mistake.  Unlike just manually editing the .hc file, this fix allows using the rest of the recently added processes related to head motion.  If we don't want it in the main distribution, we can remove process_swap_headcoils but we should keep the other changes so that it can still work.

### DIFF
--- a/toolbox/process/functions/process_adjust_coordinates.m
+++ b/toolbox/process/functions/process_adjust_coordinates.m
@@ -1,27 +1,27 @@
 function varargout = process_adjust_coordinates(varargin)
-% PROCESS_ADJUST_COORDINATES: Adjust, recompute, or remove various coordinate transformations.
-
-% @=============================================================================
-% This function is part of the Brainstorm software:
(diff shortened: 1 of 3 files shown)


ftadel:
I reorganized a bit the processes, let me know if you think this can be improved.
https://github.com/brainstorm-tools/brainstorm3/commit/d956ee74874ad05cfdab8138e3f73da713f9b8f0

Moo-Marc:
Seems ok.  Honestly, I find it difficult to classify many of these processes.  Of course, we should avoid moving them too often to confuse the users.  And we'll have to update the tutorials.

ftadel:
Could you update the tutorials for the processes you wrote?
I’ll take care of the other ones.

I think I will also split the events group in two because the list is now too large to be displayed on small screens... but this requires a lot of tutorial editing

Moo-Marc:
Done! (I just have the head motion one)

PR #198, opened 2019-04-26
delete folder prealocated for protocol unzipping

juangpc:
Hi, 

Context: when loading a protocol from a zip file, the code creates a folder, starts unzipping, but if there is a problem the created folder (and crappy files in it) are left there. 
Patch: if there is aproblem delete the created folder.

Code follows this scheme:
1. Create folder for the protocol. toolbox/io/import_protocol.m [line:69].
2. Call Unpack class (java/brainstorm.jar/org/brainstorm/file/Unpack.java). Which starts unzipping, creating all necessary files, etc... on the fly.
   2a. if there is no problem while unzipping, folders are detected and finally, the protocol is then loaded.
   2b. If there is a problem while unzipping, the FileOutputStream (Unpack class line 70). The exception is caught, the process is interrupted and a false status flag is set (Unpack class line83).

image

Then Matlab checks the error (toolbox/io/import_protocol.m [line 76]). The message is processed and a return is reached.

Commit adds a 'delete folder' just before the return is reached.

A more "elegant' solution would be to unzip into a temp directory and only if there was no problem, create the folder in the db folder. But this also works.

### DIFF
--- a/toolbox/io/import_protocol.m
+++ b/toolbox/io/import_protocol.m
@@ -75,6 +75,7 @@ end
 isOk = org.brainstorm.file.Unpack.unzip(ZipFile, ProtocolDir);
 if ~isOk
     bst_error('Could not unzip file.', 'Load protocol', 0);
+    rmdir(ProtocolDir,'s');
     return
 end
 
--- a/toolbox/io/import_protocol.m
+++ b/toolbox/io/import_protocol.m
@@ -88,7 +88,7 @@ if ~isOk
     file_delete(tmpProtocolDir, 1, 3); %in case of error during unzip.. rollback
     return
 end
-
+% Unzip went well. Go ahead and copy to db
 file_copy(tmpProtocolDir,ProtocolDir);
 file_delete(tmpProtocolDir,1,3); %clean up
 
--- a/toolbox/io/import_protocol.m
+++ b/toolbox/io/import_protocol.m
@@ -88,11 +88,16 @@ if ~isOk
     file_delete(tmpProtocolDir, 1, 3); %in case of error during unzip.. rollback
     return
 end
+bst_progress('stop');
+
 % Unzip went well. Go ahead and copy to db
+bst_progress('start', 'Load protocol', 'Copying files to database...');
 file_copy(tmpProtocolDir,ProtocolDir);
 file_delete(tmpProtocolDir,1,3); %clean up
+bst_progress('stop');
 
 %% ===== DETECT FOLDERS =====
+bst_progress('start','Load protocol','Parsing protocol info...');
 % Detect anatomy and datasets folders
 subjectFile = file_find(ProtocolDir, 'brainstormsubject*.mat', 3);
 studyFile   = file_find(ProtocolDir, 'brainstormstudy*.mat',   4);
--- a/toolbox/io/import_protocol.m
+++ b/toolbox/io/import_protocol.m
@@ -88,16 +88,14 @@ if ~isOk
     file_delete(tmpProtocolDir, 1, 3); %in case of error during unzip.. rollback
     return
 end
-bst_progress('stop');
 
 % Unzip went well. Go ahead and copy to db
-bst_progress('start', 'Load protocol', 'Copying files to database...');
+bst_progress('text', 'Copying files to database...');
 file_copy(tmpProtocolDir,ProtocolDir);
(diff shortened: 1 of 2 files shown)


toolbox/io/import_protocol.m
```diff
@@ -88,11 +88,16 @@ function import_protocol(ZipFile)
     file_delete(tmpProtocolDir, 1, 3); %in case of error during unzip.. rollback
     return
 end
+bst_progress('stop');
+
 % Unzip went well. Go ahead and copy to db
+bst_progress('start', 'Load protocol', 'Copying files to database...');
```
martcous: You do not need to stop and start again just to change the text of the progress bar. Use bst_progress('text', 'Copying files to database...'); instead.

juangpc: Thanks! Will do!
```suggestion
bst_progress('start', 'Load protocol', 'Copying files to database...');
```

juangpc:
Hi all,

Just wanted to describe some more the thing here.
tldr; This is an unsolicited feature dev for bst. It helps solve an issue I encountered while going through the tutorials. I completely understand it might not be a good addition for many reasons, therefore if this is the case, I'll understand it completely. No problem at all. I would, however, very much appreciate, in any case, a comment on why or why not, the problems with the code, etc... I'm basically trying to learn here.

-- 
As detailed earlier, when loading a protocol from a zip file. 
 -if there is a problem with the zip file itself. the an exception in the java class Unpack was catched but a folder with the name of the protocol remained basically as trash in the db folder.
 -if the zip file is OK but the data inside was not correct or complete, the process was interrupted in import_protocol.m (in DETECT FOLDERS block). But a folder with the name of the protocol remained basically as trash in the db folder.

With these PR both problems are solved (?). 

To recreate the first scenario create a text file, change extension to zip and try to load it as a new protocol.
To recreate the second scenario try to load the sample_introduction.zip file downloading it from the project data page, but loading it using the "load from zip file" option. (There you go, now you know how idiot I can be... )

Thanks.

ftadel:
This is a good idea to delete the folder if the protocol could not be unzipped properly.
However, I don't see any reason for which you would not unzip directly to brainstorm_db. It is not a threat to stability if there is some extra garbage in there, and if you nicely delete the folder after a failure, there is no problem anyway.

In general, prefer moving files than copying and deleting. If the protocol is huge (which is typically the case), it would require the double amount of disk space and a lot of extra time for the copy (making the whole process 2x longer).

Additional potential problems: The temp folder is for most users in $HOME/.brainstorm/tmp, which on university or hospital networks is commonly on a network drive with limited space available, while the the brainstorm_db folder is most likely on a local drive or a USB drive. First it could crash for a lack of space in $HOME. Then, even moving the files instead of copying/deleting would cause heavy copy over the network, which can cause more problems (can be long, prone to failures and rights issues). If the user doesn't have the right to create the final folder in brainstorm_db, the whole thing would be unzipped but could not be moved to the final destination.

ftadel:
Couldn't you simply add another "file_delete(ProtocolDir, 1, 3);" immediately after the first isOk test (line 77) ?

juangpc:
> Couldn't you simply add another "file_delete(ProtocolDir, 1, 3);" immediately after the first isOk test (line 77) ?

I tried that. It was my first approach, but it doesn't work.
If there's a problematic zip file, an exception is thrown within the java Unpack class. Please see my first comment. Then JavaException.m is called and then bst_error handles more or less everything (closes the progress bar (bst_progress('stop')), displays a new java dialog as error message, and so on.

So the rest of the import_protocol.m does not actually get to be executed. The way I see it the options are: 
 - Modify Unpack class.
 - Modify the handling of the exception (JavaException.m) so that it knows what's going on and deletes the created folder. Although this will also include modifying Unpack class, because right now the only info JavaException receives is:
**Java exception occurred: java.lang.NullPointerException at org.brainstorm.file.Unpack.unzip(Unpack.java:87)**
 - Create a temp folder (and the temp directory seems a good place).
Or, leave it as just a crap folder inside the db folder. The next time you attempt to create a protocol with that same name it will bug you about an existing folder. 
 - Create a temp folder in the db_folder with a random name and if everything is good, rename the folder. The only problem with this option is that if there is a problem you end up accumulating temp folders.
I do like the feature of brainstorm that every time it starts it empties the tmp folder.
Let me implement this last option.

PR #199, opened 2019-04-30
Events structure: Removed "samples", added "channels" and "notes"

ftadel:


martcous:
What will be the use of channels and notes?
Also, I hope this is not a stupid question but with times being float and samples being integers, why not only keep samples rather than times?

ftadel:
> What will be the use of channels and notes?

Channels: list of channels attached to an event. Useful in clinical context with marking epileptic spikes of HFOs for instance - clinicians want to mark them for one or multiple channels. Ask Jeremy for more info.
Notes: additional comments for an event. Also for clinical review (eg. to indicate what is the behavioral response to an electric stimulation of a given pair of SEEG electrodes)
image

> Also, I hope this is not a stupid question but with times being float and samples being integers, why not only keep samples rather than times?

Indeed, I could have done the contrary and kept the samples instead.
Semantically, what was called "samples" was not corresponding to samples (ie. 0 or 1 being the beginning of the file), but the time multiplied by the sampling frequency and rounded. In most cases, we were converting times into samples and not the contrary.
Time values are also easier to import/export because they are independent from the sampling rate, and definitely more intuitive to manipulate.

juangpc:
Hi, this is interesting!
Maybe (probably) I'm wrong, but would appreciate if someone could clarify. I went into fiff reader script (~/toolbox/io/in_fopen_fif.m) , and the time vector is built upon sample information. So in the raw file there is no array with times, just samples. We build the time array while reading the file: 

image

And then while manipulating stuff we go the other way around. So we do: sample->time->sample. Am I right?
I do consider as well, that it is not a big deal, that it is a good idea to keep both, but up until now, I didn't know that all the time-related information was built upon sample info (which is built during the file-reading process).

Could this generate a latency problem of +-1sample due to the rounding of (floating point) time to (integer) sample number?

toolbox/io/in_bst_data.m
```diff
@@ -136,6 +136,17 @@
...
 
+% ===== FIX EVENTS STRUCTURES =====
+% Imported file
+if isfield(DataMat, 'Events') && ~isempty(DataMat.Events)
+    DataMat.Events = struct_fix_events(DataMat.Events);
+end
+% Link to raw file
+if isfield(DataMat, 'F') && ~isempty(DataMat.F) && isstruct(DataMat.F) && isfield(DataMat.F, 'events') && ~isempty(DataMat.F.events)
+    DataMat.F.events = struct_fix_events(DataMat.F.events);
+end
```
martcous: Right now the fixed events are not saved so the user will get the warning every time they open an old raw file until they make a modification to it. I would personally add a isModified return flag to struct_fix_events() and save the raw file if it is modified.

toolbox/inverse/bst_inverse_linear_2018.m
```diff
@@ -163,7 +163,7 @@
     return
 end
 
-fprintf('\nBST_INVERSE (2018) > Modified Apr 2019\n\n');
+fprintf('\nBST_INVERSE (2016) > Modified Feb 2018\n\n');
```
martcous: Overwriting previous change here

Moo-Marc:
Other than the PR I just submitted to François's fork, I'm ok with this.

Moo-Marc:
> Could this generate a latency problem of +-1sample due to the rounding of (floating point) time to (integer) sample number?

Not really.  My understanding is that event times always correspond to a specific sample, so any precision error should be tiny and rounded properly.  

On the other hand, when comparing times, precision may be an issue, whereas for integer samples it wasn't.

ftadel:
Thank you all for your reviews and comments!

Some more toughts about about this topic:
The main problem here was that the two fields .times and .samples were redundant (one was always based on the other) but not always the same field was used as a reference. It is OK as long as we are always 100% sure the two fields are mirrored. But it becomes dangerous when the two fields are set independently with non-matching values, or using a different convention, in these case the outcome could be random.
I realized it was urgent to fix this when reviewing some recent code from Konstantinos: he was assuming (legitimately) that "samples" had to be positive numbers, and was cutting the events outside of the time definition with max(samples,0). This is valid when working with raw signals in file formats that always start at zero, but not for data epoched around an event of interest, where the baseline corresponds to negative times and therefore negative "samples". 
I thought that with an increasing number of contributors, interpretation "errors" like this one would become more and more frequent and it was better to fix the design error before this gets out of control and propagates nasty hidden bugs.

> Could this generate a latency problem of +-1sample due to the rounding of (floating point) time to (integer) sample number?

We should be careful at always set rounded values in the times fields:
times = round(times .* sampling_frequency) ./ sampling_frequency
But even without this rounding, I doubt that we would ever accumulate so many rounding errors that we would reach a full sample.

> On the other hand, when comparing times, precision may be an issue, whereas for integer samples it wasn't.

Yes, for this we should be more careful. We have two possible approaches:
1) Either we convert to samples and do all the comparisons in samples (I kept this logic in many event management processes).
2) or we truncate the all time values before comparisons at a decimal that preserves the time precision but removes the floating point epsilons (<1e-11) => I used 10e-9s because I doubt we will ever go beyond a 0.00000001s precision with this software. See for example: https://github.com/ftadel/brainstorm3/blob/master/toolbox/process/functions/process_evt_merge.m#L156

toolbox/inverse/bst_inverse_linear_2018.m
ftadel: Thanks for spotting this. Fixed: https://github.com/brainstorm-tools/brainstorm3/pull/199/commits/1def1c333880dd163f782bf38143878415fad344

toolbox/io/in_bst_data.m
ftadel: What about this?
https://github.com/brainstorm-tools/brainstorm3/pull/199/commits/ed8a195e6652cb4522b9b39e7d4c3d8fd6ec33cc

martcous: Yep, working fine!

Edouard2laire:
Hello, 
When are you planning to merge this PR ? Nirstorm use "samples" in several process. I think we need to fix this before merging this. cc @thomas-vincent 

Sincerely, 
Edouard

ftadel:
@Edouard2laire @thomas-vincent
We'll do it as soon as everybody tested their code. 
Can you let us how much time you'd need?

ftadel:
Nope... I'm sorry, I pushed it to the wrong fork... 
Hummmm... do you mind if it stays here? 

Otherwise I have to commit an undo commit, commit it to the brainstorm-tools master, then merge back with my fork. I'm not even sure I know how to do this without adding more mess... :)
Any simpler solution?

martcous:
> Otherwise I have to commit an undo commit, commit it to the brainstorm-tools master, then merge back with my fork.

That is the proper way to do it, and is pretty standard in terms of git workflow. But if you are not confident in doing so it's probably fine leaving it here if the change is not urgent.

mpompolas:
I run all the importers and the functions from my toolbox.
Small corrections in: https://github.com/ftadel/brainstorm3/pull/2

ftadel:
I will merge this PR tomorrow if this is OK with everybody

ftadel:
@Edouard2laire @thomas-vincent
Last chance to fix your code before the merge

mpompolas:
so

juangpc:
178! 0_0

thomas-vincent:
I think we could also robustify the conversion from times to samples to take into account the origin.
For example in [process_split_raw_file.m](https://github.com/brainstorm-tools/brainstorm3/blob/e4012c90662a765a6788ef7f57b502df0bf0159d/toolbox/process/functions/process_split_raw_file.m#L89):
` sEvent.samples = round(sEvent.times .* sFile.prop.sfreq);`
assumes that the reference temporal axis always starts at t = dt so that the first time point maps to index = 1.

A safer way would be:
`samples = round((times_to_convert - t_origin_ref) .* freq_ref) + 1;`

Moo-Marc:
No such assumption.  It's just that **by definition** the sample numbers are always round(times*sfreq), and **not** indices into arrays.  In particular they can be 0 or negative.  From what I understand, it's always been this way, which is why it was safe to remove them completely.

thomas-vincent:
I would rather say **by convention** than **by definition**.
I was not questioning the removal of the .samples field but their meaning.
If they're never meant to be used for array indexing then it's fine. I wasn't aware of this.

However, there may be this case in [bst_events_2track.m](https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/io/private/bst_events2track.m):
```matlab
% Get list of samples to set to this event
    samples = [];
    for iOccur = 1:size(sFile.events(i).times, 2)
        occ = round(sFile.events(i).times(:,iOccur) .* sFile.prop.sfreq);
        % Single events
        if (size(occ, 1) == 1)
            samples = [samples, occ + (1:evtDuration)];
        % Extended events
        else
            if (occ(1) ~= occ(2))
                samples = [samples, occ(1)+1:occ(2)];
            else
                samples = [samples, occ(1):occ(2)];
            end
        end
    end
```
Then samples are used as array indexes, though time origin it not taken into account before:
```matlab
track(samples) = evtVal(i);
```

Moo-Marc:
Indeed, that's what I meant.
And I agree time(1) should be taken into account in this and similar cases.  Good catch.

PR #201, opened 2019-05-06
Added database structure updater and re-enabled new fiber connectivity features

martcous:
My new fiber connectivity features were disabled because they required changing the database structure. I added an automatic way to update the database structure when you first open Brainstorm or load an external protocol/subject. Previously exported subjects did not include their database version, so they are assumed to have version 4.0 (which hasn't changed since 2013). In the unlikely event that a user accesses a new database with an old version of the software, they are prompted to update Brainstorm. This PR also re-enables the new fiber connectivity features.


toolbox/core/bst_startup.m
```diff
@@ -170,7 +170,7 @@ function bst_startup(BrainstormHomeDir, GuiLevel, BrainstormDbDir)
 % Get user database file : brainstorm.mat
 dbFile = bst_get('BrainstormDbFile');
 % Current DB version
-CurrentDbVersion = 4;
+CurrentDbVersion = 4.1;
```
ftadel: Call it "5", we won't have that many versions...

juangpc:
Hi, related to this, 

@ftadel a couple of biweekly meeting ago, you mentioned that you had in mind changing the db structure. I'd be more than happy to help. During the last week I've been learning how it works and *apart from screwing up* I have a couple of ideas. If you want to discuss it whenever you find it a good time, let me know.

ftadel:
The database system currently in use was designed by @sbaillet more than 15 years ago and is not well adapted to the future objectives of the project (multi-user databases shared over networks).
Redesigning it completely is a project that @sbaillet and @martcous have for the next few years. It is not part of the immediate plans, but it will be discussed at some point at our biweekly meetings. 

This PR was some sort of temporary patch (for a few years) to add a new file type that Martin needed to handle, and maybe a few other things we may need in the future before we switch completely to something else.

@sbaillet @martcous: I don't know if you already want to start discussing this topic?

PR #202, opened 2019-05-07
EDF+D discontinuities now handled by in_fopen_edf

martcous:
With @Moo-Marc's  help, I dug into the EDF specs to properly handle EDF+Ds. When Brainstorm detects that an EDF file is a discontinuous file, the user is warned and prompted that Brainstorm can try to detect discontinuities and fix event timings. This flags discontinuities, both in the command window as a warning and as a new event, and adds the timing difference to subsequently found events so that their final times are correct.

Note that we found another potential problem as from what we understand from the specs, you are allowed to specify multiple events of the same timings at once (i.e. "+45s|simultaneous event 1|simultaneous event 2" would work the same as +45s|simultaneous event 1||+45s|simultaneous event 2"), but it didn't seem worth the effort for now as it required major changes and no user has ever reported an issue (most likely that other software never export events together like that).

I think this is safe to merge as-is since the user is prompted and warned beforehand (a.k.a. use at your own risk).


ftadel:
Good, it will probably help some people.
This format is so poorly designed...

One thing you should change: 
these low-level functions are supposed to run automonously without any user interaction, so that they can run deported on headless systems. If you want to ask a question to the user, you need to check first that you're allowed to display interactive messages.
In in_fopen_edf, ImportOptions is already available. If you can interact with the user, ImportOptions.DisplayMessages is set to 1.
If it is set to 0, you can either report an error or define a default behavior.

PR #205, opened 2019-05-09
Fix bst correlation

juangpc:
Hi, 

I've stumbled upon this bst_correlation.m file. Very nice. 

A brief description of the fix and why. 

1st commit (2add):
 - changed the name of the function to match the file name.
 - changed the use of variable named 'inputs' for 'cfg'
2nd commit (3431):
 - lines 131, 132, also 160,161 and finally 187,188 share the use of 'end' indicating the last element in an array. But the 'end' is inside a function call which can cause problems because the matlab does not have to know if that 'end' is a variable or what. In this case, it relates to the index of the last element one of the dimensions of a matrix (2nd) of which the result of the function call, has to then refer to... (Yes, a bit confusing).
Maybe I'm wrong but I found out that 'end' refers in fact to the number of columns, which is the number of samples, referred to as nTimes in the script. And considering it is already calculated, I decided to use it, instead of 'end'.

With this script, you can test if the 2nd commit is doing more harm than good. My results... differences: zero, zero.

```matlab

nsensors=272;
nsamples=3000;
X=rand(nsensors,nsamples);
Y=rand(nsensors,nsamples);

[c,pval,d]=bst_correlation(X,Y,[]);

cfg.maxDelay=10;
[c_d,pval,d]=bst_correlation(X,Y,cfg);

%imagesc(c) %just to check

c_master = c;
c_d_master = c_d;
%-->switch here to testing branch
[c,pval,d]=bst_correlation(X,Y,[]);
cfg.maxDelay=10;
[c_d,pval,d]=bst_correlation(X,Y,cfg);
%result
sum(sum(c-c_master))
sum(sum(sum(c-c_master)))
```

### DIFF
--- a/toolbox/connectivity/bst_correlation.m
+++ b/toolbox/connectivity/bst_correlation.m
@@ -1,4 +1,4 @@
-function [connectivity, pValues, delays] = bst_correlation_sergul(X, Y, inputs)
+function [connectivity, pValues, delays] = bst_correlation(X, Y, cfg)
 % BST_CORRELATION   Calculate the covariance OR correlation between two multivariate signals, assuming any drift has been removed.
 %
 % Inputs:
@@ -7,7 +7,7 @@ function [connectivity, pValues, delays] = bst_correlation_sergul(X, Y, inputs)
 %   Y                 - Second set of signals, one signal per row
 %                       [Y: B x N matrix]
 %                       (default: Y = X)
-%   inputs            - Struct of parameters:
+%   cfg             - Struct of parameters:
 %   |-normalize       - If true, output correlation
 %   |                   If false (default), output covariance
 %   |                   [F: boolean]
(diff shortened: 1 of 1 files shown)


Moo-Marc:
Using "end" to indicate the last element in that dimension is common Matlab syntax.  A(:, end) is the last column, A(end, :) is the last line.  You actually can't have a variable named "end" as it is a reserved word (also used a lot for ifs, loops, etc).
Cheers

juangpc:
Hi Marc, thanks. Yeah, I use it a lot, too. :smirk: 
But it was Matlab who actually pointed that to me. It is quite an erratic behavior and that's why I didn't mention it initially. Sometimes it happens sometimes it doesn't (and I've no idea what it depends on). But when setting a breakpoint in this file, sometimes it won't let you, because of the use of 'end' in these lines I point out. 
Besides, having nTimes already calculated... and so close... I decided to propose using it.

martcous:
I personally think using "end" is clearer when reading the code than using yet another variable (you know for sure you're refering to the last column), but I guess that's a matter of personal preference. I agree with renaming "inputs", and "cfg" is clear, although I feel like "options" is used more often in the codebase as the variable name of structures of parameters.

juangpc:
Yeah, thanks @martcous for commenting. I do think 'end' is ok too. I would only like to add that I wouldn't want to change someone else's code just for preference. The reasons why I included this change in the pr are> (1) it sometimes(!?) wouldn't let you debug. (2) it is not another variable since I understand it was already calculated. I've tested for a few different cfg and results where exactly the same. Hope I don't introduce a bug.
Oh, and the 'options' choice, I'll use from now on.

PR #209, opened 2019-05-21
Add documentation to specify event fields.

thomas-vincent:
Picking up on #199, here are some additional comments to try and clarify the fields of the event structure.

Especially, whether lists can be empty or not. I found that all lists should never be empty except `reactTimes`. Right?

I would be happy to have you confirmation on these.

### DIFF
--- a/toolbox/db/db_template.m
+++ b/toolbox/db/db_template.m
@@ -437,17 +437,18 @@ switch lower(structureName)
         
     % ==== EVENT ====
     case 'event'
+        % See: https://neuroimage.usc.edu/brainstorm/Tutorials/EventMarkers#On_the_hard_drive
         template = struct(...
-            'label',      '', ...
-            'color',      [], ...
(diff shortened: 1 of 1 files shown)


toolbox/db/db_template.m
```diff
@@ -437,17 +437,18 @@
...
-            'color',      [], ...
-            'epochs',     [], ...      % [list of epochs indices]
-            'times',      [], ...      % [list of time values]
-            'reactTimes', [], ...      % [list of reaction times, when applicable]
-            'select',     1, ...
-            'channels',   [], ...
-            'notes',      []);
-        template.channels = {};
-        template.notes = {};
+            'label',      '', ...      % str, label of the event group. Can be empty.
```
ftadel: Not a good idea to mention this can be empty... It is tolerated because some file formats allow empty labels, but it is not something we can do from the interface.

PR #210, opened 2019-05-22
Added possibility for time series Std to be an interval rather than a scalar

martcous:
Right now if you want to plot a standard deviation / error halo around your time series data, you must populate the Std field with a single value per channel & sample in order to plot a halo perfectly centered around the data, as in [Data - Std, Data + Std]. @mpompolas requested for his new batch of changes for the electrophysiology toolbox to be able to plot a confidence interval around the data, which is not perfectly centered around the data and therefore cannot be as simple as +/- Std.

I added support for an optional third dimension to the Std field for timeseries and when it is present it plots the halo as [Std(:,:,1), Std(:,:,2)] rather than [F(:,:) - Std(:,:), F(:,:) + Std(:,:)]. It seems to be working fine for matrices and data timeseries in both butterfly and column view, but there are probably some cases I missed where it might create issues, let me know!


mpompolas:
Thank you @martcous for doing this so quickly.
The equidistant stds would be ok if the firing follows a Gaussian distribution for each bin of the trials but this might not be true in some experimental designs were the neurons adapt to the repeated stimulation.

martcous:
So you want me to add this dimension to every data types with Stds, not just time series?

ftadel:
This Std field is handled by several functions regardless of the type of type it is.
If you extend the capabilities of this Std field, it has to be compatible with what is already existing (using 4th dimension).

And after doing that, it's very little work to extend it to all the types of files, so why not doing it :)

jcmosher:
We recently were trying to calculate RMS rather than STD (i.e., not remove any mean from the data first). We wrote our own routines. Any interest from others in RMS? May be too custom.

ftadel:
@jcmosher You can store your RMS values in this Std field, no problem.

martcous:
Thanks Francois. I added the lower/upper bound as the 4th dimension of the STD matrix of every data type. Everything should still be functional if you do not have this dimension. Can you help me test this? It seemed to me like it's only displayed on timeseries figures (as a halo), let me know if I missed anything.

PR #211, opened 2019-05-23
Openmeeg memory estimate fix for surface source models

Moo-Marc:
Fixes openmeeg memory estimate for surface source models.  
I wasn't sure how mixed models work, have a look if this makes sense please.

Note that this still doesn't match what openmeeg uses as per this issue: https://github.com/openmeeg/openmeeg/issues/364

### DIFF
--- a/toolbox/forward/panel_openmeeg.m
+++ b/toolbox/forward/panel_openmeeg.m
@@ -179,7 +179,16 @@ function [bstPanelNew, panelName] = CreatePanel(sProcess, sFiles)  %#ok<DEFNU>
         end
 
         % === COUNT MEMORY NEEDED ===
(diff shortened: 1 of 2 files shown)


ftadel:
Wow, this piece of code was indeed totally off...

I fixed it in a different way because in most cases we actually don't know how many dipoles we have... For volume and mixed source models, the grid of dipoles is computed or loaded after this OpenMEEG option panel is displayed. And yes, it was totally wrong even for surface source models :)
https://github.com/brainstorm-tools/brainstorm3/commit/483b1759bb73cf81a38897413af60e412a2b06e5#diff-7b198e1479d30390a900e3b3156fe9e3

Issue #140, opened 2018-11-29
Inverse: Fix the dSPM averaging


ftadel:
April 2015: Scaling of dSPM values for averaged files (ERP/ERF) by the number of trials was added:
https://neuroimage.usc.edu/brainstorm/News#April_2015

July 2018: It was reverted it in this commit:
https://github.com/brainstorm-tools/brainstorm3/commit/4b32f0fc1e3272e27ca45c0d979368324588ff05
It is now documented in this tutorial:
https://neuroimage.usc.edu/brainstorm/Tutorials/SourceEstimation#Averaging_normalized_values

November 2018: Dimitrios suggests to reimplement it again, maybe under a different format. Some important decisions have to be made here, as it impacts dramatically the output results.

ftadel:
Last messages exchanged on this topic.

Dimitrios: 
> Attached slides: [Frontiers dataset - Dimitrios Pantazis.pptx](https://github.com/brainstorm-tools/brainstorm3/files/2628852/Frontiers.dataset.-.Dimitrios.Pantazis.pptx)
>
> A summary of some findings:
> -The |Faces| - |Scrambled| contrast works well for sources
> -Applying z-score after low-pass filtering leads to noisy maps. It should be applied before.
> -I think dSPM maps produce better results that min-norm (more sensitive). (Results should be the same if we apply min-norm + z-score, though I did not test)
> -We may need a better colormap than the red-blue-white
> -I reproduced the MNE paper results reasonably well but with FDR thresholds
> -Cluster size inference failed to produce good clusters (unlike the MNE paper). I used the fieldtrip cluster-size test implemented in brainstorm. It took over 10 hours and I don’t know why the results don’t match the MNE paper ones.
> -Critically, the |Faces – Scrambled| contrast is good for sensors, but does not appear to be good for source maps. In particular, the maps appear ok but the Chi-squared statistic fails miserably and I could not find a way to make it work (everything is significant). I think we seriously need to consider removing it from our recommendations. Which is unfortunate because I suggested it before. It was the only option I could think of to perform statistics on a single condition case (when there is nothing to permute). This means we do not have a statistical procedure for a single condition, but the Chi-squared does not work anyway. In a single condition case, researchers should use condition vs. baseline (thus making it a 2-condition test) and use |condition| - |baseline| for sources.

Sylvain:
> > -The |Faces| - |Scrambled| contrast works well for sources
> 
[...]

PR #212, opened 2019-05-28
Added effective number of averages "Leff" for correct dSPM scaling

ftadel:
This addresses the issue https://github.com/brainstorm-tools/brainstorm3/issues/140

A new field "**Leff**" is introduced, to track continuously the "Effective number of averages" for all the files in the Brainstorm database.  The name of this variable was chosen to match Matti's manual, but also because it sounds a lot better than "eff_nAvg" :)
image

When combining multiple files with effective number of averages `Leff_i` using linear weights `w_i`, the resulting file has: 
`Leff_new = 1 / ( sum_i ( w_i^2 / Leff_i ))`

A short toy example, from the introduction tutorials: 
A=deviant=average of 39 trials, B=standard=average of 193 trials
- Average: (A+B)/2
```
  N=2, w1=1/N, w2=1/N
  Leff = N^2 / (1/Leff_A + 1/Leff_B)) = 4/(1/39+1/193) = **129.8**
```
- Weighted average: (nAvgA * A + nAvgB * B) / (nAvgA + nAvgB)
```
  w1=LeffA/(LeffA + LeffB), w2=LeffB/(LeffA+ LeffB)
  Leff = Leff_A + Leff_B = 39+193= **232**
```
- Difference: (A - B)
```
  w1=1, w2=-1
  Leff = 1 / (1/Leff_A + 1/Leff_B)) = 1/(1/39+1/193) = **32.4**
```
- Weighted different:  A-sqrt(nAvgB/nAvgA)*B
```
  w1=1, w2=-sqrt(nAvgB/nAvgA)
  Leff = 2*Leff_A = 2*39 = **78**
```

The dSPM maps are always computed **without** taking into account this value, so that it remains fully linear, similarly to current density maps and sLORETA values. To obtain correctly scaled dSPM values, one has to call the process "Sources > **Scale averaged dSPM**", as explained in this tutorial: https://neuroimage.usc.edu/brainstorm/Tutorials/SourceEstimation#Averaging_normalized_values

---

@dimitriospantazis @jcmosher @sbaillet :
1) Please review all the code modifications and let me know if you think this matches your expectations, and if you think the method to compute this Leff value is correct in all the funtions (click on tab "Files changed" in this pull request, then click on the little + next to the lines to add comments).
2) Please make sure the online documentation for dSPM remains correct with this definition: https://neuroimage.usc.edu/brainstorm/Tutorials/SourceEstimation#Averaging_normalized_values
3) After calling this process "Scale averaged dSPM", what do we do with the field Leff in the output file? do we leave it unchanged with the risk that the process can be called again on it, or that the file gets re-averaged with other files, or do we set it back to 1? or to 0?
4) For weighted averages: should we compute the actual average based on nAvg or Leff? For now I kept it with nAvg: [bst_avg_files.m](https://github.com/ftadel/brainstorm3/blob/master/toolbox/process/bst_avg_files.m#L480), [process_average_ab.m](https://github.com/ftadel/brainstorm3/blob/master/toolbox/process/functions/process_average_ab.m#L95)
5) I think there are additional functions you may want to fix to benefit fully from this new Leff information:
  - **Noise covariance**: [bst_noisecov](https://github.com/ftadel/brainstorm3/blob/master/toolbox/math/bst_noisecov.m) + [in_bst_data_multi](https://github.com/ftadel/brainstorm3/blob/master/toolbox/io/in_bst_data_multi.m): 
  - **Dipole scanning**: [process_dipole_scanning.m](https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/process/functions/process_dipole_scanning.m#L118), line 118
  - Minimum norm: [bst_inverse_linear_2018.m](https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/inverse/bst_inverse_linear_2018.m)  - Or should we remove all references to nAvg/Leff if these fields are not used anymore there?
  - **Whitened data**: [tree_callbacks/SaveWhitenedData](https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/tree/tree_callbacks.m#L3256) line 3256

---

There are some other functions that I didn't really know how to fix:
- process_pac_*: @SoheilaSamiee ?
- process_pls*: @sbaillet ?


ftadel:
I finished the modifications we discussed yesterday:
1) When trying computing dSPM, the user sees this clickable warning:
image 
image
image


2) The files are now labelled "dSPM-unscaled" by default.
image 

3) After running the process "Scale averaged dSPM", the tag "-unscaled" disappears, the fields "Leff" and "nAvg" remain unchanged and the field "Function" is changed from '**dspm2018**' to '**dspm2018sc**', and :
image
image

4) Modified the following functions to use Leff instead of nAvg:
  - bst_avg_files: The function that averages all the values in Brainstorm
  - process_average_ab
  - bst_noisecov: Computation of the noise covariance
  
5) The functions below still need to be modified: John? Dimitrios?
  - **Noise covariance**: bst_noisecov + in_bst_data_multi:
  - **Dipole scanning**: process_dipole_scanning.m, line 118
  - **Minimum norm**: bst_inverse_linear_2018.m - Or should we remove all references to nAvg/Leff if these fields are not used anymore there?
  - **Whitened data**: tree_callbacks/SaveWhitenedData line 3256

dimitriospantazis:
Hi Francois,

I had a look at your ftadel:master version of brainstorm and the Leff variable and dSPM-unscaled appears to work well. There is one case I want to discuss:

**[For weighted averages: should we compute the actual average based on nAvg or Leff? For now I kept it with nAvg:  ]**
***We should use Leff!

Here is the toy example I used.
DataA = Average of 86 face trials: LeffA = 86, nAvgA = 86
DataB = Average of 83 object trials: LeffB = 83, nAvgB = 83
DataC = Difference of the two averages: LeffC = 42.23, nAvgC = 169
Up to here, everything is fine. But then I computed:
DataD = Weighted average(DataA, DataC): LeffD = 128.23, nAvgD = 255

What weights are used for the computation of the weighted average?
Case1: w1 = LeffA / (LeffA + LeffC) = 0.67 , w2 = LeffC / (LeffA + LeffC) = 0.33
or
Case 2: w1 = nAvgA / (nAvgA + nAvgC) = 0.33 , w2 = nAvgC / (nAvgA + nAvgC) = 0.66

If you use Case 1, then LeffD is computed correctly as LeffD = LeffA + LeffC.
If you use Case 2, then LeffD is not computed correctly.
[In both cases nAvgD is meaningless and this is why I don’t like the variable]

We should be using Case 1 and revise how the weighted average is described. We should not describe it as:
•	Weighted average: (nAvgA * A + nAvgB * B) / (nAvgA + nAvgB)
But rather as:
•	Weighted average: (LeffA * A + LeffB * B) / (LeffA + LeffB)
This should be reflected in the software and tutorials.

In general, I believe ** *all* computations should be performed using the Leff variable and never the nAvg variable**. The latter is either equal to Leff, or it becomes (statistically) meaningless when they are different. This comment also applies for the computation of the noise covariance, whitened data, etc. (John may also want to comment here since he was involved in these functions)

I also propose you revise the pop-up warning message as follows:
The values are not scaled by the effective number of trials any more.


**[After calling this process "Scale averaged dSPM", what do we do with the field Leff in the output file? do we leave it unchanged with the risk that the process can be called again on it, or that the file gets re-averaged with other files, or do we set it back to 1? or to 0?]**
We should not allow any further averages, differences etc after “Scale averaged dSPM” has been called. Maybe replace Leff with 0 to indicate it has been applied?
A potentially controversial topic that needs to be discussed with the team: I suggest that statistics should also not be allowed with the scaled version (E.g. dSPM1 > dSPM2 is deceiving because it does not imply signal1 > signal2; however we can do statistics with dSPM1-unscaled > dSPM2-unscaled). 
Best,
Dimitrios

ftadel:
@dimitriospantazis  
Thanks for the review. I addressed all your requests.
You can update the code and check if it works the way you wanted.

> What weights are used for the computation of the weighted average?

Case 1: Only Leff is used in the averaging. nAvg is kept for backwards compatibility, in case anybody uses it, but it should not be used anywhere for scaling/weighting anymore.

> We should not allow any further averages, differences etc after “Scale averaged dSPM” has been called. Maybe replace Leff with 0 to indicate it has been applied?

Done. And I added and error in bst_avg_files.m to prevent files with Leff=0 to be processed.
This averaging function is used in all the processes that need to compute mean and variance across files, so stats are also forbidden on scaled dSPM as well.

There are still additional reviews and modifications needed from @sbaillet and @jcmosher

ftadel:
closes https://github.com/brainstorm-tools/brainstorm3/issues/140

PR #216, opened 2019-06-07
Added permutation test for 95% confidence intervals on Tuning Curves and Spike Triggered Average

mpompolas:
Removed normalization parameter (maybe add it again)
Added 1000 permutations by default

image

This functions uses personal code for the visualization of the halo.
The halo creation should be modified once Martin finalizes the STD intervals changes, and ultimately get integrated with the Brainstorm window configuration (right now it just creates a new Matlab window for plotting).


mpompolas:
Spike Triggered Average:

image

PR #217, opened 2019-06-10
ignore empty line in coords file

Edouard2laire:
Just a small fix to avoid error when there is an empty line at the end of the fiducial/optode file.


ftadel:
Thanks

PR #218, opened 2019-06-11
Removed bi2de dependency, added magic number check for both .rhd and .rhs files

mpompolas:


ftadel:
Thanks for the fast reply!

PR #228, opened 2019-09-03
Connectivity part1

HosseinShahabi:
Richard suggested some changes in available  Coherence functions. My functions will come in the next pull request. Thanks

### DIFF
--- a/toolbox/connectivity/bst_cohn.m
+++ b/toolbox/connectivity/bst_cohn.m
@@ -9,7 +9,7 @@ function [Gxy, pValues, freq, nWin, nFFT, Messages] = bst_cohn(X, Y, Fs, MaxFreq
 %    - Fs      : Sampling frequency of X and Y (in Hz)
 %    - nFFT    : Length of the window used to estimate the coherence (must be a power of 2 for the efficiency of the FFT)
 %    - Overlap       : [0-1], percentage of time overlap between two consecutive estimation windows
-%    - CohMeasure    : {'mscohere', 'icohere'}
+%    - CohMeasure    : {'mscohere', 'icohere', 'lcohere'}
 %    - isSymmetric   : If 1, use an optimized method for symmetrical matrices
 %    - ImagingKernel : If not empty, calculate the coherence at the source level
 %    - waitMax       : Increase of the progress bar during the execution of this function
@@ -20,7 +20,8 @@ function [Gxy, pValues, freq, nWin, nFFT, Messages] = bst_cohn(X, Y, Fs, MaxFreq
 %     Gyy:  autospectral density of y
 %     Coherence function (C)            : Gxy/sqrt(Gxx*Gyy)
 %     Magnitude-squared Coherence (MSC) : |C|^2 = |Gxy|^2/(Gxx*Gyy) = Gxy*conj(Gxy)/(Gxx*Gyy) 
-%     Imaginary Coherence (IC)          : imag(C)^2 / (1-real(C)^2)
+%     Imaginary Coherence (IC)          : abs(imag(C))
+%     Lagged Coherence (IC)             : sqrt(imag(C)^2 / (1-real(C)^2))
 %
 % Parametric significance estimation:  
 %     When overlap=0%   [Syed Ashrafulla]
@@ -33,6 +34,7 @@ function [Gxy, pValues, freq, nWin, nFFT, Messages] = bst_cohn(X, Y, Fs, MaxFreq
 %        While Matlab defines the coherence as the square of the cross sepctra divided by the product of the spectra, Bokil (2007)
 %        and Schelter (2006) work with the square root within the correction and significance determination
 %
+% Warning:: pValues might not be accurate for imaginary and lagged coherence
 % References:
 %   [1] Carter GC (1987), Coherence and time delay estimation
 %       Proc IEEE, 75(2):236-255, doi:10.1109/PROC.1987.13723
@@ -64,7 +66,8 @@ function [Gxy, pValues, freq, nWin, nFFT, Messages] = bst_cohn(X, Y, Fs, MaxFreq
 % For more information type "brainstorm license" at command prompt.
 % =============================================================================@
 %
-% Authors: Sergul Aydore, Syed Ashrafulla, Francois Tadel, Guiomar Niso, 2013-2014
+% Authors: Sergul Aydore, Syed Ashrafulla, Francois Tadel, Guiomar Niso,
+% 2013-2014, Hossein Shahabi, 2019
 
 
 %% ===== INITIALIZATIONS =====
@@ -148,6 +151,7 @@ iWin = [iStart; iStop];
 % Number of samples that are taken into account in the windowing
 nSamples = iWin(end);
 % Frequency smoother (represented as time-domain multiplication)
+% smoother = bst_window('hamming', nFFT) ;
 smoother = bst_window('parzen', nFFT) .* bst_window('tukey', nFFT, 0.1);
 % smoother = bst_window('parzen', nFFT);
 smoother = smoother / sqrt(sum(smoother.^2));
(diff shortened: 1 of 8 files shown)


toolbox/connectivity/bst_cohn.m
```diff
@@ -240,8 +244,20 @@
             else
                 pValues = max(0, 1 - abs(Gxy).^2) .^ floor(nSamples / nFFT);
             end
-            % Imaginary Coherence = imag(C)^2 / (1-real(C)^2)
-            Gxy = imag(Gxy).^2 ./ (1-real(Gxy).^2);
+            % Imaginary Coherence = abs(imag(C))  % Only consider the value
+            Gxy = abs(imag(Gxy));
```
ftadel: It would be pretty bad to simply change the measure without mentioning anything...
Ideally, old scripts should keep on running exactly in the same way. We should aim for reproducibility across versions.

What you could do instead is the following:
- Keep the existing 'icohere' unchanged
- Add a new method 'imag' or 'icoherenew' or whatever you'd like
- Change the description of the input options of the process to have one per line, and include the actual  measure in the description (eg. "abs(imag(Gxy))"), mentioning that there is a new and and old implementation of the "imaginary coherence"

toolbox/process/functions/process_cohere1.m
```diff
@@ -53,13 +53,13 @@
     sProcess.options.label2.Comment = '<BR><U><B>Estimator options</B></U>:';
     sProcess.options.label2.Type    = 'label';
     % === COHERENCE METHOD
-    sProcess.options.cohmeasure.Comment = {'Magnitude-squared', 'Imaginary', 'Measure:'};
+    sProcess.options.cohmeasure.Comment = {'Magnitude-squared', 'Imaginary', 'Lagged', 'Measure:'};
```
ftadel: Include more descriptions of the measure, maybe with one measure per line, with the option type "radio_label" and the description in grey (like in many other processes).

toolbox/process/functions/process_cohere1.m
```diff
@@ -53,13 +53,13 @@
...
     sProcess.options.cohmeasure.Value   = 1;
-%     % === OVERLAP
-%     sProcess.options.overlap.Comment = {'0%', '25%', '50%', '75%', 'Overlap:'};
-%     sProcess.options.overlap.Type    = 'radio_line';
-%     sProcess.options.overlap.Value   = 3;
+    % === Overlap
+    sProcess.options.overlap.Comment = 'Sliding window overlap:';
+    sProcess.options.overlap.Type    = 'value';
+    sProcess.options.overlap.Value   = {50, '%', []};
     % === MAX FREQUENCY RESOLUTION
```
ftadel: You can't change this, it breaks the compatibility with previous scripts...
If you want to change it, you need to add another option type, set the previous one as "hidden", and then add some tests in the Run() function to see which of the new or the old option is defined...

toolbox/process/functions/process_cohere1.m
```diff
@@ -109,11 +109,12 @@
     OPTIONS.RemoveEvoked  = sProcess.options.removeevoked.Value;
     OPTIONS.MaxFreqRes    = sProcess.options.maxfreqres.Value{1};
     OPTIONS.MaxFreq       = sProcess.options.maxfreq.Value{1};
-    OPTIONS.CohOverlap    = 0.50;
```
ftadel: Same as before.

toolbox/process/functions/process_cohere1n.m
```diff
@@ -53,13 +53,13 @@
...
     sProcess.options.cohmeasure.Value   = 1;
-%     % === OVERLAP
-%     sProcess.options.overlap.Comment = {'0%', '25%', '50%', '75%', 'Overlap:'};
-%     sProcess.options.overlap.Type    = 'radio_line';
-%     sProcess.options.overlap.Value   = 3;
+    % === Overlap
+    sProcess.options.overlap.Comment = 'Sliding window overlap:';
+    sProcess.options.overlap.Type    = 'value';
+    sProcess.options.overlap.Value   = {50, '%', []};
     % === MAX FREQUENCY RESOLUTION
```
ftadel: Same as above

toolbox/process/functions/process_cohere1n.m
```diff
@@ -109,11 +109,12 @@
     OPTIONS.RemoveEvoked  = sProcess.options.removeevoked.Value;
     OPTIONS.MaxFreqRes    = sProcess.options.maxfreqres.Value{1};
     OPTIONS.MaxFreq       = sProcess.options.maxfreq.Value{1};
-    OPTIONS.CohOverlap    = 0.50;
```
ftadel: Same as above

toolbox/process/functions/process_cohere2.m
```diff
@@ -52,13 +52,13 @@
...
     sProcess.options.cohmeasure.Value   = 1;
-%     % === OVERLAP
-%     sProcess.options.overlap.Comment = {'0%', '25%', '50%', '75%', 'Overlap:'};
-%     sProcess.options.overlap.Type    = 'radio_line';
-%     sProcess.options.overlap.Value   = 3;
+    % === Overlap
+    sProcess.options.overlap.Comment = 'Sliding window overlap:';
+    sProcess.options.overlap.Type    = 'value';
+    sProcess.options.overlap.Value   = {50, '%', []};
     % === MAX FREQUENCY RESOLUTION
```
ftadel: Same as above

toolbox/process/functions/process_cohere2.m
```diff
@@ -108,11 +108,12 @@
     OPTIONS.RemoveEvoked  = sProcess.options.removeevoked.Value;
     OPTIONS.MaxFreqRes    = sProcess.options.maxfreqres.Value{1};
     OPTIONS.MaxFreq       = sProcess.options.maxfreq.Value{1};
-    OPTIONS.CohOverlap    = 0.50;
+    OPTIONS.CohOverlap    = sProcess.options.overlap.Value{1}/100;
     OPTIONS.pThresh       = 0.05;  % sProcess.options.pthresh.Value{1};
```
ftadel: Same as above

toolbox/process/functions/process_cohere2_time.m
```diff
@@ -58,13 +58,13 @@
...
     sProcess.options.cohmeasure.Value   = 1;
-%     % === OVERLAP
-%     sProcess.options.overlap.Comment = {'0%', '25%', '50%', '75%', 'Overlap:'};
-%     sProcess.options.overlap.Type    = 'radio_line';
-%     sProcess.options.overlap.Value   = 3;
+    % === Overlap
+    sProcess.options.overlap.Comment = 'Sliding window overlap:';
+    sProcess.options.overlap.Type    = 'value';
+    sProcess.options.overlap.Value   = {50, '%', []};
     % === MAX FREQUENCY RESOLUTION
```
ftadel: Same as above

toolbox/process/functions/process_cohere2_time.m
```diff
@@ -122,12 +122,13 @@
     OPTIONS.RemoveEvoked  = sProcess.options.removeevoked.Value;
     OPTIONS.MaxFreqRes    = sProcess.options.maxfreqres.Value{1};
     OPTIONS.MaxFreq       = sProcess.options.maxfreq.Value{1};
-    OPTIONS.CohOverlap    = 0.50;
+    OPTIONS.CohOverlap    = sProcess.options.overlap.Value{1}/100;
     OPTIONS.pThresh       = 0.05;  % sProcess.options.pthresh.Value{1};
```
ftadel: Same as above

toolbox/connectivity/bst_cohn.m
HosseinShahabi: We add a note on the tutorial. It is worst to continue with the wrong definition. We cannot name squared lagged coherence the imaginary coherence.

toolbox/process/functions/process_cohere1.m
HosseinShahabi: The complete description is available online on the tutorial. 
https://neuroimage.usc.edu/brainstorm/Tutorials/Connectivity#Coherence_.28FFT-based.29

toolbox/process/functions/process_cohere1.m
HosseinShahabi: You can change it to predefined first and then if it is specified in the newer version get the new one.

toolbox/connectivity/bst_cohn.m
ftadel: Existing scripts must keep on giving exactly the same results. Reproducibility is a strong objective in our work. 
Whenever possible, we should add new options instead of modifying the existing methods. In this case, it is easy to add an option and remove the old one from the GUI so that people stop using it, but keeping it in the code.

toolbox/process/functions/process_cohere1.m
ftadel: As there is some confusion in the measure involved, it wouldn't hurt to add some more details in the process window, if it fits (it has to remain displayable on a screen with a low resolution). Similarly to the statistics processes, for instance:
https://neuroimage.usc.edu/brainstorm/Tutorials/Statistics#Example_1:_Parametric_t-test_on_recordings

toolbox/process/functions/process_cohere1.m
ftadel: I'm sorry, I didn't get this. 
It looks to me like if you apply this change, the existing scripts would stop running (or would use and overlapping of 2% or 3%...)

PR #231, opened 2019-09-23
Add retry feature to download utility for BEst

aliobaibk:
Bug:
Using the function `gui_brainstorm('DownloadFile', ...` the first download attempt often fails but not the subsequent ones.
Tried with an url pointing to different hosts (DropBox, GDrive, GitHub).

Temporary fix (aka hack):
Calling `gui_brainstorm('DownloadFile', ...` twice seems to fix the issue.

### DIFF
--- a/external/brainentropy/be_install.m
+++ b/external/brainentropy/be_install.m
@@ -98,8 +98,9 @@ if ~isdir(BEstDir) || isempty(ls(BEstDir)) || FORCE
     end
     
     % Download file
+    nbAttempts = 3;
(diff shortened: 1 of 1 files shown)


ftadel:
@martcous You had been investigating these download issues from Github isn't?

martcous:
Yes, the issue is twofold. First, GitHub recently required connections to use TLSv1.2 which initially broke our Java downloader. I fixed that. The second issue is that the links GitHub uses for downloading files is indirect, as in it includes a hidden redirection that makes it seem like the link points to nothing ("File is empty" error). Adding an automatic retry if the first connection fails does fix this issue in most cases.

PR #232, opened 2019-09-23
Also project the standart deviation when projecting sources

Edouard2laire:


ftadel:
This is a good idea, but it is unfortunately incorrect mathematically.
Averaging standard errors (or interpolating on a different surface, which leads to some spatial weighted averaging) does not give the same results as computing the standard error on the projected sources.
If you need this Std value: project to the template and then average.

Edouard2laire:
Thanks for your reply Francois.

Here, i am using the process to project the result of our implementation of the GLM on the template. After we fit the GLM on our data, we then compute a contrast to have the beta map and standard deviation corresponding to the contrast vector (eg Task - Rest).  These are the map we want to project on the template to be able to run statistics at the group level. 

The following image show the  the results of student t-test on the subject anatomy (left), after projecting beta on the template (center) and after projecting beta and std on the template (left)  :  


Especially, if we look at the std maps for the subject (left), with the current implementation (center) and when projected on the template (right), we can see that projecting the std seems to give reasonable results : 



Maybe I am wrong, but to me the co-regristration using FreeSurfer registered spheres  was mainly doing a mapping between the subject cortex and the template cortex. So that to each vertex of the subject cortex correspond a vertex of the template cortex.

ftadel:
> Maybe I am wrong, but to me the co-regristration using FreeSurfer registered spheres was mainly doing a mapping between the subject cortex and the template cortex. So that to each vertex of the subject cortex correspond a vertex of the template cortex.

Except that this is not a one-to-one correspondence: each vertex in the destination surface is an average of a few nearby vertices, weighted by the distance. For all the processes that average the input values, the Std matrix is set to [] because we can't compute it correctly. The values in the Std field are always non-linear measures, and averaging them is not advised.

I don't understand what your figures show: the Std values are never displayed on the the cortex surface. It is used only to display halos around time series.
image

Edouard2laire:
> I don't understand what your figures show: the Std values are never displayed on the cortex surface. It is used only to display halos around time series.

Sorry, i didn't explained well. Here,  we are having a similar approach as in fMRI analysis by computing a General Linear Model :  here  ImageGridAmp contains the estimation of the parameters and std contains the covariance of our estimated parameters .  

what we want to do is to register these map on a template to perform group analysis.

Maybe i am wrong, but it seems that it is what is done in fMRI : Do you think they are using an other way to register this covariance to the template ? 
 
(source : https://ftp.nmr.mgh.harvard.edu/pub/docs/SavoyfMRI2014/fmri.april2011.pdf )

ftadel:
What you are doing is probably valid. But it is beyond the documented use of Brainstorm. The change you propose in this PR is not necessarily appropriate for the general use of Brainstorm.

If you are planning to distribute your GLM implementation within Brainstorm (which would be more than welcome!) then include this change in the general PR, and make sure it is used only in the case of the projection of a GLM file. Otherwise, modify the function on your local installation only.

Edouard2laire:
We had a meeting with Christophe Grova this morning about NIRSTORM. 
it seems that it would require too much work to implement the GLM in brainstorm (for EEG/MEG) as the current version integrated in Nirstorn is fNIRS oriented and mainly work for either fMRI analysis and fNIRS.  

How could I make sure that this process is only use in the case of a GLM file ? 
One option that I see would be to add additional structure to handle GLM data that would be reusable if we decide to implement the GLM latter inside brainstorm. what do you think of that ? 

An other question that I have is : how to update tutorial that are on brainstorm website  (https://neuroimage.usc.edu/brainstorm/Tutorials/NIRSFingerTapping) and what is the process to add new tutorial ? 

Sincerely
Edouard

ftadel:
The modification you suggest is not a valid operation for the data saved in this "Std" field (which are either standard errors or standard deviations).
If you need to store additional values for your own analysis, avoid re-using data fields that are already used for different purposes. You can add all the fields you want in your own files.

I'm sorry, we will not add structures that we won't be using at all.
Either you work on the same trunk as we do and merge your GLM tools as well, or you work on your branch and modify all the structures the way you want.

I will close this PR for now. Please open a new one if you want to push more code related with the GLM.

ftadel:
(I'll send you by email the instructions for editing the wiki)

Edouard2laire:
> The modification you suggest is not a valid operation for the data saved in this "Std" field (which are either standard errors or standard deviations).
> If you need to store additional values for your own analysis, avoid re-using data fields that are already used for different purposes. You can add all the fields you want in your own files.

So if we put this data in a separate filed such as `con_std` that will only used by our GLM, we can add a line in bst_project_sources.m  so that is it also projected when present ?

ftadel:
In your own branch, you can do whatever you want...

Edouard2laire:
That I understand. But maybe I am wrong and @thomas-vincent can correct me in that case, but user from NIRSTORM should be able to use the "official" version of brainstorm. 
If I only make this change in my branch, that mean that people will have to use this version of brainstorm who will not benefit from the update made on the main branch of brainstorm and that people who will use official branch of brainstorm will get wrong result when using our GLM.

ftadel:
Ok, you won :)
I've just discussed this with all the Brainstorm group and they're OK to consider that this is an "interpolation for producing cortical maps for display" and not an "averaging step with the purpose of doing further statistical analysis explicitly expecting the Std field to contain a variance".  Plus it has currently no impact on the rest of Brainstorm.

I didn't realize that what you were asking was to keep the plug and play property of NIRSTORM, I thought it was part of some extra internal developments you were not planning on sharing.

As long as it is not part of Brainstorm itself, it's fine with us if you keep this "hack" of using this Std field for storing your GLM results.  It would be easier than carrying on an additional field in the data files (that would probably need to be more invasive in the code isn't it?)

Unless you have another better solution, I would merge your PR as is, let me know.

thomas-vincent:
I agree that projecting Std values is not correct, though it provides "reasonable" outputs in most cases. It should not be offered as a stable feature. 

As you mention François:
> "interpolation for producing cortical maps for display"

It would not only be for display since it will be stored in the data, right?

> not an "averaging step with the purpose of doing further statistical analysis explicitly expecting the Std field to contain a variance"

I'm afraid the end goal is to actually compute stats, as Edouard pointed out:

> These are the map we want to project on the template to be able to run statistics at the group level.

The correct way is to project/reconstruct signals and then perform processing that compute effects and their associated Std in the source/cortical space. And this is what is done in the current GLM pipeline of nirstorm: whole time-series signals are reconstructed on the cortical surface and then the GLM is applied on them.

@Edouard2laire could you indicate a use case where you would need the Std projection in nirstorm?

Edouard2laire:
> I didn't realize that what you were asking was to keep the plug and play property of NIRSTORM, I thought it was part of some extra internal developments you were not planning on sharing.

Sorry, I didn't introduce myself. I am finishing my master 2 internship in Chistophe Grova's lab where I have been working on GLM for fNIRS along the cortical surface and I will start my PhD on fNIRS in January in the same lab. Don't hesitate to ping me (edoudela12) if you see question about NIRS on the forum. 



> @Edouard2laire could you indicate a use case where you would need the Std projection in nirstorm?

As I said previously in this topic, and explained during my master defense,  here we are doing the first level analysis on the subject specific anatomy ( so we do the pre-processing, compute the inverse problem using wMNE, and then fit the GLM and compute contrast). We then project this contrast on the template to compute Group level analysis using Mixed-Effect.

ftadel:
I am still OK to merge this PR, as long as it has no impact for other Brainstorm users. What you do on the Nirstorm side is your own responsibility.

@Edouard2laire Do you still need this?
@thomas-vincent Are you against it?

Edouard2laire:
> I am still OK to merge this PR, as long as it has no impact for other Brainstorm users. What you do on the Nirstorm side is your own responsibility.
> @Edouard2laire Do you still need this?
yes, for me its good to merge it like this.

PR #233, opened 2019-09-24
Modified the way of opening EZF and added descriptions

silencer1127:
Modified the way of opening EZF due to the change of how MATLAB manages its Apps after R2018a

Also added the descriptions back again with a try and catch block. If failed, simply return a constant string description.


ftadel:
This doesn't work: it freezes Matlab at startup on many computer configurations:
https://neuroimage.usc.edu/forums/t/brainstorm-doesnt-open-after-the-september-23th-update/14269/7

I deleted the process functions, so that it doesn't block too many people:
https://github.com/brainstorm-tools/brainstorm3/commit/42c21318aa4b775964cd849caeb9aaa522491b75

Please fix these functions and then open a new PR.
If you could get rid of the Matlab APP thing, that would help a lot... this is quite unreliable across Matlab versions, we'll be continuously annoying by backward and forward compatibilities.

martcous:
Hello,

Just a follow up on this as someone just asked me where did the process go. For now, I pointed them to your [website](https://silencer1127.github.io/software/EZ_Fingerprint/ezf_main) and told them they can install and use the app independently of Brainstorm.

While I agree with François that the optimal solution would be to not use a Matlab app at all, I feel like a super quick fix would simply be to remove the call to ezf_getProcessDescription() in your processes' GetDescription() function. The GetDescription() functions of every process are called at startup and since yours accesses and loops through all Matlab apps, this can significantly slow down or even hang the software. So if you hardcode your description in the GetDescription() function and only check for installation when the user clicks run (Run() function), it should still be functional but not hang at startup.

PR #240, opened 2019-11-29
New database search GUI & query language

martcous:
This is a big one: a brand new search GUI that allows you to narrow down the list of files in the database using boolean logic. 

### Introduction

image
image

With a combination of OR and AND blocks, exact or contain equality, case sensitive or not, and across 4 types of file metadata (name in Brainstorm, object type, path in brainstorm_db, and name of parent(s)), users can build rich search queries that can be easily accessed using a simple tab system on top of the Brainstorm database.

image

Search queries can be saved for re-use later, even after Brainstorm restarts. They can also be converted to text and copy-pasted for use in the Select Files process, in the Filter process box, or shared with collaborators.

image

### Query syntax

Text queries follow a standard format that is documented in this tutorial page: https://neuroimage.usc.edu/brainstorm/Tutorials/PipelineEditor#Filter_by_name_or_comment
The query above would look like this:

> ([name CONTAINS "foobar"] OR ([name CONTAINS "test"] AND [type EQUALS "Matrix"]))

The process box filter now works using this type of query, or a shortened version: since the filter toolbar has options to specifically search by name, file paths or parent, you can omit them to take advantage of the boolean logic still. Provided that the filter toolbar has the "Search names" option selected, the above query could simply be written in the process box filter as:

> "foobar" OR ("test" AND [type EQUALS "Matrix"])

One can also still use the filter without the query language as before, provided that their query does not include double quotes (no boolean logic will be applied in this case).

### Technical stuff

Added a new structure under GlobalData.DataBase.Searches. It includes

- iSearch: the ID of the search currently active (or 0 if whole database shown)
- Active: lists all of the searches currently active (opened in database tabs). It also saves the parent and currently selected Java node of the database for all three display modes (Anatomy, Functional sorted by subject, Functional sorted by condition) so that searches do not have to be re-computed when you're going through your search tabs quickly. See db_template('ActiveSearch').
   - I can remove this if you think this might create memory leaks or clog the memory, however it's been working pretty well in my tests so far. When it's been a while you have visited a tab, the garbage collector will have deleted the object and the search is simply re-applied on the fly. If we remove this, I don't think we should keep tabs because changing tabs is not seamless anymore.
 - All: lists all saved searches (these are saved when Brainstorm is closed)

Added a search parameter structure for building blocks of searches (see db_template('SearchParam'))
 - Value: The text the user is searching for
 - SearchType: The metadata the user wants to compare to Value (File name, type, path, parent name)
   - This is saved as a number rather than string to speed up search computation
- EqualityType: The type of equality to apply, either Contains (match part of the word) or Equals (exact search)
   - This is saved as a number rather than string to speed up search computation
- CaseSensitive: boolean, self-explanatory

Added a search node structure used to build advanced searches. This is essentially the node of a tree with the root being saved as the whole search, see db_template('SearchNode'). It can be converted on the fly to and from a string query.

- Type: 1 = SearchParam building block, 2 = boolean, 3 = parent block (no value, only children)
- Value: the SearchParam structure or boolean operator
- Children: a list of other SearchNode structures, if parent block

The node_create_*.m files were modified such that not all files are created as Java objects in the Brainstorm database. If a search is active, file metadata is sent to the new function node_apply_search() and only if the file passes the search (or, for parent objects, if one of its child passes the search) is the object created. Functions node_create_study() and node_create_subject() now return the number of objects visible such that its callers know whether the study/subject will be visible with the search active.

Finally, the filter process box now creates a SearchNode structure out of the entered text and concatenates it to the active search structure to create a single complex search that is applied only once to all the files in the process box to save computation time.

Related Java PR for icon: https://github.com/brainstorm-tools/bst-java/pull/7


martcous:
Apologies, meant to open this on my own repo for a self review.

ftadel:
Should I wait until you push the modifications to the menus before reviewing this?

ftadel:
I recompiled and pushed a brainstorm.jar with your last modifications to bst-java: https://github.com/brainstorm-tools/bst-java/commit/e5a14879bf2e26030de35bed03c3df49b0f9f0ae
I didn't realize it was part of this PR as well, sorry if it causes some confusion...

martcous:
No problem, I'll fix the conflict and update the GUI tomorrow as I'm swamped today. Since the modifications requested are only related to the menus, I think it's fine if you start the review before I push these if you have time (no rush at all). Thanks!

martcous:
I moved the menus like discussed. I also decided not to change the Select By Tag process and create a new one instead. I added a "Generate pipeline" option that opens up the new search process with the active search string already typed. However, I am not sure if this is useful: the pipeline editor already takes into account the active search, so adding a search process when a search is active does not do anything. I agree that it would be useful in scripts though, but keeping it as is might be misleading to novice users that do not use the pipeline editor to generate scripts. Maybe an option "Generate script" directly rather than going through the pipeline editor? Let me know what you think.

martcous:
Apologies for the technical issues today. I updated the screenshots to reflect the GUI changes thanks to your suggestions. The above point still remains to be discussed, RE: pertinence of the Generate pipeline option.

ftadel:
Beautiful!

Indeed, this "generate pipeline" thing was not very useful, because it doesn't to the initial file selection itself... I replaced it with something that creates the process string and copies it to the clipboard.

martcous:
Thanks a lot for the review, François! This is a pretty major change, I expect multiple bug reports to be coming our way in the coming days, I'll do my best to fix them quickly. In the meantime, I'll slowly go through the tutorial pages and update the screenshots.

ftadel:
I've been playing with it for a while and it seems to be working beautifully so far!

For the documentation, you could add a new section after the one introducing the filter box (as you already started, https://neuroimage.usc.edu/brainstorm/Tutorials/PipelineEditor#Filter_by_name_or_comment)
With lots of screen captures :)

The screen captures and text that need to be fixed because of the renaming from "file comment" to "file name":
- https://neuroimage.usc.edu/brainstorm/Tutorials/Scripting#Running_the_script_again
- https://neuroimage.usc.edu/brainstorm/Tutorials/PipelineEditor#Process:_Select_files_with_tag
- https://neuroimage.usc.edu/brainstorm/Tutorials/PipelineEditor#Error_management
- https://neuroimage.usc.edu/brainstorm/Tutorials/PipelineEditor#Control_the_output_file_names
- https://neuroimage.usc.edu/brainstorm/Tutorials/CtfCoilOrientBugFix
- https://neuroimage.usc.edu/brainstorm/ExportSpm8#Using_unconstrained_sources

PR #243, opened 2019-12-03
Add possibility to generate 4 layers FEM mesh from surfaces 

tmedani:
Update the way to define the regions (code adapted for multilayer mesh).
Add an option to ask the user to generate either 3 layers model (as it is in the previous version)
or generate 4 layers. 
In the case of the 4 layers, a new surface will be generated by inflating the
outer skull (function bst_inflateORdeflate_surface at line 194) by skullThiskness value (at line 336).
Update the FemMat structure in the case of 4 layers.


ftadel:
Given the size of the compiled binaries, we won't integrate them in the main Brainstorm distribution directly. We're trying to keep the basic source package as small as possible, so that updating Brainstorm regularly is not too much of a pain even with a slow internet connection.

Instead, we will go through an automatic download, like what is available for iso2mesh and roast (see process_generate_fem).

Which means that we need to move most of the code in the PR to a different repository, and that we will need to create a .zip package for Brainstorm to download (made available for download on the neuroimage webserver).

I created a new repo, and a team bst-duneuro that has write access to it:
https://github.com/brainstorm-tools/bst-duneuro

Could you try splitting this PR in two?
1) the part that belongs to brainstorm (probably only the process functions) => open PR on this repo
2) the "bst_duneuro" toolbox => push it directly on the master branch of the other repo (once you accept the invitation to the bst-duneuro team, you should be able to push to the repo instead of opening pull requests all the time)

I'll review all this next week (and when it's split in two).

Thanks!

tmedani:
Hey, 

Ok, I will do it asap. 

I will finish the MEG integration and the combined MEG/EEG, 
then I will push all to the new rep.

Thanks

juangpc:
Hi, 

Is it worth to do a step by step compilation guide? Or users will not bother, or the ones who would want to compile the app will know how to do it right?

ftadel:
Yes, we do need a compilation guide.
Not necessarily for the Brainstorm users, but at least for us, internally. We need to have the possibility to maintain this tool easily in the future. 
Please make notes of all the steps you're going through to obtain the binaries, so that someone else (eg. Martin or myself) could reproduce it easily.

The best place to document the compilation would probably be the README of the new repo where the code and binaries will be pushed (Markdown syntax).

Thanks!

tmedani:
Hey,

Yes, of course, it makes sense and it's important.  

@juangpc I think you have already made good documentations for the duneuro compilation part. 
On the  [SVD bst Git repo](https://github.com/svdecomposer/brainstorm-duneuro)
We need to update it and add it to the new repo.

For the EEG/MEG computation, the Matlab implementation that we are going to add will be documented as well. 
I have already started this [page ](https://neuroimage.usc.edu/brainstorm/Duneuro?highlight=%28duneuro%29) for documentation, but it needs to be completed with a full tutorial when the integration is completed. 


@ftadel what is 'Markdown syntax'?  
is it a specific format?

Thanks & A+

ftadel:
This all looks great! Thanks!

> what is 'Markdown syntax'?

Markdown is a widely used syntax for formatting text, for wikis or online documentation. The text of the README file is displayed in github using a Markdown parser.
https://guides.github.com/features/mastering-markdown/

You guys are already using it in the README.md file in https://github.com/svdecomposer/brainstorm-duneuro (.md stands for MarkDown)

PR #245, opened 2019-12-09
BST vs "clear all"

juangpc:
~~ NOT URGENT ~~

Hi all,
I've found a problem which might be of interest to others. The problem is related to the behavior of bst when the GlobalData variable has been deleted. 

Context:
 - Start bst normally. While running bst.
 - execute ```clear all```.
 - try to close bst. 

There is a problem with GlobalData, precisely with the way Matlab treats the dot operator of undeclared variables. My doubts are not precisely focused on the fact that ```clear all``` corrupts brainstorm's instance. My doubts are related to the fact that if a user is not comfortable with bst's underpinnings, he/she might need to restart Matlab altogether, in order to close bst's main window. I think this is undesirable. ```clear all``` is too common in many scripts, in my opinion, to force the user to restart Matlab. The fact that GlobalData is 'hidden' does not help, because the user will not know ```clear all``` is actually erasing anything since the Workspace window will appear empty. But that is my opinion.

One quick fix is to call ``` jBstFrame.dispose(); ``` directly breaking execution inside closeWindow_Callback, but I think a regular user would not know how to do that. 

If other people feel that this is not a problem that needs to be fixed: sorry for the interruption. I'm just trying to help. It would be great if somebody could shed a bit of light on why we should not consider this a problem and/or why is this a preferred behavior for the toolbox. If anybody has a better solution or can find a problem with the code I propose, please let me know.

As in the previous PR, this is not urgent, at all. 

Proposed solutions:
 - One first issue is that the 'X' button callback (aka closeWindow_Callback) tries to save the position of the window for future runs. Which I think is legit but comes costly. A try/catch solves that.
 - Once the previous problem is solved, the final issue is that bst_exit is also interrupted mid-through (due to the same problem with GlobalData and the dot operator). The second time the button is pressed, however, it will exit with a 0 code, and thus ```jBstFrame.dispose();``` will be executed OK. A second try/catch structure in bst_exit function solves the problem.

Thanks all,

### DIFF
--- a/toolbox/core/bst_exit.m
+++ b/toolbox/core/bst_exit.m
@@ -77,20 +77,25 @@ try
         end
     end
 catch
-  status = 0;
+  status = -1;
   return;
 end
 
 %% ===== CLOSE WINDOW =====
 % Only in the GUI was created
-if (GlobalData.Program.GuiLevel >= 0)
-    % Hide all the registered panels
-    listPanels = GlobalData.Program.GUI.panels;
-    for iPanel = 1:length(listPanels)
-        gui_hide(listPanels(iPanel)); 
-    end
-    % Close Brainstorm main window
-    ctrl.jBstFrame.dispose();
+try
+  if (GlobalData.Program.GuiLevel >= 0)
+      % Hide all the registered panels
+      listPanels = GlobalData.Program.GUI.panels;
+      for iPanel = 1:length(listPanels)
+          gui_hide(listPanels(iPanel)); 
+      end
+      % Close Brainstorm main window
+      ctrl.jBstFrame.dispose();
+  end
+catch
+  status = -1;
+  return;
 end
 % Release Brainstorm global mutex
 bst_mutex('release', 'Brainstorm');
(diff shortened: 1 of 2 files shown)


toolbox/core/bst_exit.m
```diff
@@ -50,33 +50,37 @@
 % Stop execution
 rmappdata(0, 'BrainstormRunning');
 % Only in the GUI was created
-if (GlobalData.Program.GuiLevel >= 0)
```
juangpc: The code inside the ```if(GlobalData.Program.GuiLevel >=0)``` is preserved inside the try/catch structure. Don't know why git's diff parser did not catch that.

ftadel:
These are good remarks but I implemented it in a different way:
https://github.com/brainstorm-tools/brainstorm3/commit/aaefea8a20f521d9ebbd23ea3fe07b8fd3e661eb

I tend to prefer testing the existence of fields and variables rather than using generic try/catch blocks, we test exactly what we expect for, and the behavior is more reproducible across Matlab versions.

On the functional part, your suggested code had one issue: it bst_exit returns -1, we do NOT want to force quit. This happens when you try to close Brainstorm with unsaved modifications (eg. modified events): brainstorm asks you "do you want to save the modifications", if you answer "cancel" you get -1 in return and expect the Brainstorm window to remain open to finish your work.

Other than that, does this commit do what you were expecting to address?

juangpc:
Tested and working just fine. No need to restart Matlab altogether. Great!
Thanks!

Juan

PR #246, opened 2020-01-08
add a menu popup to generate FEM mesh from surfaces files

tmedani:
add a menu item to call the function GenerateFemMesh that generates the FEM mesh from the selected surface files.


toolbox/tree/tree_callbacks.m
```diff
@@ -3053,6 +3055,12 @@ function SurfaceAverage(TessFiles)
     end
 end
 
+%% ===== GENERATE FEM MESH =====
```
ftadel: No need to create an extra useless function here. 
The callback should point directly at fem_mesh.m. Or probably at process_generate_fem.m after we reorganize the other PR.

tmedani: yes, I totally agree.

ftadel:
Done: 
https://github.com/brainstorm-tools/brainstorm3/commit/b4e92163a820fa0f75108eade974909130b1087c#diff-2ef654c5427b5793045f26aba06dfc8aR1083

PR #247, opened 2020-01-08
Create fem_mesh.m

tmedani:
a function called by "tree_callbacks" when the user clicks on the menu: Generate FEM mesh
It will generate the tetrahedral mesh from the selected surfaces files (or from the merged surfaces file) 
This function will also check if Iso2mesh is installed (the same function used within process_generate_fem.m).

### DIFF
--- a/toolbox/anatomy/fem_mesh.m
+++ b/toolbox/anatomy/fem_mesh.m
@@ -1,0 +1,312 @@
+function  fem_mesh(TessFiles)
+% FEM_MESH: generate tetrahedral mesh from surface files.
+%                       This function calls the iso2mesh toolbox (function surf2mesh and raytrace)
+% USAGE:  fem_mesh(TessFiles)
+%
+%
+% INPUT:
+%    - TessFiles   : Cell-array of paths to surfaces files to concatenate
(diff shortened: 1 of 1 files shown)


tmedani:
Hi, 
These modifications allow the generation of FEM mesh from surfaces mesh. 
The idea is to allow the user to generate its own tetra head model from selected surfaces.
The surfaces will be either from the template (ICBM) or from any bst surface file format.
User can select the desired surfaces and then click on the menu 'Generate FEM mesh' 

(The screenshots in this thread are just for illustration, not for real cases.)

image

Then, he can select the desired resolution : 
image

The get the FEM mesh (within bst databas)

image


This modification works also for merging surfaces.
image

And then call the process of FEM generation : 

image

a quick example : 
image

ftadel:
Good.
But there is already process_generate_fem.m that handles the creation of tetrahedral meshes, let's not create multiple entry points to the same thing...
What do you want to do with this existing function?

tmedani:
Hi, 
yes as discussed, it's could be better to integrate this function within process_generate_fem. 
Thanks

PR #255, opened 2020-01-29
Update process_generate_fem.m

tmedani:
Use the function "mergesurf" unstead of "mergemesh".
mergemesh merge simply the surfaces and if there are intersection between the surfaces, the volume mesh generation will fail.
mergesurf can handles surface mesh intersection by splitting the intersecting elements and creates new nodes.
with this modification, the volume mesh generation will work.

### DIFF
--- a/toolbox/process/functions/process_generate_fem.m
+++ b/toolbox/process/functions/process_generate_fem.m
@@ -252,7 +252,7 @@ function [isOk, errMsg] = Compute(iSubject, iMris, isInteractive, OPTIONS)
                 bemMerge = cat(2, bemMerge, BemMat.Vertices, BemMat.Faces);
             end
             % Merge all the surfaces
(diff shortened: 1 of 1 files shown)


tmedani:
From software view, this is a safe way to generate a correct volume mesh than using mergemesh.
It seems that in the case of the surface intersection, surf2mesh will fail and then the previous correct results of surf2mesh will be loaded ... this is a strange behavior of iso2mesh. 

From the user's view, they are not aware that the surfaces are intersecting, therefor their model could be wrong.  We need to check and add a warning

ftadel:
> We need to check and add a warning...

How do you check this?
Do I wait until you push this additional check before merging this PR?

tmedani:
> Do I wait until you push this additional check before merging this PR?

we could merge this new modification since it's better than the previous. 
 


> How do you check this?

I tested some functions, but for now, I don't have any solution.

PR #257, opened 2020-01-29
Connectivity part2 draft

HosseinShahabi:
The second part of the connectivity section

### DIFF
--- a/toolbox/connectivity/bst_connectivity.m
+++ b/toolbox/connectivity/bst_connectivity.m
@@ -8,7 +8,7 @@ function OutputFiles = bst_connectivity(FilesA, FilesB, OPTIONS)
 % This function is part of the Brainstorm software:
 % https://neuroimage.usc.edu/brainstorm
 % 
-% Copyright (c)2000-2019 University of Southern California & McGill University
+% Copyright (c)2000-2020 University of Southern California & McGill University
 % This software is distributed under the terms of the GNU General Public License
 % as published by the Free Software Foundation. Further details on the GPLv3
 % license can be found at http://www.gnu.org/copyleft/gpl.html.
@@ -22,7 +22,7 @@ function OutputFiles = bst_connectivity(FilesA, FilesB, OPTIONS)
 % For more information type "brainstorm license" at command prompt.
 % =============================================================================@
 %
-% Authors: Francois Tadel, 2012-2015; Martin Cousineau, 2017; Hossein Shahabi, 2019
+% Authors: Francois Tadel, 2012-2015; Martin Cousineau, 2017; Hossein Shahabi, 2020
 
 
 %% ===== DEFAULT OPTIONS =====
@@ -591,22 +591,17 @@ for iFile = 1:length(FilesA)
                 size(sInputA.Data,1), size(sInputB.Data,1)));
             Comment = [OPTIONS.tfMeasure ' | ' OPTIONS.CohMeasure ' | ' sprintf('%ds',OPTIONS.WinParam(1)) ' | ' ...
                 sprintf('%ds',prod(OPTIONS.WinParam)) ' | '] ;
-            
-            % ======
+            % Additional options for the progress bar inside the function
             OPTIONS.SampleRate        = sfreq;
             OPTIONS.ExtraInf.nTrials  = length(FilesA);
             OPTIONS.ExtraInf.trialNum = iFile;
             OPTIONS.Freqs             = OPTIONS.Freqrange ;
-
             % ======
-            [cOut,timeSamples] = bst_hcoh(sInputA.Data,OPTIONS);
-            R4d = cOut ;
-            sInputB.Time = timeSamples + sInputB.Time(1) ;
-            [~,~,nTime,nFreq] = size(R4d) ;
-            R = reshape(R4d,[],nTime,nFreq) ;
-            
-            % =======
-            OPTIONS.Nwin = nTime ;
+            [R4d,timeSamples] = bst_hcoh(sInputA.Data,OPTIONS);
+            sInputB.Time      = timeSamples + sInputB.Time(1) ;
+            [~,~,nTime,nBand] = size(R4d) ;
+            % Rehaping a 4D matrix to 3 dim
+            R = reshape(R4d,[],nTime,nBand) ;
             
         otherwise
             bst_report('Error', OPTIONS.ProcessName, [], ['Invalid method "' OPTIONS.Method '".']);
--- a/toolbox/connectivity/bst_hcoh.m
+++ b/toolbox/connectivity/bst_hcoh.m
@@ -1,0 +1,231 @@
+function varargout = bst_hcoh(Data,OPTIONS)
+%% Inputs and Outputs Definition:
+%
+% INPUTS:
+%   - Data                   : Input signals (nSignals x nSamples)
+%   - OPTIONS: 
+%     - Fs                   : Sampling frequency
+%     - Method               : Desired measure of correlation
+%                              'coh' - coherence
+%                              'lcoh'- lagged coherence
+%                              'env' - envelope correlation
+%     - Windowing parameters : winSize:  window size in second for dynamic network reconstruction
+%           noverlap: overlap between windows in second, should be smaller
+%           than winSize.
+%
+% OUTPUTS:
+%
+% ------------------------------------------
+% Example:
+% @=============================================================================
+% This function is part of the Brainstorm software:
+% https://neuroimage.usc.edu/brainstorm
+% 
+% Copyright (c)2000-2020 University of Southern California & McGill University
+% This software is distributed under the terms of the GNU General Public License
+% as published by the Free Software Foundation. Further details on the GPLv3
+% license can be found at http://www.gnu.org/copyleft/gpl.html.
+% 
+% FOR RESEARCH PURPOSES ONLY. THE SOFTWARE IS PROVIDED "AS IS," AND THE
+% UNIVERSITY OF SOUTHERN CALIFORNIA AND ITS COLLABORATORS DO NOT MAKE ANY
+% WARRANTY, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO WARRANTIES OF
+% MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE, NOR DO THEY ASSUME ANY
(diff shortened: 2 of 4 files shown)


toolbox/connectivity/bst_connectivity.m
```diff
@@ -591,22 +591,17 @@
...
-            R4d = cOut ;
-            sInputB.Time = timeSamples + sInputB.Time(1) ;
-            [~,~,nTime,nFreq] = size(R4d) ;
-            R = reshape(R4d,[],nTime,nFreq) ;
-            
-            % =======
-            OPTIONS.Nwin = nTime ;
+            [R4d,timeSamples] = bst_hcoh(sInputA.Data,OPTIONS);
+            sInputB.Time      = timeSamples + sInputB.Time(1) ;
+            [~,~,nTime,nBand] = size(R4d) ;
```
ftadel: Please replace the ~ with "tmp" or some unused variable name. This syntax is not supported by version of Matlab, it makes the parser crash.

toolbox/connectivity/bst_hcoh.m
```diff
@@ -0,0 +1,231 @@
...
+%                              'lcoh'- lagged coherence
+%                              'env' - envelope correlation
+%     - Windowing parameters : winSize:  window size in second for dynamic network reconstruction
+%           noverlap: overlap between windows in second, should be smaller
+%           than winSize.
+%
+% OUTPUTS:
+%
+% ------------------------------------------
+% Example:
```
ftadel: Remove the lines above if you are not describing the output or including examples

toolbox/process/functions/process_hcoh.m
```diff
@@ -0,0 +1,192 @@
...
+%
+% Author: Hossein Shahabi, 2020
+eval(macro_method);
+end
+
+
+%% ===== GET DESCRIPTION =====
+function sProcess = GetDescription() %#ok<DEFNU>
+    % === Description the process
+    sProcess.Comment     = 'Coherence/Envelope by Hilbert/Morlet NxN (2020)';
```
ftadel: Could you come up with a more compact label?

toolbox/process/functions/process_hcoh.m
```diff
@@ -0,0 +1,192 @@
...
+    sProcess.options.win.Type    = 'value';
+    sProcess.options.win.Value   = {1500, 'ms', []};
+    % === overlap
+    sProcess.options.overlap.Comment = 'Sliding window overlap:';
+    sProcess.options.overlap.Type    = 'value';
+    sProcess.options.overlap.Value   = {50, '%', []};
+    % === Parallel processing for envelope correlation
+    sProcess.options.parMode.Comment = 'Parallel processing (envelope correlation)';
+    sProcess.options.parMode.Type    = 'checkbox';
+    sProcess.options.parMode.Value   = 0 ;
```
ftadel: You should be also testing whether the parallel processing toolbox is available.
Example from process_pac.m:
```
    if ~exist('matlabpool', 'file') && ~exist('parpool', 'file')
        sProcess.options.parallel.Hidden = 1;
    end
```

toolbox/connectivity/bst_hcoh.m
```diff
@@ -0,0 +1,231 @@
...
+Data = bst_bsxfun(@minus, Data, mean(Data,2)) ;
+
+% Block analysis for large files 
+if numBlocks>1
+    tfBlockRem = rem(N,numBlocks) ;
+    % Adding zero to the end of the data (few samples)
+    if tfBlockRem ~= 0
+        Data(:,end:end+(numBlocks-tfBlockRem)) = 0 ;
+    end
+    [~,N]  = size(Data) ;
```
ftadel: Avoid the ~ syntax

toolbox/connectivity/bst_hcoh.m
```diff
@@ -0,0 +1,231 @@
...
+    error('Invalid call.');
+end
+
+% Options for connectivity analysis
+[nSig,N]     = size(Data) ;
+Fs           = OPTIONS.SampleRate ;
+winSize      = OPTIONS.WinParam(1) * Fs ;
+overLap      = OPTIONS.WinParam(2) * winSize ;
+numBlocks    = OPTIONS.tfSplit ;
+parMode      = OPTIONS.parMode ;
```
ftadel: You should be also testing whether the parallel processing toolbox is available.
Example from process_pac.m:
```
    OPTIONS.isParallel   = sProcess.options.parallel.Value && ((exist('matlabpool', 'file') ~= 0) || (exist('parpool', 'file') ~= 0));
```

toolbox/connectivity/bst_hcoh.m
```diff
@@ -0,0 +1,231 @@
...
+        elseif k == numBlocks
+            range2D = (N-blockLen3D+(tranLen/2)+1):N ; 
+            range3D = 1:(blockLen3D-(tranLen/2)) ;
+        else
+            range2D = ((k-1)*actLen -(tranLen/2) ) + (1:blockLen3D) ;
+            range3D = 1:blockLen3D ; 
+        end
+        Data3D(:,range3D,k) = Data(:,range2D) ;
+    end
+    [~,tfBlockLen,~] = size(Data3D) ;
```
ftadel: Remove ~

toolbox/connectivity/bst_hcoh.m
```diff
@@ -0,0 +1,231 @@
...
+else
+    [~,tfBlockLen] = size(Data) ;
```
ftadel: Remove ~

toolbox/connectivity/bst_hcoh.m
```diff
@@ -0,0 +1,231 @@
...
+%% Parallel Processing 
+function HParaStart(parMode)
+g1 = gcp('nocreate') ;
+if isempty(g1) || (~isempty(g1) && g1.NumWorkers ~= parMode)
+    if ~isempty(g1)
+        delete(gcp) ;
+    end
+    myCluster = parcluster('local') ;
+    myCluster.NumWorkers = parMode ;
+    parpool(parMode)
```
ftadel: Not all the versions of Matlab work in the same way... 

Either you adopt a syntax that supports Matlab <= 2013 (example from bst_process.m):
```
            try
                if (bst_get('MatlabVersion') >= 802)
                    hPool = parpool;
                else
                    matlabpool open;
                end
            catch
            end
```

Or you consider that if the version of Matlab is too old, you just don't use the parallel process toolbox (example from process_spike_field_coherence.m):
```
        % Prepare parallel pool, if requested
        if sProcess.options.paral.Value
            try
                poolobj = gcp('nocreate');
                if isempty(poolobj)
                    parpool;
                end
            catch
                sProcess.options.paral.Value = 0;
                poolobj = [];
            end
        else
            poolobj = [];
        end
```

The best solution: let Brainstorm start the parallel processing pool for you:
If your process option is called "parallel" and its value is equal to 1, the bst_process.m will create a pool before calling the Run() function in your process:
https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/process/bst_process.m#L83
(maybe this code requires a little upgrade...)

Example of use of this procedure: process_pac.m
https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/process/functions/process_pac.m#L100

toolbox/process/functions/process_hcoh.m
```diff
@@ -0,0 +1,192 @@
...
+    % === Time window
+    sProcess.options.win.Comment = 'Estimation window length:';
+    sProcess.options.win.Type    = 'value';
+    sProcess.options.win.Value   = {1500, 'ms', []};
+    % === overlap
+    sProcess.options.overlap.Comment = 'Sliding window overlap:';
+    sProcess.options.overlap.Type    = 'value';
+    sProcess.options.overlap.Value   = {50, '%', []};
+    % === Parallel processing for envelope correlation
+    sProcess.options.parMode.Comment = 'Parallel processing (envelope correlation)';
```
ftadel: Name this option "parallel" (see previous comment)

ftadel:
> I will update the parameter discussion in the function description (any comments?)

Yes, some description of the measures would be good in the header of bst_hcoh.m, similarly to what is in bst_cohn.m.

> The progress bar needs your attention, I have provided all necessary variables for that (number of files, frequency bands, time windows

The example in bst_cohn.m are not sufficient?

> Do we need more parameter predefinition for this function?  

I'm not sure what this means.

> Although the output file (connectivity) matrix has both frequency and time dimensions, I cannot show the time domain. Maybe I need to change some other files as well. Same for Spectrum visualization you said one time but I forgot. 

You can search for the string 'plvt' in the code (CTRL+SHIFT+F, search .m files including subfolders), and see where there are special cases in the display functions, and maybe you just need to add 'hcoh' next to 'plvt' to get the same display exceptions.
Or compute the time-resolved coherence and see what the differences are with your files.

ftadel:
@HosseinShahabi @richardmleahy Any update?

HosseinShahabi:
@ftadel I'm updating these functions based on your suggestions and their latest version. 
Regarding parallel processing, I tried to use Brainstorm but there is an issue. 
The Brainstorm starts parallel processing before going to run section. However, Parallel processing is only useful for "envelope correlation" and I don't want to use it for coherence parts. 
The problem is that you cannot have an "if" clause which depends on other fields of sprocess options inside the GetDiscription section. Can you help me with that? 
Thanks

ftadel:
Aren't coherence and envelope correlation executed from different processes?
If so the latter would have sProcess.options.parallel set, but not the former.

If this all has to be executed from the same panel, then for now you can start with an explicit checkbox (option .parallel) and add to its label that the parallel option is only used in one of the options.
If all cases,  you have to offer the parallel processing as an option, as it does not necessarily speed up things in all the configurations.

Maybe I'm missing your point... Could you post a screen capture of what your processes option panels look like?

HosseinShahabi:
Thanks. 

All functions use the same panel and code for running. I guess for now I would mention that in the label. 

I fixed the progress_bar as well but there is one point. It only shows the progress for the current "file" and doesn't consider the total number of files (like trials). I noticed there is a separate progress_bar in "bst_connectivity" but since it probably doesn't stay on the screen for a long time, it is hard to find out. I can add incorporate the number of files in my function but it might contradict the general framework.  

Also, as you know because of the nature of parallel processing it is complicated to estimate the actual progress. I think we don't have to address that issue for now. 

The last thing is the naming of the function which I will bring it up in our meeting.  

image

ftadel:
Should this PR be closed now?

PR #259, opened 2020-02-07
extract the cortex surface and save to database 

tmedani:
This part of the code will extract only the cortex surface from the FEM mesh. 
This surface will be used later as the source space (shrunk by X mm).


tmedani:
> @tmedani
> This works but doesn't save the correct surface.
> If I run it on the ICBM152 template with 3 layers, I get this as the "cortex_fem" file saved in the database:
> image

indeed! 
in the case of three layers, the label should be "<=1" and not "=<2 "
It should be before the transformation.
Thanks for testing and solving it :) 

As Iso2mesh, we need also to keep tracking the SimNibs changes in the future,
if they include other tissues the labels will changes.

PR #261, opened 2020-02-13
Update process_generate_fem.m

tmedani:
https://github.com/brainstorm-tools/brainstorm3/issues/185#issuecomment-585502671


ftadel:
@tmedani I fixed the surface ordering in a different way from what you suggested: 
https://github.com/brainstorm-tools/brainstorm3/commit/409b64125dbb0d37d13bc4695a06664e9cfe906a

I removed the automatic reordering (which has high chances of not working), and rely only on the surface comments or types (which in most cases will be set correctly). 
I still am not sure to understand how this initial regions definition works, and how it relates to the labels and order of the final elements, but it seems to work the way to coded it.

Does this work for you?
Feel free to reopen the pull request if what I did doesn't work.

PR #262, opened 2020-02-19
choose between "mergsurf" or "mergemesh"

tmedani:
adding a checkbox option that asks the user if they want to
 call "mergsurf"  or  call "mergemesh"

In most case the mergemesh is sufficient. 
mergesurf takes longer since there is the self-intersection checking.
The execution time between mergsurf and mergemesh:
        tic; [newnode, newelem] = mergesurf(bemMerge{:}); tmergsurf = toc;
        tic; [newnode, newelem] = mergemesh(bemMerge{:}); tmergemesh = toc;
tmergsurf = 30.3988 s (here with three surfaces ... should be longer with more surfaces )
tmergemesh = 0.1330 s

### DIFF
--- a/toolbox/process/functions/process_generate_fem.m
+++ b/toolbox/process/functions/process_generate_fem.m
@@ -309,8 +309,20 @@ function [isOk, errMsg] = Compute(iSubject, iMris, isInteractive, OPTIONS)
                 bemMerge = cat(2, bemMerge, BemMat.Vertices, BemMat.Faces);
             end
             % Merge all the surfaces
-            % [newnode, newelem] = mergesurf(bemMerge{:});
-            [newnode, newelem] = mergemesh(bemMerge{:});
+            % NOTE: mergesurf more robust for intersecting meshes than mergemesh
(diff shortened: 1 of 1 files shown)


ftadel:
You should not add any interactive selection of options in the processing part of the process (functions Run and Compute): this must be able to run on a computation cluster with no display attached.
I adapted your suggestion here, with a new process option: https://github.com/brainstorm-tools/brainstorm3/commit/d39e95186f93f87ad957de79efb38f9b4ee16e4a

ftadel:
PS: make sure you pull the last modifications before you edit a function you want to commit on the brainstorm master branch.

tmedani:
> You should not add any interactive selection of options in the processing part of the process (functions Run and Compute): this must be able to run on a computation cluster with no display attached.
> I adapted your suggestion here, with a new process option: [d39e951](https://github.com/brainstorm-tools/brainstorm3/commit/d39e95186f93f87ad957de79efb38f9b4ee16e4a)

Thank you, I don't know this. 

> PS: make sure you pull the last modifications before you edit a function you want to commit on the brainstorm master branch.

Yes... sorry! 
I have multiple versions on my computer.

PR #264, opened 2020-02-20
Update process_generate_fem.m

tmedani:
add "try catch" to prevent the crash of the mergesurf function on windows OS


tmedani:
Add the possibility to display hexahedral FEM mesh by converting the hexa to tetra. 
This is based on the subfunction hex2tet and its dependencies (element2patch).

PR #266, opened 2020-02-24
convert mesh on database

tmedani:
add on the FEM menu the functions that convert the mesh (tetra to hexa / hexa to tetra).
The associated functions "fem_tetra2hexa.m" and "fem_hexa2tetra.m" 
are saved on bst-duneuro/matlab

Thes options are useful for testing and integrating the Hexa mesh (displays and comparisons of the FEM computation (tetra vs Hexa) on the same model).


ftadel:
I handled your request in this way:
https://github.com/brainstorm-tools/brainstorm3/commit/c923d19de374e0f0cd9af51dbc3f1478c059fdb0

Because you put fem_hexa2tetra and fem_tetra2hexa in bst-duneuro, we need to test for the installation of bst-duneuro first.

tmedani:
Thanks :)

PR #272, opened 2020-03-13
Update tree_callbacks.m

tmedani:
add the option to display the leadfield on right-click menu of the head model


ftadel:
Done in: https://github.com/brainstorm-tools/brainstorm3/commit/4fe1a9087bb0f005a16fe96dc43accd827947554

PR #273, opened 2020-03-22
Update process_generate_fem.m

tmedani:
add option to generate the hexa mesh from simnibs
- The hexa mesh is generated from the final segmentation msks of simnibs
- The hexa mesh is transformed to brainstorm coordinates 
- users have the option to select tetra/hexa or both for the mesh generation
- The cortex surfaces could be extracted from both methods 

image

image

image
image

### DIFF
--- a/toolbox/process/functions/process_generate_fem.m
+++ b/toolbox/process/functions/process_generate_fem.m
@@ -615,8 +615,22 @@ function [isOk, errMsg] = Compute(iSubject, iMris, isInteractive, OPTIONS)
             end
                   
             % === IMPORT OUTPUT FOLDER ===
-            bst_progress('text', 'Importing SimNIBS output...');
-            % Import FEM mesh
+        bst_progress('text', 'Importing SimNIBS output...');
+        
+         % === IMPORT CAT12 OUTPUT ===
+        CatDir = bst_fullfile(simnibsDir, ['m2m_' subjid], 'segment', 'cat');
+        if isdir(CatDir)
+            catErrMsg = import_anatomy_cat(iSubject, CatDir, OPTIONS.NbVertices, 0, [], 0, 2);
+            if ~isempty(catErrMsg)
+                bst_report('Warning', 'process_generate_fem', [], ['Could not import CAT12 segmentation: ' 10 catErrMsg]);
+            end
+        else
+            bst_report('Warning', 'process_generate_fem', [], ['CAT12 segmentation not found in SimNIBS output folder: ' 10 CatDir]);
+        end       
+       
+        %%=== IMPORT FEM MESH OUTPUT ===
+        % Import the tetrahedral mesh
+        if ~isempty(strfind(OPTIONS.MeshType, 'tetrahedral'))
             mshfilename = bst_fullfile(simnibsDir, [subjid '.msh']);
             if ~file_exist(mshfilename)
                 errMsg = ['Could not find SimNIBS output mesh: ' mshfilename];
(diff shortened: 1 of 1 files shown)


ftadel:
Before I review the code and fix a few things, I'd like to understand better the goal of your work: 
Why do we need these hexahedral meshes? 
Does this give any advantage in terms of computation of the FEM forward model? 

It just seems less practical and heavier, these meshes are so dense that we wouldn't be able to practically compute a FEM solution for them... Maybe not all your experiments need to end up in the Brainstorm distribution, on this branch I'd like to focus only on tools that we will actually distribute.
There is enough work to do not to develop directions we won't use. 

Do you think this will be part of final tools we will recommend using?

tmedani:
Hi Francois, 
indeed, all these questions are relevant. 

From my point of view, the tetra is more realistic, however, a lot of FEM users still use the hexa as well. 
It has some advantages related to fast mesh generation (as is the case on fieldtrip option, which converts the voxel to cube).
and in some cases, it allows discontinuities on surfaces (cortex can touch the skull), which is not the case (or not recommended) for tetra.

> Why do we need these hexahedral meshes?

indeed, we can just use the tetra. 
however, I'm following the brainstorm project description and it mentioned the use of the hexa as well (AIM 2 : @jcmosher & @richardmleahy ) and since duneuro offers this option, I thought it's also good to have access to it. 

> Does this give any advantage in terms of computation of the FEM forward model?

in some case, it can be faster, 
quick thought, in a similar dimension, 1 hexa cube is divided into (at least)  5 tetra and could add more nodes, which will increase the size and the computation time.

For the accuracy, the adapted hexa geometry seems to have better results than the regular Hexa,  comparing adapted hexa with tetra, I don't have the correct answer,  it's in debates...

I will share your question with Carsten and Christian to have their answers. 

Do you think this will be part of the final tools we will recommend using?

since we have the duneuro computation available and also the mesh, I think it nice to have it,
it costs us more work (almost done).
with this, the users will have a choice, and brainstorm will be the unique tools to have all these features from a wonderful GUI.

ftadel:
Ok for adding solutions for producing hexa meshes.

But what I understand from your code: SimNIBS does NOT produce hexahedral meshes. You just mesh the segmented volume using FieldTrip... And this block of code using ft_prepare_mesh already exists, I'd like to find an efficient way to factorize it.
The hexahedral/tetrahedral option is not properly an option of SimNIBS, it's a subsequent process.  The creation of the hexahedral mesh from the MRI is fast in comparison to the SimNIBS execution, therefore it looks like it would be better as a separate process.

1) To obtain something versatile, that can be recombined easily as new methods appear, I think we could do the following: Keep the SimNIBS approach as it is (only tetrahedral), but save the final masks in the database (_final_contr.nii.gz). And later, if we want the FieldTrip hexahedral mesh, we right-click on the mask > Create hexa FEM mesh with FieldTrip.
The 'fieldtrip' option would also create a similar volume mask (just like the existing process_ft_volumesegment.m does, but maybe grouping all the masks into one file?).

2) I don't see the point of saving an extra cortex surface coming from the hexahedral mesh. If we want this, we don't need FreeSurfer or CAT, we just mesh the segmented masks produced by SPM12 in  a few minutes...
Am I missing something?

tmedani:
> Ok for adding solutions for producing hexa meshes.
thanks, regarding Carsten answers, since the hexa vs tetra is not solved, having both options is just great since users will have also the option to compare performance between these two methods.  

> But what I understand from your code: SimNIBS does NOT produce hexahedral meshes. You just mesh the segmented volume using FieldTrip... And this block of code using ft_prepare_mesh already exists, I'd like to find an efficient way to factorize it.

indeed, just putting all the pieces together in an optimal way. 
 
> 1. To obtain something versatile, that can be recombined easily as new methods appear, I think we could do the following: Keep the SimNIBS approach as it is (only tetrahedral), but save the final masks in the database (_final_contr.nii.gz). 

sounds good and better, since we will need also the masks for other usages (DTI process).

>And later, if we want the FieldTrip hexahedral mesh, we right-click on the mask > Create hexa FEM mesh with FieldTrip.

also agree... 

>    The 'fieldtrip' option would also create a similar volume mask (just like the existing process_ft_volumesegment.m does, but maybe grouping all the masks into one file?).

I'm not sure that the quality of the masks is the same. 
all of these methods use some differents values on the SPM parameters for the segmentation, and it seems some other methods introduce modification on the output masks (correction and smoothing)


> 2\. I don't see the point of saving an extra cortex surface coming from the hexahedral mesh. If we want this, we don't need FreeSurfer or CAT, we just mesh the segmented masks produced by SPM12 in  a few minutes...
>     Am I missing something?

I added this option just to see how is the quality of the cortex in the case of the hexa, most of the peoples smooth this surfaces before putting it on their papers.  
so you can remove it.

We need also to update the function that extracts the surface from the mesh, it works for ttra but not for hexa. 
A quick way is to convert the hexa to tetra and then run the same process. 
I will work on that asap. 

@ftadel 
 I noticed a slight bug on the option 'swap face' on the corte_fem 
I did not manage to solve it, it's in the historic concatenation, could you please check it.

also, the 'Hexa mesh display' seems to be darker (maybe related to the normals), 
I don't check how can I fixed it. 

Thank you.

ftadel:
I've reorganized the code that calls SimNIBS: https://github.com/brainstorm-tools/brainstorm3/commit/4e37f910777316b8c93d0e73023dcbced7335957

I moved the code to read the SimNIBS output to an independent function import_anatomy_simnibs, and now a SimNIBS folder can be importer using the menu "Import anatomy" folder, like the other segmentation programs we support (FreeSurfer, BrainSuite, CAT,...)
I also added the import of the masks (_final_contr.nii.gz) which can now be displayed on top of the MRI, really good for debugging as well.

image  
image

image

ftadel:
More updates: https://github.com/brainstorm-tools/brainstorm3/commit/4bed91100fb9daad5710be7925ace130d956fc37

- Tissue volume saved during SimNIBS import
- Tissue volume generated with FieldTrip MRI segmentation process and popup menu
- New process and menu to generate hexa mesh with FieldTrip from the Tissue volume
- In process_generate_fem: the FieldTrip option calls successively these two new processes

ftadel:
> I noticed a slight bug on the option 'swap face' on the corte_fem
> I did not manage to solve it, it's in the historic concatenation, could you please check it.

History is not a string... always use bst_history to manipulate it.

> also, the 'Hexa mesh display' seems to be darker (maybe related to the normals),

I flipped the tetrahedrons and it looks good now.

Last commit corresponding to this pull request: 
https://github.com/brainstorm-tools/brainstorm3/commit/52ef40544c13f74820beb367901a821525c6dae0

tmedani:
Hey @ftadel , 

When calling the option 'Generate BEM Surfaces' the fieldtrip option is called.

Also with the "field trip"  the cancel button is not working.

ftadel:
> When calling the option 'Generate BEM Surfaces' the fieldtrip option is called.

This because there was no cortex available, therefore the brainstorm version is not possible. But I agree it's confusing... So I changed the logic and now always display both options, but display when selecting Brainstorm if not cortex is available.

> Also with the "field trip" the cancel button is not working.

Fixed: https://github.com/brainstorm-tools/brainstorm3/commit/57a4e728f93c7910172d73f6adca727a55aae988

tmedani:
Thank you @ftadel 

indeed, I have too many testing protocols and subjects, and in the tested subject there is no cortex. 

I will check this. 

Also, it's also good to have the segmented tissue in brainstorm database, 
this is needed for the anisotropy estimation, 
and in the next future, we will use these data for the modern FEM methods that do not use mesh (based on level set). 
but we need to keep brainstorm alive ... :)

PR #274, opened 2020-03-23
Database search: added option to hide parent nodes in search results

martcous:
There are many small improvements and fixes in this PR, but the main one I would like to get help testing is the "hide parent nodes from search" option in searches. It allows users to select all files of the same type at once to perform an operation on them.

Search GUI:
image

Search result with parent nodes shown (default):
image

Search result with parent nodes hidden:
image

You can now select all files of a same type at once (SHIFT + click or CLICK + click) in order to perform an aggregated operation on them:
image

Thanks!


toolbox/gui/panel_search_database.m
```diff
@@ -86,6 +86,10 @@
     
     %% Buttons
     jPanelBtn = gui_component('Panel');
+    % Hide parent nodes
+    jPanelBtnNorth = java_create('javax.swing.JPanel');
+    jCheckHideParent = gui_component('CheckBox', jPanelBtnNorth, [], 'Hide parent nodes from search?');
```
ftadel: Checkbox => no question mark at the end of the label
Maybe just "Hide parent nodes"?

Could you align this box  with the controls below?

toolbox/gui/panel_search_database.m
```diff
@@ -740,15 +750,15 @@ function RemoveOrSeparator()
 
 % Hard coded search type values and their associated dropdown label
 function [labels, values] = GetSearchTypeValues()
-    labels = {'Channel', 'Data', 'Dipoles', 'Fibers', 'Folder', ...
-        'Head model', 'Kernel', 'MRI', 'Matrix', 'Noise covariance', ...
-        'Power spectrum', 'Raw data', 'Source', 'Statistics', ...
-        'Subject', 'Surface', 'Time-frequency', 'Video'};
+    labels = {'Channel', 'Data', 'Dipoles', 'Fibers', 'Folder (non-raw)', 'Folder (raw)', ...
+        'Folder (all)', 'Head model', 'Kernel', 'MRI', 'Matrix', 'Noise covariance', ...
+        'Power spectrum', 'Raw data', 'Source (full map)', 'Source (link)', 'Source (all)', ...
```
ftadel: Confusing: sources can be full, kernel or link; not only link/non-link.
Maybe "Source (file)" vs. "Source (link)" ?

toolbox/gui/panel_search_database.m
martcous: I aligned the checkbox to the right, is that what you meant?

toolbox/gui/panel_search_database.m
martcous: I now have Source (file) vs Source (link) vs Source (kernel). I removed "all".

ftadel: I'm not sure this is how the tool would get used.... This is too complicated, and kernel vs full is not an ontological difference, it's a storage optimization. 
Either 'file'=full+kernel and 'link', or just "Source"=everything.

toolbox/gui/panel_search_database.m
ftadel: No, below = at the bottom, left of the validation buttons
This way it doesn't take any extra space...

toolbox/gui/panel_search_database.m
martcous: I think there's a big difference between kernel and full, people don't use kernels in the process box but use both link and full maps instead. This is why I had initially grouped both links and full maps. Maybe we're not using the same terminology here, I am using the node type names in the database, see the image below:

image

I can also just go back to source = everything if that's clearer.

Moo-Marc: One use case is to delete large files when archiving a protocol (what I was doing last week).  In that case, one may want to only select full maps.  There aren't as many kernels, so I wanted to keep those.

toolbox/gui/panel_search_database.m
martcous: Is this better?

image

ftadel: I think so, but that's totally arbitrary :)
(More expression of my OCDs: I think I would align it on the left instead of the right, so that it's on the side of the options we might not need instead of the buttons we need to click...)

toolbox/gui/panel_search_database.m
ftadel: @martcous Oh, I see... 
What you called "kernel" here, I would call "shared kernel".
What you called "full map" can be either a "kernel" or a "full map", depending on if you select the option "Full results" in the min norm option panel. By default, you get a file with the tag "KERNEL" if you right-click on a single trial > Compute sources. This is only referring to the optimized storage of a source file (ImagingKernel+Data) instead of the full flat source matrix (ImageGridAmp).
These KERNEL files become full maps when any non-linear operation is applied to them. Therefore I was not sure about the semantic relevance of this full/kernel distinction.

But as Marc mentioned, it could be useful to catch big files.
Unless.... you add a new search parameter to search files by size and/or creation dates :-)

I don't think we care much about having a few extra menus that almost never used in this list, but we need to find some wording that users can understand.

Is this too much?
Sources (files):  physical results_*.mat files, both kernels and full maps
Sources (links): only links, not real .mat files attached
Sources (full maps): Only the files that have ImageGridAmp defined (if @Moo-Marc thinks it's useful and if we can't filter otherwise by size)
Sources (all):  files+links
Sources (shared kernels): Only these white brain icon things that can't be selected in Process1 otherwise - at the moment, there is no way to select and delete a shared kernel from a script...

ftadel:
@martcous Feel free to merge when you're done

toolbox/gui/panel_search_database.m
martcous: To the left it goes :)

toolbox/gui/panel_search_database.m
martcous: Thanks François, that's clearer. Since the search is based solely on the node types, I'll stick to: Source (file), Source (link), Source (all) and Source (shared kernel).

martcous:
Thank you both for your comments! I have taken note of your suggestions left, and I will add them to my backlog. I'll go ahead and merge this for now since what is left to do requires major changes.

ftadel:
Some bug fixes:
https://github.com/brainstorm-tools/brainstorm3/commit/a2bd30168277aada2087b0f2fb4229d67f7d1fee

PR #278, opened 2020-04-14
modify default sMri Comment value

juangpc:
Hi, 
I've come across this little bug while importing a freesurfer folder. The mri importer was unable to set the name of the mri properly and thus, the sMri.Comment field remained with the default value at the beginning of import_mri.m.

The effect of keeping the default value in Comment is that when setting up the gui, this sMri.Comment field will be used to set the name of the MRI in the tree and the java method setText will not accept the default value of (double)1 (actually it accepts it, but since 1 is non-printable... it prints a square!). 
A more important problem rises when you try to righ-click on the scalp node. It crashes.

**Context:** If you want to replicate, send me an email and I can send you a link to download the fs folder that triggered this error. Otherwise with an mri making bst's mri importer to be unable to set the Comment field properly.

**Changes I propose:** 
- defulat Comment value. Why would you want the default value for a string to ever be a double?
- The logic for error checking of ```sMri.Comment``` value. 
 In line 270 (io/import_mri.m) there is this comment ```% Add a Comment field in MRI structure, if it does not exist yet``` which was not the logic implemented at all. I made it so that it checks if the field exists, and if it desn't, it creates it with the value of a valid Comment variable.
If it exists, it checks if it is valid in order to avoid the problems of having non-string values, which will be later used as input to java methods (gui).

@tmedani thanks for your help.

### DIFF
--- a/toolbox/io/import_mri.m
+++ b/toolbox/io/import_mri.m
@@ -47,7 +47,7 @@ if (nargin < 5) || isempty(isAutoAdjust)
     isAutoAdjust = 1;
 end
 if (nargin < 5) || isempty(Comment)
-    Comment = 1;
+    Comment = 'MRI';
 end
 % Initialize returned variables
 BstMriFile = [];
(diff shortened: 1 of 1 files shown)


ftadel:
I agree with changing the default MRI comment, this value "1" doesn't make any sense, I probably didn't finish to edit my copy-paste... I fixed this and the index of the parameter in this commit:
https://github.com/brainstorm-tools/brainstorm3/commit/c87609a231c34cf933ca79d4570da840004772d8

I'm sorry it breaks your PR, but I was not sure about rest of the PR:  the change you proposes modifies the logic of this input Comment field. Its goal was to override the existing comment from the imported file, which I think is not possible anymore with your code. See this call for instance:
https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/io/import_anatomy_simnibs.m#L170

Let me know what else you need to be added in order to make your FreeSurfer folder work.
No need to add more tests about this parameter "Comment", it is supposed to be used only internally.
If one of the in_mri_* functions may return something that is not digestible by import_mri.m, I'd rather fix the low-level function than adding too many tests that are in practice never used in import_mri.m. 
(in theory, it would be good to always test the types of all the input arguments, as Matlab is not a language that do not allow any form of strict prototyping, but I personally think that in practice it makes the code less readable and heavier to maintain, and has very little added value for functions that are not called by brainstorm users, but only by brainstorm itself)

juangpc:
Ok. I'll test the folder only with master + modifying the default value of Comment. I'll write back.

juangpc:
Thanks. It works just fine.

Closing...

PR #279, opened 2020-04-15
Fix import fiff missing ssp names

juangpc:
Hi, 

I've encountered a file wich had ssp vector information in it, but the labels for each projector vector had been removed. This is all related to the way in which those labels are stored in a fif file. I think Matti knows about this (see image below) but either way, this small modification will ensure that the file can be used even when the actual labels of the projectors are missing. Maybe it can come in handy in the future.

image

### DIFF
--- a/external/mne/matlab/fiff_read_proj.m
+++ b/external/mne/matlab/fiff_read_proj.m
@@ -87,7 +87,6 @@ for i = 1:length(items)
             error(me,'Projection item description missing');
         end
     end
-    
     tag = find_tag(item,FIFF.FIFF_PROJ_ITEM_CH_NAME_LIST);
     if ~isempty(tag)
         namelist = tag.data;
--- a/external/mne/matlab/fiff_read_proj.m
+++ b/external/mne/matlab/fiff_read_proj.m
@@ -74,7 +74,11 @@ for i = 1:length(items)
     end
     tag = find_tag(item,FIFF.FIFF_DESCRIPTION);
     if ~isempty(tag)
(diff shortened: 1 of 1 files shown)


external/mne/matlab/fiff_read_proj.m
```diff
@@ -74,7 +74,11 @@
     end
     tag = find_tag(item,FIFF.FIFF_DESCRIPTION);
     if ~isempty(tag)
-        desc = tag.data;
+        if(isstring(desc))
+          desc = tag.data;
+        else
+          desc=['proj_item_' num2str(i)];
```
agramfort, line 80:
```suggestion
          desc = ['proj_item_' num2str(i)];
```

external/mne/matlab/fiff_read_proj.m
```diff
@@ -74,7 +74,11 @@
     end
     tag = find_tag(item,FIFF.FIFF_DESCRIPTION);
     if ~isempty(tag)
-        desc = tag.data;
+        if(isstring(desc))
```
agramfort, line 77:
```suggestion
        if isstring(desc)
```

martcous: It would be best to avoid using the [isstring](https://www.mathworks.com/help/matlab/ref/isstring.html) function altogether since it was only introduced in Matlab 2016b. Maybe [ischar](https://www.mathworks.com/help/matlab/ref/ischar.html) has an equivalent behavior in this case?

PR #280, opened 2020-04-17
Import graph events alternate style

juangpc:
Hi, 
This PR makes bst compatible with a file format used in Elekta-Neuromag sites to store events. It is a text file with the same *.evl extension, although it slightly differs from the default .evl file.

I have been told that it is not only here in Houston where they use these files, quite a few other Elekta-MEG clinical labs, like Cleveland for instance, use it. This is why I'm proposing this PR. 

The default *.evl file is a text file which has the following structure: 
```
(graph::saved-event-list 
 :source-file "/path/file.fif"
 :events '(
  ((:time  28.8425) (:class :manual) (:length  0.069))
  ((:time  31.194) (:class :manual) (:length  0.0415))
))
```

This other file format (also with *.evl extension) has the following structure.

```
(graph::saved-event-list 
 :source-file nil
 :events '(
  ((:time  42.463) (:class "A") (:level  2.35578691971e-11))
  ((:time  42.463) (:class "A") (:level  1.123456e-11))
  ((:time  47.561) (:class "A") (:level  2.14289228723e-11))
))
```


ftadel:
Most of the code is actually the same in in_events_graph.m/in_events_graph2.m and out_events_graph.m/out_events_graph2.m

You could efficiently factorize most of this code and avoid creating new files by adding a parameter to the existing functions, ie. variable Version=1 or 2, which would just branch with an "if" or "switch" statement where the two versions differ. Add the documentation of the two versions in the header.

Would this make sense to you?

juangpc:
> Most of the code is actually the same in in_events_graph.m/in_events_graph2.m and out_events_graph.m/out_events_graph2.m

Actually, there is no out_events_graph.m. It is my plan create it, but since I don't have easy access to elekta's Graph now, and I want to test if the file is successfully being read, and also I want to test if you can create instantaneous events, not only segments, I decided to wait.

> Would this make sense to you?

Yes, it would. I thought about that too. And I'm intrigued by your answer. You know how much I'm interested in being a better sw developer (not to be confused with scientist-writing-code 🤦). Been reading about how to deal with code complexity in very long-complicated projects. And one of the key things I've learned is to try to always go for: "interface segregation".  It is also part of the (in?)famous solid principles. That is why I proposed it like this. You don't have to change anything, you don't need to remember anything, it is just a new option... Plus, maybe... one day, some MDs will use this option, at least that is the idea. Imagine explaining about version number you need to change to an out-of-time stressed MD? I don't. So: (1) whatever you decide, I'll update/push no problem. (2). Please elaborate. 
Thanks, Francois, take care.

ftadel:
Well.... this seems to be related with two different versions of the same software Elekta Graph, and two minor variants of the same file format (LISP-based).
One function for reading the Elekta Graph files seems appropriate to me, with an additional option to tune the actual version of Graph, to adapt to small differences between the versions. Writing two methods for the same software and same file format does not seem to be appropriate to me, but this is a personal point of view. 

Since, we already have dozens of reading functions that switch internally for the version of the format, I suggest we keep the same convention and deal with reading everything Graph-related in one single function.

Reducing the number of code lines is also an interesting objective when we are very few people maintaining a large code base...

juangpc:
Ok. But I want to abstract the reading of the file without having to modify it.
How about a try catch structure in in_events_graph.m ?
in_events_graph.m starts assuming default Graph version, and if it fails, it will try the alterantive version and if it fails, it will be collected by bst_call.m
I see a problem with that in the sense that there is already a try-catch scheme implemented with bst_call. But that would allow us to abstract the reading of different file versions. What would the problems with that? 

And how to abstract the file export? Eventually we want MDs to be able to generate events with brainstorm and then read them back with Graph and xfit, etc... with this alternate graph event file they are used to.

ftadel:
You can have multiple levels of try-catch, no problem about that. It's a common thing to rethrow an error when it can't be handled locally. 

You can also still use two different entries in the list of file formats, like you did (Graph and Graph2), and having the switch in import_events that point at the same function with a different input parameter. This is already the case in import_channels.m for instance, which calls multiple times in_channel_ascii in various ways.

Do you really need to export back the events to the Graph format?
If you don't need it and no one ever asked for it, it might not be necessary to spend too much time on it.

juangpc:
> Do you really need to export back the events to the Graph format?

Yes I do. In order to give the MDs the ability to review a file with BST and then go back and see with Graph the events identified initially.

juangpc:
Been trying to follow your instructions. 

- in_events_graph2 no longer exists.
- in_events_graph iterates through an array of fcn handles defining each version to read.
- newer versions would imply just to add another function.
- Wanted to preserve the error messages that is why I concatenate the errors in the final message if there was any.
- out_events_graph.m is prepared for when someone needs the default format. but it will now be used just with this alternative style.

ftadel:
I was suggesting to keep the two entries in the list of formats, but if it works this way it's even better!
Thanks!

PR #281, opened 2020-04-20
Moving bst-duneuro Matlab code to Brainstorm repo

ftadel:
I integrated all the Matlab code from bst-duneuro in Brainstorm.
The main objective is for being able to distribute it with the compiled version of Brainstorm: all the code that is not compiled together with the main project can't be executed after. It's not possible to have "plugins" in this compiled Brainstorm.

Most functions were called only at one place, but were requiring lots of input structures to be rewritten and tested at the beginning of each function, all the parameters ended up having 3 or 4 different names across functions, it was difficult to follow. 
Since the only goal of this Matlab code is to use it from Brainstorm, I linearized everything made it more consistent with the rest of the application. It resulted three main functions: bst_duneuro, duneuro_install and duneuro_defaults. Plus a few additional functions for IO, which I moved to the toolbox/io folder, with the same calling and naming conventions as the others.
The total volume of code is now much smaller.

@tmedani @juangpc Sorry if you can't recognized much of the code that you wrote...

Along the way, I removed most of the test/todo lines.
Here is a list of the comments you might be interested to keep, or add again to the code:

```
https://docs.google.com/spreadsheets/d/1MqURQsszn8Qj3-XRX_Z8qFFnz6Yl2-uYALkV-8pJVaM/edit#gid=0

% TODO : need to fixe ... it seems that duneuro uses 3 orientations per
% coil and brainstrom only one orientation .... maybe it needs local
% projection (scalar product to the correct orientation)
		
% TODO : Check if the geo file could be used for the hexa

% TODO : Optimisation ... use binary input/output
%               Use a standard format for all modalities
%               Discuss with duneuro team to add the I/O files

(bst_duneuro)
% TODO : Adapt this for anisotrpy // maybe could be done from outside and
% put the tensor on OPTIONS.Conductivity or create new field OPTIONS.ConductivityTensor

% % the final value ==>  cfg.fem_meg_lf = B = Bp + aBs ? check if the case for bst

(bst_prepare_head_model)
%% @@@ WARNING @@@
% For the combined MEEG .... if the user select specific layers for the MEG
% the new model will be also used by the EEG .... and this is not correct
% FOR the MEEG and cfg.layerToKeep contient des zeros .... == > add
% condition that generate two head model and run duneuro for each
% modalities. ==> Open an issue ... this could be also a proble for
% openmeeg 

(out_fem_dgf)
% TODO : Check if the index of the nodes and the element -1 (index on cpp)

(bst_postprocess_lf)
%TODO : check the minifile parameters and adapt this code
```

Additional todo from me:
- out_fem_*: Could not test the output of .geo, .dgf or .knw files...


ftadel:
At this moment, the options panel look like this (depending on the options selected):

image

image

image

image

juangpc:
No problem on my side. Totally understand. 
Will review today.

tmedani:
Hey Francois, 

Thank you for all the efforts and sorry if you found some difficulties in this toolbox.

I totally agree with all the changes and all modifications that make the code well organized and used. 

Regarding all the todo tasks, most of them are fixed, I will update it. 

I will review this new version asap. 

Best

ftadel:
@tmedani @juangpc I made all the modifications we discussed by email in the commits above.

Could you check that this version still gives exactly the same results as before?
Please re-compute at least one FEM forward model you already have in your database and make sure my code rework produces files that are mathematically equivalent.

I have rewritten almost everything, so I wouldn't be surprised if I made at least one mistake somewhere that changes the results. I apologize in advance if it's a pain to debug...

The new panel looks like this (removed fitted/unfitted and GeometryAdapted options, added Help button):
image

ftadel:
@juangpc @tmedani Can I merge this? 
Or will break things that you really need to be working now?

ftadel:
Ok, thanks!
then I will merge it and we'll keep on working on the master branch directly.

tmedani:
> I've tested with a subject and everything went great.
I did not check the last updates, but when testing with @juangpc 
the cfg.node and cfg.elem were missing and we corrected it manually without submitting the change. 

I will pull the new version and check.

> Not incorrect, but if the idea is to compare head models, maybe we could define the conductivity values according to the values used in previous head models computed. Or maybe it is better to let the user read that in the tutorials. Or maybe change the default ones to match? @tmedani ?

yes, as discussed before, with 3-shell sphere and Overlapping sphere, Brainstorm does not ask the user to set the conductivity values and uses the default values 0.330 0.004 0.330.
in order to compare the head models, we need to use the same values.

PR #282, opened 2020-04-22
generate FEM tensors from DWI

tmedani:
Add the item to the right-click menu 
that calls the process to generate the FEM tensors.

@ftadel following your [recommendation](https://github.com/brainstorm-tools/brainstorm3/issues/267#issuecomment-617096797), 

The function "bst_generate_conductivity_tensor" is located on the bst-duneuro repo (need to pull the new release).
There are also the functions that call the Brainsuite process, conversion of the tensor from voxel to scs and the function to display tensors as an ellipse on the mesh.

All these functions are working as they are but still under development/improvement/cleaning.  

In order to call this function from brainstorm, right-click on the MRI and then "Generate FEM tensor"

The installation of the Brainsuite software http://forums.brainsuite.org/download/ is required.

You need to have the DWI data (nifty, bval and bvec). 
As it's implemented now, you need to have all these files in the same folder and have the same name (s01.nii, s01.bvec and s01.bval). 
You need just to select the nii file during the uiopen is called. 

There is no output to brainstorm database, you need to add a breakpoint on the function to see what is going on. 

The main output data needed from this process is either (or both) the :
- DTItensor on each element
- Conductivity tensor en each element

### DIFF
--- a/toolbox/tree/tree_callbacks.m
+++ b/toolbox/tree/tree_callbacks.m
@@ -1028,6 +1028,7 @@ switch (lower(action))
                     end
                     if ~isAtlas && (length(bstNodes) <= 2)
                         gui_component('MenuItem', jPopup, [], 'Generate FEM mesh', IconLoader.ICON_FEM, [], @(h,ev)bst_call(@process_generate_fem, 'ComputeInteractive', iSubject, iAnatomy));
+                        gui_component('MenuItem', jPopup, [], 'Generate FEM tensor', IconLoader.ICON_FEM, [], @(h,ev)bst_call(@bst_generate_conductivity_tensor, iSubject, iAnatomy));
                     end
                     if ~isAtlas && (length(bstNodes) == 1)
                         AddSeparator(jPopup);


ftadel:
> The function "bst_generate_conductivity_tensor" is located on the bst-duneuro repo (need to pull the new release).

FYI, this is not in bst-duneuro, but in you personal repo duneuro_interface: 
https://github.com/tmedani/duneuro_interface/tree/master/inProgress

Are the functions available there the latest versions you have?

tmedani:
> FYI, this is not in bst-duneuro, but in you personal repo duneuro_interface:
> https://github.com/tmedani/duneuro_interface/tree/master/inProgress


indeed, I did not push the change to git, it was only on my local version.


> 
> Are the functions available there the latest versions you have?

all the functions related to the DWI/DTI are still under development, 
I'm still working on them and need more inputs, testings, and validations. 

I will push my updates regularly and it nice to keep it on the bst-duneuro for now in order to avoid pull requests to bst-master. 

@ftadel  do you agree ?

ftadel:
> I will push my updates regularly and it nice to keep it on the bst-duneuro for now in order to avoid pull requests to bst-master.
> @ftadel do you agree ?

OK. 
Is it a bad idea if I start working on their organization in the Brainstorm framework at the same time? So you work on code that is closer to the final implementation, and maybe easier for you to test correctly with the correct file formats/coordinate systems?
(eg. I'd like to remove your calls to ft_read and your conversion functions from FieldTrip to Brainstorm)
That would require that I also update your code on bst-duneuro, maybe just for a few days.
Are you working on it continuously?

tmedani:
> Is it a bad idea if I start working on their organization in the Brainstorm framework at the same time?

You can work on the organization on Brainstorm side and then I can work directly on brainstorm.

In the meantime, I will do all the testing on bst-duneuro and I think the adaptation for the final version will be easier.

tmedani:
Hi @ftadel , 

In order to test the functions that map and convert the tensors  

You could add these two lines to  tree_callbacks at line 1192/1193 
section : %% ===== POPUP: FEM HEAD MODEL =====
                    gui_component('MenuItem', jPopup, [], 'Generate FEM tensor', IconLoader.ICON_FEM, [], @(h,ev)bst_generate_fem_tensor(iSubject));
                    gui_component('MenuItem', jPopup, [], 'Display FEM tensor', IconLoader.ICON_FEM, [], @(h,ev)bst_view_tensor(iSubject));

It will allow to have two more items for the FEM mesh that either convert or display the tensors. 

As discussed before, this is not a final version but just testing and discuss how to integrate all of this. 

I tried to make it as easy as possible to understand, but the code is not optimized.

The dependencies are all on the bst-duneuro/inProgress 

Let me know if it works for you or I can create a new pull request. 

Thanks

tmedani:
quick explanation

first, you need to have the EIG-DTI already on the subject, then you can call the function 'Generate FEM Tensor' by right click on a FEM mesh (tetra)

After this process, a new field 'tensor' is added to the femHead structure, it contains the FEM tensor information.
as said this is just for testing purposes.

image


After this, you can call the function that displays these tensors, 
for now, the display is outside brainstorm, it still needs to be improved and make it faster. 

For testing, you can generate lower FEM mesh (faster)

I will continue working on these functions to remove all not useful parts for brainstorm.

ftadel:
Thanks for reorganizing the code.
I am sorry, I got caught in other things this week, I will start exploring and reformatting it next week. 

There is one last thing that I don't understand well: I though that what we wanted to display was the  DTI-EIG file, but it appears it is the output of bst_generate_fem_tensor.
Why do we need the DTI-EIG file for?
Is it really interesting to have it saved in the database for anything? Do we want to display it?

If all the displays and computations are based only on the "tensors" field you added to the FEM mesh file, then we might not need to save it.
Instead of right-clicking on the subject > Convert DWI to DTI, couldn't we directly right-click on the FEM mesh > Compute FEM tensors? This menu would ask for the DWI files and do the two steps at once (DWI>DTI and computation of FEM tensors), skipping the necessity of saving the large DTI-EIG file.

Am I missing something?

tmedani:
Hi @ftadel, 




> I am sorry, I got caught in other things this week, I will start exploring and reformatting it next week.

no problem, you are doing too many things on different topics ... 

> There is one last thing that I don't understand well: I thought that what we wanted to display was the DTI-EIG file, but it appears it is the output of bst_generate_fem_tensor.

at the end yes, it's the most useful for the users of the FEM.

> Why do we need the DTI-EIG file for?
> Is it really interesting to have it saved in the database for anything? Do we want to display it?

We had a similar discussion here, where we discussed how it's useful to have the 4D
https://github.com/brainstorm-tools/brainstorm3/issues/267#issuecomment-619802730

we end up to the conclusion to add the 4d since it can be useful for fmri. 
and we can add the option to display the DTI tensors on the MRI (as said before low priority)

 
> If all the displays and computations are based only on the "tensors" field you added to the FEM mesh file, then we might not need to save it.
> Instead of right-clicking on the subject > Convert DWI to DTI, couldn't we directly right-click on the FEM mesh > Compute FEM tensors? This menu would ask for the DWI files and do the two steps at once (DWI>DTI and computation of FEM tensors), skipping the necessity of saving the large DTI-EIG file.

Yes, this is all the process, it was the first version of the function bst_generate_conductivity_tensor.m 
https://github.com/brainstorm-tools/brainstorm3/pull/282/files

but you propose to divide the process into two subparts.

by the way, we can also integrate it to fem_process and then call all of this once from the same function. 


Check the rest of the process as it's now and let me know which way you think is the best. 

Maybe you need to pull the new modifications that I may add these days.

Have a nice weekend.

ftadel:
I'm trying to understand what is going on in bst_generate_fem_tensors.m (migrating towards process_fem_tensors.m). There are a few options for which I had coded in a panel, but I realized after that they have no effect... Is this expected?...

Example:
AnisotropyModel =  {'Effective Medium Approach [EMA]','EMA with Volume Constraint','Artificial/simulated anisotropy'}
Only the first (isArtificial=0) and the last (isArtificial=1) seem to be actually used.

Could you confirm the options that you need, and their dependencies?
(I'm not keeping any "future options" here, only things that are already wired to things that are already happening).

-----

So far, we have a new popup menu on the FEM meshes: "Compute FEM tensors" (and the corresponding process: process_fem_tensors.m)
image

It offers the choice for each tissue to select a constant isotropic value, or something anisotropic computed from the DTI:
image

Please tell me if this looks good to you, and what other options you need and where.

ftadel:
Something like this?
image

What are the ranges of values and precision needed for the **conductivities** and **ratio** options?

ftadel:
```
            % Call the main function : bst_compute_anisotropy_tensors
            % METHOD 1 : direct transformation approach [Güllmar et al   NeuroImage 2010 & Tuch]
            % METHOD 2 : direct transformation approcah with volume constraint  [Güllmar et al NeuroImage 2010]
            % METHOD 3 : Artificial anisotropy with volume constraint [Güllmar et al NeuroImage 2010]
            % METHOD 4 : The volume normalized approcah  [Won Hee Lee et all  IEEE 2015]
            % METHOD 5 : Wang-constraint model and volume-constraint model  [Wolters CH 2006]
            % METHOD 6 : Wang-constraint model and volume-constraint model  [Wolters CH 2006]
            % METHOD 7 : As SimBio toolbox  [Rullmann et al 2008 / Vorwerk et al 2014 ]
            usedMethod = [2, 7];
            options.aniso_method = usedMethod(anisotropyOption.anisotropeMethod);%  [ 1, 7]
```

What did you intend to do here?
There are 3 possible values for "anisotropeMethod" (1,2,3), but only 2 in "usedMethod".

We will change this to labels, indices are impossible to manage.
Please let me know which methods you'd like to keep, and what label you want to give it (something simple, a few letters)

tmedani:
> ```
>             % Call the main function : bst_compute_anisotropy_tensors
>             % METHOD 1 : direct transformation approach [Güllmar et al   NeuroImage 2010 & Tuch]
>             % METHOD 2 : direct transformation approcah with volume constraint  [Güllmar et al NeuroImage 2010]
>             % METHOD 3 : Artificial anisotropy with volume constraint [Güllmar et al NeuroImage 2010]
>             % METHOD 4 : The volume normalized approcah  [Won Hee Lee et all  IEEE 2015]
>             % METHOD 5 : Wang-constraint model and volume-constraint model  [Wolters CH 2006]
>             % METHOD 6 : Wang-constraint model and volume-constraint model  [Wolters CH 2006]
>             % METHOD 7 : As SimBio toolbox  [Rullmann et al 2008 / Vorwerk et al 2014 ]
>             usedMethod = [2, 7];
>             options.aniso_method = usedMethod(anisotropyOption.anisotropeMethod);%  [ 1, 7]
> ```
> 
> What did you intend to do here?
> 
> We will change this to labels, indices are impossible to manage.
> Please let me know which methods you'd like to keep, and what label you want to give it (something simple, a few letters)


Hi @ftadel, 

yes, I spent some time reading and check most of the available methods then I have implemented most of them. 

These functions were independent initially and then I merged them in order to call these methods. 

From the 7 methods, I selected only the method 2 and 7, and the artificial was re-implemented with other options. 

> There are 3 possible values for "anisotropeMethod" (1,2,3), but only 2 in "usedMethod".

the third one is artificial and it's outside of this function since it uses other parameters, sorry for the complex arrangement. 

I can work on top of what you did in order to simplify the call.

ftadel:
> From the 7 methods, I selected only the method 2 and 7, and the artificial was re-implemented with other options.

So you do need the three "Anisotropy methods"?
=> 'Effective Medium Approach [EMA]'  =  Method 2 = '**ema**'
=> 'EMA with Volume Constraint' = Method 7 = '**emavol**'
=> 'Artificial/simulated anisotropy' = no method = '**simulated**'

Please look carefully at the screen captures I send, because it doesn't match what you're saying here.

> sorry for the complex arrangement.

It is indeed quite cumbersome...

> I can work on top of what you did in order to simplify the call.

No, just answer carefully all the questions I ask as I go, and I'll do all the code reorganization.

ftadel:
(PS: I'll remove all the code that is not used - this includes all the unused methods in bst_compute_anisotropy_tensors.m)

tmedani:
> I'm trying to understand what is going on in bst_generate_fem_tensors.m (migrating towards process_fem_tensors.m). 

agree, these two processes are highly related, 

>There are a few options for which I had coded a panel, but I realized after that they have no effect... Is this expected?...

I'm not sure that I understood this question?

> Example:
> AnisotropyModel = {'Effective Medium Approach [EMA]','EMA with Volume Constraint','Artificial/simulated anisotropy'}
> Only the first (isArtificial=0) and the last (isArtificial=1) seem to be actually used.

The two first methods  ('Effective Medium Approach [EMA]','EMA with Volume Constraint' ) are not artificial, they need the DTI data in order to generate the tensors and here isArtificial = 0;

The third one is artificial, where the tensors are simulated from the artificial direction (normal and tangential vectors to the element), and the simulated eigenvalue from the ratio = longitudinal/transversal that the user needs to set.

I have added an extra option to the artificial method where the direction could be extracted from the DTI. In this case, the orientation is real but the values are simulated. 
But as you said, this is a strange option that can be removed for now. 

Furthermore, most of these methods need the value of the conductivity, which means that we have to know apriori the conductivity value from this step... so they will not be reset during forward model computation.  


> Could you confirm the options that you need and their dependencies?
> (I'm not keeping any "future options" here, only things that are already wired to things that are already happening).

agree, let me draw a detailled scheme for all the process and share it with you.

tmedani:
> (PS: I'll remove all the code that is not used - this includes all the unused methods in bst_compute_anisotropy_tensors.m)

ok,

ftadel:
Just discovered that parameter "ratio" is actually useless...
I keep deleting what I've been coding today... not very efficient.

tmedani:
> Just discovered that parameter "ratio" is actually useless...

why? I think it's relevant and needed for the artificial anisotropy
 
> I keep deleting what I've been coding today... not very efficient.

I think a big part of it will be reformulated and re-organized.

ftadel:
> > Just discovered that parameter "ratio" is actually useless...
> why? I think it's relevant and needed for the artificial anisotropy

I got confused with all these parameters that change names between functions and deleted the wrong things... I realized that, restored the code and deleted that comment :)

> I think a big part of it will be reformulated and re-organized.

The volume of code is shrinking like neige au soleil :-)

tmedani:
Hi @ftadel 

Here is a simple schems that shows the main steps. 

image

tmedani:
Hi

> I got confused with all these parameters that change names between functions and deleted the wrong things... I realized that, restored the code and deleted that comment :)

Indeed, I have coded all this at different times and I did it on the previous toolbox (duneuro-interface), then I linked altogether, it works bu not clean :(

> The volume of code is shrinking like neige au soleil :-)

hhh, summer soleil, 

indeed, once we have the right idea and the right code it will be clean and just in a few lines. 

Regarding my schema, here is a quick explanation : 
-1- set iso conductivity: this is required at this level since the tensors will be normalized according to these values. 
-2- tissues to considered as aniso: not all of the tissues are anisotropic. The most important one is the WM.
Some publications consider also the GM as anisotropic, but it seems to be difficult to estimate it from the DWI, since it's highly affected by the CSF tissue. 

So, with the DTI only the WM and GM anisotropy can be estimated (soft tissues). 

The CSF is considered isotropic, however, the skull is not. 
The anisotropy of the skull is difficult to estimate therefore most of the publications use an artificial skull anisotropy where the longitudinal direction is lower than the transverse direction. A ratio between 5 to 10 is used to model the skull tensor. However, in most cases, the skull is considered isotropic with a mean scalar value of its tensor [s = mean(trace(L))]
The scalp is considered also isotropic. 

-3- it follows what I have explained before, we need to specify which method to apply for each tissue. 
Ideally, it will be real for GM and WM and simulated for the skull, but it would be nice to leave as an option where the users can change the methods or use simulated values for all tissues, this is used to study the influence of the tissues anisotropy on brains signals and source localization.  

-4- the options for each method. 
-5- Generally the volume constraint (wolters')  is used for WM and GM and the Wang's constraint is used for the skull. 
The final results will be tensors for the whole model, all the tissues will have a tensor even the isotropic ones (diagonal matrix tensor). 

I hope this explanation can help, all the code do all of this +/- some other options.

ftadel:
Thanks for the flowchart, it really helps.

I came up with the interactive options panel described below.
There are still some pending questions, please address all the items below, check all the labels of the interface and comment if anything does not make sense:

If all the layers are set to "isotropic", all the other options are hidden:
image

1) What are the ranges of values we are supposed to accept for the conductivity values (max number of significant digits, max and min values)? Right now there is no constraint, so users can even enter letters of leave the field empty... 
2) If a layer is set to "Anisotropic": are we supposed to still be able to edit the conductivity value? At the beginning I was disabling the conductivity text box, but from your explanations, it looks like we need the conductivity in both cases ("Isotropic" and "Anisotropic"), therefore I put it before the Isotropic/Anisotropic choice.
3) Is it OK to leave all the layers to "isotropic" by default?

When selecting at least one "Anisotropic" layer, the "Anisotropy method" panel gets displayed:
image

4) Please confirm that: 
   - 'EMA' = Method 2 = [Güllmar 2010, NeuroImage]
   - 'EMA+VC' = Method 7 = [Rullmann 2008 / Vorwerk 2014], used in Simbio

When selecting "Simulated", more options appear in panel "Simulated anisotropy":
image

5) What is the range of values for the "Ratio longitudinal/transversal"? (max number of significant digits, max and min values) - I need to add constraints to this text box as well.
6) What is the default constraint? Wang's?

> Generally the volume constraint (wolters') is used for WM and GM and the Wang's constraint is used for the skull.

7) This last comment **contradicts everything** that was discussed above, everything that I coded in the past days, and all of your code organization... If you need to be able to select all the options independently for each tissue, most of what I did so far is useless, and all the data structures need to be revised again. Let's try to be clear: **do you need more flexibility than what is offered at the moment**? If so, tell me asap, so I start everything from scratch before I move forward...
=> Please try to draft the ideal interface you would like to have to select all the combinations of options you need.

ftadel:
8. Why do we need the volume with the tissue segmentation?
=> It looks like you use it only to get the size of the voxel, but this is something you could also get from the T1 volume.
=> Isn't the DTI-EIG volume created by BrainSuite always interpolated on the T1 volume? 
=> It seems to me that by construction we don't need the tissue segmentation volume at all because by construction the DTI-EIG is already registered to the T1 volume.

ftadel:
I commited a first draft, only to validate parts of the structure of the code and data with you:
https://github.com/brainstorm-tools/brainstorm3/commit/64225e6270979095976ed990c830bf0e5d463df4#diff-8096856ce21df0a8efe25c56b274105dR132

- I changed the storage of the tensors: now a matrix Tensors [nElements x 12] is stored in the FEM mesh: it will be much easier and faster to manipulate than millions of cell arrays (which get very slow for some operations, like saving data to disk...)
- Can you please check that if you right-click on a FEM mesh > Compute FEM tensors, and leave everything to isotropic, get something that makes sense?
- Same question if you select one tissue isotropic and select "Simulated"?

I'll wait for your feedback on this and the 8 previous questions on the GUI before moving further, and will finish the anisotropic case tomorrow.

juangpc:
This looks amazing. I'm starting to receive emails of people interested on the fem computation. @tmedani  @ftadel  this is great what you've done.

tmedani:
Hi @ftadel 

> f all the layers are set to "isotropic", all the other options are hidden:
> image

Yes, this is a good way. 
Furthermore, with this, we can check the performance of duneuro (execution time and results).
So if we use the previous approach with scalar conductivities duneuro uses the .msh and .con files.
When we will use the isotropic tensor (as here), duneuro will use the .geo and .knw files 
So it's nice to have these both methods for later (debugging, checking, users errors ...).




> 1. What are the ranges of values we are supposed to accept for the conductivity values (max number of significant digits, max and min values)? Right now there is no constraint, so users can even enter letters of leave the field empty...

I think it's better if we can add some tests on the values, to avoid letters and zeros/nan and void values. 
Regarding the values, it can be up to the users, however in reality the max is the value of CSF, the min is the skull and the ratio csf/skull ~= [150, 250]. Maybe we can add a test on the max ratio. 
Regarding the precision, maybe we can go up to 1/10000, assuming that unites will be always S/m. 
 


> 2\. If a layer is set to "Anisotropic": are we supposed to still be able to edit the conductivity value? At the beginning I was disabling the conductivity text box, but from your explanations, it looks like we need the conductivity in both cases ("Isotropic" and "Anisotropic"), therefore I put it before the Isotropic/Anisotropic choice.

Based on the Tuch principle we don't need the iso conductivity, since the ansio conductivity will be just the eigenvalue scalled by a known factor, this is method 1 in the big function. 
however, it was shown later that having a realistic conductivity is useful since it will be used to scale the eigenvalue in order to keep the same volume as the isotropic volume (Wolter's constraint). 
Therefore we need the scalar value even for the anisotropic case (methods 2 to 7). 

> 3\. Is it OK to leave all the layers to "isotropic" by default?

Yes, 

> When selecting at least one "Anisotropic" layer, the "Anisotropy method" panel gets displayed:
> image

This is nice, thanks. 

> Please confirm that:
> * 'EMA' = Method 2 = [Güllmar 2010, NeuroImage]
> * 'EMA+VC' = Method 7 = [Rullmann 2008 / Vorwerk 2014], used in Simbio

Yes, in any case, I will review the code, and we can keep the other implementations for the future if we will have requests from users. 

> When selecting "Simulated", more options appear in panel "Simulated anisotropy":
> image

perfect.

> 5. What is the range of values for the "Ratio longitudinal/transversal"? (max number of significant digits, max and min values) - I need to add constraints to this text box as well.

This is mainly integer, from the literature the largest ratio is 10 and it is for the WM. 
But we can go give larger range may be up to 25, it's up to the user. 



>6. What is the default constraint? Wang's?
> 

Let's use the volume as the default.

> > Generally the volume constraint (wolters') is used for WM and GM and the Wang's constraint is used for the skull.
> 
> 1. This last comment **contradicts everything** that was discussed above, everything that I coded in the past days, and all of your code organization... If you need to be able to select all the options independently for each tissue, most of what I did so far is useless, and all the data structures need to be revised again. Let's try to be clear: **do you need more flexibility than what is offered at the moment**? If so, tell me asap, so I start everything from scratch before I move forward...
>    => Please try to draft the ideal interface you would like to have to select all the combinations of options you need.

sorry for this confusion. 
this is mostly done on the simulation part and not the with DTI. 
we are using the volume constraints within method 7. 

Regarding the skull, the future direction is not to use a simulated tensor but by segment it into two sublayers, the skull compacts, and the skull spongiosa. We will go throw this direction later with the new coming process from Marios. 
So, no need to change, and this version is just perfect from my point of view. 


> 8. Why do we need the volume with the tissue segmentation?
>    => It looks like you use it only to get the size of the voxel, but this is something you could also get from the T1 volume.
>    => Isn't the DTI-EIG volume created by BrainSuite always interpolated on the T1 volume?
>    => It seems to me that by construction we don't need the tissue segmentation volume at all because by construction the DTI-EIG is already registered to the T1 volume.

Yes, this was in the previous version, where the dti process and conversion is done in the same script. 
With this new code, we don't need it here. 




> I commited a first draft, only to validate parts of the structure of the code and data with you:

I will test and check and come back to you in my next comment. 

> * I changed the storage of the tensors: now a matrix Tensors [nElements x 12] is stored in the FEM mesh: it will be much easier and faster to manipulate than millions of cell arrays (which get very slow for some operations, like saving data to disk...)

You are right, and I was using this before moving to the cells, since the display of the tensor was based on cells. 
We can change the display code by following this later. 

> * Can you please check that if you right-click on a FEM mesh > Compute FEM tensors, and leave everything to isotropic, get something that makes sense?
> * Same question if you select one tissue isotropic and select "Simulated"?

Ok, I will test all and then come back to you. 

> I'll wait for your feedback on this and the 8 previous questions on the GUI before moving further, and will finish the anisotropic case tomorrow.

Ok, sounds good. 

Thanks for all your efforts and feedbacks.

tmedani:
> This looks amazing. I'm starting to receive emails of people interested on the fem computation. @tmedani @ftadel this is great what you've done.

This is nice to hear that. Looking forward to their feedback and results 

Thanks, @juangpc, we did it ;)

tmedani:
> * Can you please check that if you right-click on a FEM mesh > Compute FEM tensors, and leave everything to isotropic, get something that makes sense?
> * Same question if you select one tissue isotropic and select "Simulated"?

Hey Francois, 

Just reviewed the code, regarding the simulated version with iso and aniso with the two conditions. 

The results look correct, this means that the centroids, the basis vectors, their values and orientations are correct.

image


I have updated the function bst_view_tensor in order to use the new format of the tensors.
btw, I think also this will reduce the time for displaying the tensors for the realistic head model. 

I have also an improved version for displaying that I will push to the folder inProgress. 
We will discuss this part later when we finish the computation. 

A+

ftadel:
And here is the last part of this full recoding work: 
https://github.com/brainstorm-tools/brainstorm3/commit/3c43818975f794d2b7cd51f07c032da852b4d682

All your code is now condensed in one single function, all vectorized. It fits now in about 150 lines, all the rest is my interface wrapper :)
https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/process/functions/process_fem_tensors.m#L151
The only part I haven't found an easy way to remove the for loop is the computation of the tangential vectors. We can keep working on this if this a major bottleneck.

Please review and test very carefully. I'm rewriting blindly, I could not test if the outputs are still the same as before, as I didn't any previous results. Make sure it gives the same results as before.
I would be very surprised if I everything was working exactly as it should.... apologies in advance if it takes some extra debugging work. But on the other hand, now the code is so compact and readable, debugging should be very easy.

I am very suspicious about the 'ema' case: line 302 removes a fixed value that is higher than the DTI eigenvalues, leading to negative values, and complex values in the normalized values (line 306), as the cubic root of a negative value is complex.
https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/process/functions/process_fem_tensors.m#L302

I'll work on the display next week.

tmedani:
> And here is the last part of this full recoding work:
> [3c43818](https://github.com/brainstorm-tools/brainstorm3/commit/3c43818975f794d2b7cd51f07c032da852b4d682)
> 
> All your code is now condensed in one single function, all vectorized. It fits now in about 150 lines, all the rest is my interface wrapper :)
> https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/process/functions/process_fem_tensors.m#L151

Thanks for rearranging all this in a compact and clean code. 



> Please review and test very carefully. I'm rewriting blindly, I could not test if the outputs are still the same as before, as I didn't any previous results. Make sure it gives the same results as before.
> I would be very surprised if I everything was working exactly as it should.... apologies in advance if it takes some extra debugging work. But on the other hand, now the code is so compact and readable, debugging should be very easy.

will do it and give you feedback and even some slight corrections by Tuesday. 

> I am very suspicious about the 'ema' case: line 302 removes a fixed value that is higher than the DTI eigenvalues, leading to negative values, and complex values in the normalized values (line 306), as the cubic root of a negative value is complex.
> https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/process/functions/process_fem_tensors.m#L302

I will check and correct it, there is another paper that I would check where the scaling **k** and the shift **d** are set to other values. 

There is also a unit factor between the DTI coordinate system and the SCS that I need to check.
I will let you know and push the change asap.


> I'll work on the display next week.

There is another version of the tensor display here:
[displayEllipse](https://github.com/brainstorm-tools/bst-duneuro/blob/master/inProgress/displayEllipse.m)
and the function displayEigenLine [](https://github.com/brainstorm-tools/bst-duneuro/blob/master/inProgress/displayEigenLine.m)
which should be faster than the previous version but not adapted yet to bst.

The part that takes time is the assignation of the color for each tensor which needs a new solution.

tmedani:
Hi, 

I want just to report a weird behavior that freezes Matlab and my computer. 
I have tried to save both tensor results, those from my code, and those from your new implementation. 
The final size of the file was ~4 Gb, and it was not possible to save it to the database with the option v7 ...
BST> Warning: Uncompressed file size is 2.128 Gb, saving using v7.3 format...

The weird thing is when I have the tensors in two separate files the size of one file is ~400Mb, but when I both results the size ~ 4Gb ... 

So regarding the code, from my reading and my basic testings, it seems that your implementation is correct and I have almost similar results in both methods, [emc and emc+vc].
The position, the direction, and the values of the tensors seem to be stables and corrects. 
(I have checked only the mean value and the std). 

image


Regarding the method EMA, you are right, a factor is required in order to avoid the negative values. I need to check the BDP outputs units (@ajoshiusc).

I think also in this method we will remove the normalization and keep only EFM, we will keep only the normalization on the second method.

I'm discussing with Carsten's team in order to confirm this asap.
 
Regarding the combined GM and WM tensors, your implementation is much better than mine and it's easy to incorporate both tissues. 
I have just to check the results and compare them to the published formulation in Rullmann paper and with what I have implemented in my side (in my computer). 

Regarding the part of the tangential vectors, which takes more time, I will check in my previous code, maybe I will have another version without the 'for' loop. 

Thank again you for this super clean and compacted version.

ftadel:
> I have updated the function bst_view_tensor in order to use the new format of the tensors.
btw, I think also this will reduce the time for displaying the tensors for the realistic head model.

FYI: I started working on integrating the display of the tensors in Brainstorm 3D figures. 
I will try to find nice ways to make it interactive (eg. scrolling between slices with the Resect sliders).
And I am pretty sure I can make it work a lot faster.

I will not produce anything usable before the end of the week. Maybe early next week.
In the meantime, do not spend time trying to improve your display functions.

tmedani:
> > I have updated the function bst_view_tensor in order to use the new format of the tensors.
> > btw, I think also this will reduce the time for displaying the tensors for the realistic head model.
> 
> FYI: I started working on integrating the display of the tensors in Brainstorm 3D figures.
> I will try to find nice ways to make it interactive (eg. scrolling between slices with the Resect sliders).
> And I am pretty sure I can make it work a lot faster.
> 
> I will not produce anything usable before the end of the week. Maybe early next week.
> In the meantime, do not spend time trying to improve your display functions.

I'm sure that you will :) 

I will wait for your last updates and test altogether. 

Now, I'm checking the FEM computation with the Cauchy files and I found some issues with the current version. 

I will correct it and push the changes asap. 

Thanks.

ftadel:
Tensors display commit:
https://github.com/brainstorm-tools/brainstorm3/commit/41e087faec7feb566fa38b3af13566d4ff714851

Discussed here:
https://github.com/brainstorm-tools/brainstorm3/issues/267#issuecomment-641279058

Please let me know if I need to do something else at the moment for the computation and display of the tensors.

tmedani:
wonderful outcomes from this discussion.

I'm going to close it now, most of the required functions are implemented. 

Thank you @ftadel

PR #290, opened 2020-05-05
use gray colormap for mri as default

juangpc:
Hi, 
This PR proposes to change the default colormap for mri viewer from 'bone' to 'gray'. 
It is clear that this might just be my opinion, so feel free to reject no worries. I'm opening the pr just to propose the change. A couple of weeks ago, while working with MDs, it was mentioned that the colors of brainstorm's mri viewer seemed "blueish". When comparing with other software, typically the colors are more black-and-white. 

This is an edited capture of an image we came up with where the difference is clear. A is a capture from bst, B is done with another sw. 

I think MDs like more the black-and-white map. Maybe not.

In any case thanks for reviewing the pr. Take care,

image

### DIFF
--- a/toolbox/core/bst_colormaps.m
+++ b/toolbox/core/bst_colormaps.m
@@ -119,7 +119,7 @@ function sColormap = GetDefaults(ColormapType)
             sColormap.MaxMode          = 'global';
         % Anatomy colormap
         case 'anatomy'
-            sColormap.Name             = 'bone';
+            sColormap.Name             = 'gray';
             sColormap.CMap             = bone(DEFAULT_CMAP_SIZE);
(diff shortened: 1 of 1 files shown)


ftadel:
Please ask for the group's confirmation during today's meeting, I'll merge your suggestion when everybody has approved.

ftadel:
Didn't realize that we were not having a proper meeting today...
I'll merge it and ask for feedback by email, this is easy to revert if needed :)

ftadel:
Completed here: https://github.com/brainstorm-tools/brainstorm3/commit/f65ca802537ce4a85793966d7f4b5745401f9e0c

PR #291, opened 2020-05-07
PTE: Added alternative for non-toolbox users

martcous:
As reported here: https://neuroimage.usc.edu/forums/t/can-i-run-pte-without-statistics-toolbox/18899
The PTE process required the Statistics toolbox and the Signal Processing toolbox.
While our implementation comes from an external source, I have edited it to use alternative functions for non-toolbox and older Matlab versions. Since the PTE process is very slow, I make sure to use the fastest function whenever possible. This might not be the best practice possible so I'd like a quick review before merging. Thanks!

### DIFF
--- a/external/fraschini/PhaseTE_MF.m
+++ b/external/fraschini/PhaseTE_MF.m
@@ -56,10 +56,15 @@ N = size(data,2); %number of signals
 PTE = zeros(N,N);
 dPTE = zeros(N,N);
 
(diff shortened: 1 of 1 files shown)


ftadel:
I haven't tested it, but it looks good :)

ftadel:
Since this is external code, you should suggest this change to the authors?

PR #292, opened 2020-05-13
Fix bad head coil (CTF)

Moo-Marc:
Main objective: new process to "fix" (best guess) a bad or fallen head coil.  For CTF there are usually only 3 coils, so the best we can do is complete the coil triangle using information from a previous recording, assuming the head will have about the same tilt.  (We could later extend this to cases that have more than 3 coils, in which case it would be a real correction, not a guess.)

Secondary changes:

- Adjust head position: new option to exclude bad segments.
- in_fread & ProcessFilter: more specific testing before giving errors for missing channels (CTF compensation and SSP)
- ProcessFilter: a few additional tweaks.  Most changes are related to initialization and row blocks.  In the end the change is minor and it was probably not worth the effort, but my initial thought was to make sure processed row blocks are "full" (minimize the number of calls to the external process) and if there are only a few rows to process, increase the column block size.  Again, probably just a minor efficiency increase in specific cases.

Probably needs a bit more testing since ProcessFilter is a core component.  I could use help with that.  Are there testing scripts somewhere? E.g. A script that automatically runs through the entire main tutorial series so we can verify the results are unchanged?


ftadel:
Ok for most of the PR, but I'm skeptical about all these changes in bst_process.
Everything before line 548 looks reasonable and is easy to review. But what happens after looks too complicated if the goal is to save only one call to your process function... This would be complicated to review and complicated to test, with high risks of damaging a sensitive function. 

Could you remove it from this PR?
And if you really think this represents an interesting improvement, open another dedicated PR for it.

Thanks

martcous:
I agree with François, we need to be careful and do lots of testing if we push changes to such a critical function (bst_process). The process itself seems to work fine on my dummy tests.

I think it would be important in the process options to clarify the direction of the inputs, a.k.a. which group of files will be considered bad, and which will be considered good (A vs B). It is clear on the tutorial page, but we could add a label directly in Brainstorm to indicate it as well.

As for its placement in the process2 menu (Import -> Channel file -> Fix bad head coil), I'm not so sure a subgroup is required since there is no other process in either groups. It does make it consistent with the process1 menu though, so up to you!

Moo-Marc:
I understand your caution.  So I take it there are no testing scripts.  This would be very useful to have.  

If I were to manually go through the tutorials with no error and reproduce the expected results, would you be convinced that nothing is broken?  I already ran a few different processes successfully, including resampling, filtering, importing, ..., so I'm growing confident it's ok.  

The only complicated part is for raw files when there is a selection of channels and no time change.  In that case we need to read the missing channels.  In all other cases (non raw, time change, or all channels), my changes are minor and simplify the code: initializing with the previous data instead of initializing with zero and then gradually inserting the unprocessed data which is already in a variable.  So in terms of testing, it's not really complicated...

So while we could just not do it, since I went through the trouble and it's not really that hard to test I would prefer going ahead.

Martin, do you mean put a comment in the _process options window_ to say A = bad, B = good?
As to the menu, I don't really mind either way.  The extra layer is not necessary, but also very fast to glide through with the mouse.

ftadel:
>  So I take it there are no testing scripts. This would be very useful to have.

For testing before/after, you can use all the tutorial scripts.
```
brainstorm tuto_name tuto_dir;  
% tuto_name: calls the function ['tutorial_' tuto_name]
% tuto_dir: parent folder for the dataset downloaded from the website
```

Or this simple test script that simulates data and does not need any dataset in input (function test_all.m):
```
brainstorm test empty_folder
```

This would definitely benefit for being improved. Spending time to write a better coverage test could possibly be more beneficial than rewriting code that already works :)

> So while we could just not do it, since I went through the trouble and it's not really that hard to test I would prefer going ahead.

Can you please split this PR in two (or maybe three), so we can first validate what this PR is about (fixing bad head coils for CTF data) and then have a better view on what you are trying to do with the general processing? 
It would also make it easier to debug or revert in case of problems.

Moo-Marc:
> Can you please split this PR

Will do.  Thanks

Moo-Marc:
Done.  I'll make a separate PR for the stuff related to row blocks.

ftadel:
Thanks!
I haven't tested the modifications you made to bst_process, but they look reasonable and if they break something ,it will be obvious and we'll get complains very soon of the forum :)

PR #300, opened 2020-05-31
Nirs : fix sensor display and MRI registration

Edouard2laire:
Hi, 

This PR fixes an old bug that occurs when importing nirs. When importing nirs data, at the end of the registration, instead of showing the montage, Brainstorm was showing this figure : 

Capture d’écran 2020-05-31 à 20 29 15
Each white dot is the middle of a channel, so not very informative to see if the registration was ok. 
This was due to the fact that sometimes, the modality is defined as 'NIRS' and sometimes as 'NIRS-BRS'.   One thing that could be great, I don't know if you can help me with that, would be to allow like with EEG to project probe on the scalp. 

This commit also fixes the display when we use 'Display Sensors > NIRS (scalp)' by adding the name of the probe in the figures. 

Sensors display (scalp & pairs) : 
Capture d’écran 2020-05-31 à 22 24 24

MRI registration > check 
Capture d’écran 2020-05-31 à 22 24 43

MRI registration > edit 
This only fixes the display, but I am almost certain we need to make changes to handle the difference between EEG and nirs as we want to make change on the probe and not channels ( pair of two probes).  

Capture d’écran 2020-05-31 à 22 25 01

Best regards, 
Edouard

### DIFF
--- a/toolbox/gui/figure_3d.m
+++ b/toolbox/gui/figure_3d.m
@@ -4207,7 +4207,7 @@ function hPairs = PlotNirsCap(hFig, isDetails)
         locPairDet = cat(1, locPairDet{:});
         
         % Make the position of the links more superficial, so they can be outside of the head and selected with the mouse
-        if any(~locPairDet(:,3)== locPairDet(1,3) ) || any(~locPairSrc(:,3)==locPairSrc(1,3)) 
+        if length(unique(locPairDet(:,3))) > 1 || length(unique(locPairSrc(:,3))) > 1
             normSrc = sqrt(sum(locPairSrc .^ 2, 2));
             normDet = sqrt(sum(locPairDet .^ 2, 2));
             locPairSrc = bst_bsxfun(@times, locPairSrc, (normSrc + 0.0035) ./ normSrc);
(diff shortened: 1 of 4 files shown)


ftadel:
> One thing that could be great, I don't know if you can help me with that, would be to allow like with EEG to project probe on the scalp.

I'm not sure how this would be different?
If you project all the locations on the scalp, you end-up projecting the "probes", isn't it?

Edouard2laire:
> > One thing that could be great, I don't know if you can help me with that, would be to allow like with EEG to project probe on the scalp.
> 
> I'm not sure how this would be different?
> If you project all the locations on the scalp, you end-up projecting the "probes", isn't it?

The option is not displayed by default for nirs.  We can make it available by changing 
https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/sensors/channel_align_manual.m#L350 to 'if gChanAlign.isMeg'  but then we have the following error when trying to project the probe : 

```
Subscript indices must either be real positive integers or logicals.

Error in channel_align_manual>ProjectElectrodesOnSurface (line 1083)
        set(gChanAlign.hSensorsLabels(iTextChan), 'Position', 1.08 *
        gChanAlign.SensorsVertices(iChanToProject(i),:));
 
Error while evaluating PushTool ClickedCallback.

```

it also change the display to: 
Capture d’écran 2020-06-01 à 17 29 00

A suggestion explaining why it might not work is that when clicking 'edit selected channel name label', Brainstorm display something like 'S1D1WL830', when it should be S1, or D1 and if a change is made, all channel containing S1 will be renamed to SXD...

Edouard2laire:
There is also a problem with the button that show or hide the optodes labels to correct it, we have to change the function 'ToggleLabels'  to because apparently nirs plot doesn't store information at the same place as eeg plot  (hSensorsMarkers vs hSensorsLabels)

```
%% ===== SHOW/HIDE LABELS =====
function ToggleLabels(varargin)
    global gChanAlign;
    % Update button color
    gui_update_toggle(gChanAlign.hButtonLabels);

    if strcmpi(get(gChanAlign.hButtonLabels, 'State'), 'on')
        if gChanAlign.isNirs
           set(gChanAlign.hSensorsMarkers, 'Visible', 'on');
        else    
            set(gChanAlign.hSensorsLabels, 'Visible', 'on');
        end    
    else
       if gChanAlign.isNirs
           set(gChanAlign.hSensorsMarkers, 'Visible', 'off');
        else    
            set(gChanAlign.hSensorsLabels, 'Visible', 'off');
        end    
    end
end

```

ftadel:
I think I addressed all these issues in this commit:
https://github.com/brainstorm-tools/brainstorm3/commit/0611d417f19515b9819b850d9ac63aed03a5d72b

Let me know how it goes

Edouard2laire:
Thanks. it's working great !

Issue #293, opened 2020-05-14
FEM: Remove bottom part of head model


ftadel:
Hi, 

Just a low priority task to add to the todo list. 

In order to reduce the FEM computation time, it's highly recommended to remove the bottom part of the head model. 

The recommendation (Lanfer et al), is to cut the head at  40mm below the lowest part of the skull. 

Since cutting the mesh if not recommended (or not possible), we need to cut the MRI or the MASK before calling any mesh method. 

We can set it as the default option or add it as an option before calling the mesh generator.

_Originally posted by @tmedani in https://github.com/brainstorm-tools/brainstorm3/issues/185#issuecomment-626079804_

ftadel:
Do you mean: before calling SimNIBS only?

We could add something that cuts the volume before saving the T1.nii in the temporary folder:
https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/process/functions/process_generate_fem.m#L585

It could be something similar to the defacing using the SPM technique, cutting below a given plane in MNI space:
https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/process/functions/process_mri_deface.m#L202

ftadel:
PS: we need to be careful about not cutting above the lowest of the electrodes.
With EGI 256 electrode hydrocel nets, the lowest electrodes are really low in the neck.

ajoshiusc:
> Hi,
> 
> Just a low priority task to add to the todo list.
> 
> In order to reduce the FEM computation time, it's highly recommended to remove the bottom part of the head model.
> 
> The recommendation (Lanfer et al), is to cut the head at 40mm below the lowest part of the skull.

@tmedani 
We can cut the mesh, that way we are not affecting other processing. To cut the mesh, you can first find the centroids of tetrahedrons and then delete the tetrahedrons that have centroid below the specified plane. After that, we need to delete the vertices that have not been used and reindex. I have code for that (deleting unused vertices). Would that make sense?

tmedani:
Hello,
Thanks  @ftade  for opening this issue.

> Do you mean: before calling SimNIBS only?

Not only simnibs but most of the methods.

> PS: we need to be careful about not cutting above the lowest of the electrodes.
> With EGI 256 electrode hydrocel nets, the lowest electrodes are really low in the neck.

Indeed, we can add it as an option that asks the user to keep the whole volume or just the top part. 

[...]

PR #301, opened 2020-06-01
Cut FEM mesh / remove bottom part

tmedani:
Hello, 

This function allows us to cut and remove the bottom part of the mesh.
This could be useful in the case where the positions of the sensors are only the top part of the head,
in this case, the bottom part is not needed and can save computation time.

with this version, the user can call it by right click on the FEM mesh and set a ratio to remove. 
The default value is 1/3, but the user can change it. 

Brainstorm will create a new head model and save it to the database.

this issue is discusses [here ](https://github.com/brainstorm-tools/brainstorm3/issues/293)

image
 
@ftadel this is the first version and I think it may be better to include it to the main [function process.m](https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/process/functions/process_fem_mesh.m) ?

Let me know if I have to change the code and add it to process_fem_mesh.m

Thanks


ftadel:
There are two problems with the way this function is designed:
- It requires the user to do a few trial and errors to get a good value that doesn't cut too much
- Cutting at z=z0 works in some cases, but won't work if the T1 is not realigned in a FreeSurfer-like orientation. In some cases the head is tilted, or the angle of the MRI slices is not strictly an axial plane...

The defacing code I initially pointed at cuts below a plane with a equation in MNI coordinates. This solves both issues. You just need to define the equation of your plane once, and it will be applicable to ALL the subjects without any individual tweaking, and it's independent on the orientation of the MRI scan. 
https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/process/functions/process_mri_deface.m#L202

I strongly encourage you to consider this kind of solution.

ftadel:
Another concern:
Wasn't the goal to minimize the computation time with SimNIBS? 
Shouldn't we cut the volume before running the segmentation, rather than after?

tmedani:
> * It requires the user to do a few trial and errors to get a good value that doesn't cut too much

agree

> * Cutting at z=z0 works in some cases, but won't work if the T1 is not realigned in a FreeSurfer-like orientation. In some cases the head is tilted, or the angle of the MRI slices is not strictly an axial plane...

Ok, thanks for letting me know this, I thought that all the bst models are oriented in the same way. 


> The defacing code I initially pointed at cuts below a plane with a equation in MNI coordinates. This solves both issues. You just need to define the equation of your plane once, and it will be applicable to ALL the subjects without any individual tweaking, and it's independent on the orientation of the MRI scan.
> https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/process/functions/process_mri_deface.m#L202
> 
> I strongly encourage you to consider this kind of solution.

will check it, this is the clean method. 




> Another concern:
> Wasn't the goal to minimize the computation time with SimNIBS?
> Shouldn't we cut the volume before running the segmentation, rather than after?

initially, it was the idea, and then the idea to cut the mesh is proposed in this [discussion](https://github.com/brainstorm-tools/brainstorm3/issues/293)

since we can do both, I think we can integrate both methods. It will give more flexibility for the users.
do you agree?

tmedani:
Hi, 
again this is great, 

I think with Z = -85, 
there is only 10 mm from the lower part of the skull to the cutting plan


image
image


So as you said this is slightly high. 


I have checked with  Z = -115, and  there is ~ 40mm 

image
image

Let's use this as the default value since it fits what is said in this paper 
https://onlinelibrary.wiley.com/doi/full/10.1002/hbm.24754


image


And their figures fit better the cut at **Z = 115**

image


We should explain this on the tutorial since this is also related to the position of the sensor. 


btw, a small bug, the 'Get Coordinates' panel is not working on the FEM mesh display. 
image

Thanks & A+

Issue #299, opened 2020-05-31
Errors after latest update 


gsdouk:
Hello,

I have combined EEG/MEG data and there is a problem with the process Apply SSP & CTF compensation. Until the day before 30/May/2020, there was absolutely no problem applying this process. Now I am getting this error:    
![Inked83353102-dee1a200-a358-11ea-8e0a-c37ef7e48caa_LI](https://user-images.githubusercontent.com/59571840/83355199-8fa26e00-a366-11ea-9d16-d24762fdd7f4.jpg)

I have run the same process many times before on the exact same data files without a problem (I was using the previous version of 29/May/2020).  On the latest version, (29/May/2020) I have the aforementioned error.

Secondly, when using the latest version of Brainstorm (29/May/2020) ICA yields different results from when using the previous version. Exact same input and different results before and after the 29th May. The first image is taken the first week of May and the second one is taken 31/May and is wrong (exact same data, exact same processing steps before ICA was applied). Is the update related with any of these processes? How can I get the same behavior as before?
![image](https://user-images.githubusercontent.com/59571840/83354730-371da180-a363-11ea-87b9-d979f817b6bc.png)
     
![Capture1](https://user-images.githubusercontent.com/59571840/83354714-1d7c5a00-a363-11ea-8460-55621b592019.PNG)

Finally, the [Download](https://neuroimage.usc.edu/bst/download.php) page is not available since 30/May (or that day i noticed it). So I cannot download the tutorial data and run the tutorials again.  

Also, I have reset/reinstall the latest version (I downloaded the code from github) of Brainstorm several times.

Thank you in advance.

ftadel:
> I have run the same process many times before on the exact same data files without a problem (I was using the previous version of 29/May/2020). On the latest version, (29/May/2020) I have the aforementioned error.

There were changes in bst_process.m in the past month, so it is plausible that some new bugs appeared. But I could not reproduce your bug you reported.
If you want us to help you debug this, could you please share a short example data file ?

- Right-click on the subject > File > Duplicate subject
- Delete all the files that are not needed to reproduce the behavior:
  - in the anatomy, keep only the MRI and the head surface
  - in the functional data, keep only the channel file and one short block of imported data (1s long or less) - delete forward and inverse models
- Make sure you can still reproduce the error with this minimal dataset
- Right-click on the subject > File > Export subject
- Upload the zip file somewhere (dropbox, google drive...) and share the download link here
[...]

PR #304, opened 2020-06-05
fix to bst_process incorrect resampling error message

Moo-Marc:
as per forum report https://neuroimage.usc.edu/forums/t/19542 and https://github.com/brainstorm-tools/brainstorm3/issues/299 

Optional fix since the revert Francois did also fixed the issue.  

This simply adds an error message if a process attempts to split by columns and change number of samples.  I also moved a few lines that only need to run once.

### DIFF
--- a/toolbox/process/bst_process.m
+++ b/toolbox/process/bst_process.m
@@ -806,19 +806,25 @@ function OutputFile = ProcessFilter(sProcess, sInput)
             end
 
             % === INITIALIZE OUTPUT ===
-            % Split along columns (time): No support for change in sample numbers (resample)
             if ismember(2, sProcess.processDim)
-                nOutTime = nCol;
                 iOutTime = iCol;
-            % All the other options (split by row, no split): support for resampling
-            else
-                nOutTime = length(sInput.TimeVector);
-                iOutTime = iCol(1) - 1 + (1:length(sInput.TimeVector));
             end
-
+            
             % Create output variable
             if isFirstLoop
                 isFirstLoop = 0;
+                % Split along columns (time): No support for change in sample numbers (resample)
+                if ismember(2, sProcess.processDim)
+                    if length(sInput.TimeVector) ~= length(iCol)
+                        bst_report('Error', sProcess, sInput, 'Split along columns (time): No support for change in sample numbers (resample)');
+                        return;
+                    end
+                    nOutTime = nCol;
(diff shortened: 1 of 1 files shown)


ftadel:
As I explained before, optional fixes to sensitive code are not particularly welcome. 

This doesn't seem very harmful, but at the same time it seems quite useless in the sense it does not bring any improvement to the way the function bst_process.m works. It captures an error that should not occur in the first place and will not bring any speed improvement. On the other hand, it alters a function that has been stable for years in a way that could lead to weird unwanted side effects (a statement that was executed multiple times is not anymore), which can cause sneaky bugs that would take hours to find. 

The previous attempt you mention here took me a while to handle, for zero improvement. 
As we are already have an infinite todo list of useful things to do (https://neuroimage.usc.edu/brainstorm/Next), could we agree on avoiding spending time on useless-but-hazardous modifications in the future?  

If you are 100% sure this is safe and you want to move on with this modification, can you please handle the merge with @martcous?

Thanks

martcous:
I agree with François here, it's better to fix the processes themselves rather than write an endless amount of potential error catches in bst_process. A bug in a single process is much less limiting than in a function called by every process. The issue seems to have been fixed by François so I would not move forward with this change.

Moo-Marc:
Not an unexpected reaction, but you did ask me, correctly, to triple check.  So I found the error (which was due to how I split my previous pull request) and fixed it. Now up to you which fix you prefer.

ftadel:
I'd prefer avoiding additional modifications to this function.
Are you confident that this partial revert I made as an emergency fix does fix all the problems that were introduced in bst_process?

Moo-Marc:
I'm aware of only this one problem.  I quickly rechecked issues and the forum.  Did I miss something?

ftadel:
I meant: 
Are you confident that this partial revert I made as an emergency fix does entirely fix this problem that was introduced in bst_process?

If you think this OK the way it is now on the repo, I would close this PR.

Moo-Marc:
Yes, this problem is fixed by the revert.

I disagree that this error message is useless though.  This is an unchecked and undocumented (without digging into the code) assumption - that resampling is not allowed when splitting by column blocks.

ftadel:
OK, it's not useless from the coding point of view.
But in practice very few people write processes, while thousands of people use the software, and we have tons of things very useful from the user point of view to do first :)

Issue #308, opened 2020-06-14
Internet Connection


gsdouk:
Hello,

I encounter a problem connecting to server via BST, cannot download updates etc. When BST initializes, the internet connectivity is fine (Checking internet connectivity... ok). I did not had this problem before the 02/June update. 

Thank you in advance.

ftadel:
Can you please describe better you issue(s)?
What is happening? When? 
Please copy-paste the full error messages you get, including the stack traces.

gsdouk:
On BST's initialization I get: 
![image](https://user-images.githubusercontent.com/59571840/84684126-e3a17b00-af40-11ea-83c9-c40dfe311e52.png)


The error I get when I am trying to update BST:
![image](https://user-images.githubusercontent.com/59571840/84684204-0469d080-af41-11ea-860e-a7faad7dfd08.png)


One more example (similar with the above), when I am trying to compute MNI transformation on the MriViewer when importing anatomy: 
![image](https://user-images.githubusercontent.com/59571840/84684568-7e01be80-af41-11ea-9bc1-bf78b1f592bf.png)


The stack traces:
![InkedCapture_LI](https://user-images.githubusercontent.com/59571840/84685348-cff71400-af42-11ea-9378-db9741898eab.jpg)



The 15/June update did not fix this error. Before the 02/June update I did not have any problem updating BST.

ftadel:
Do you have an older version of Brainstorm available somewhere?
Could you run it and try to run the update from it?

We're having certificate issues on the webserver at the moment, this might be related with this problem. If this is the cause, any older version of Brainstorm should not work any better.

What do you get if you execute the line below in your Matlab command window?
`webread('http://neuroimage.usc.edu/bst/getversion_duneuro.php')`

While we're working on this issue, please update Brainstorm manually (download a new version from the website) and/or use the compiled version.

gsdouk:
>Do you have an older version of Brainstorm available somewhere?
Could you run it and try to run the update from it?


Yes, I run an older version (23/Sep/19) and I still get the same error message. 


>What do you get if you execute the line below in your Matlab command window?
webread('http://neuroimage.usc.edu/bst/getversion_duneuro.php')


![image](https://user-images.githubusercontent.com/59571840/84689941-6a0e8a80-af4a-11ea-86d2-1c38acac7722.png)


>While we're working on this issue, please update Brainstorm manually (download a new version from the website) and/or use the compiled version.


[...]

Issue #306, opened 2020-06-10
Anisotropy in the Duneuro panel


tmedani:
Hi all, 

@ftadel this is a quick issue that we can solve but we need to discuss it before. 

In the case of the FEM Duneuro computation with the anisotropy tensors, we need to add an input to the duneuro panel. 

In this case, the iso conductivity values are not needed since the tensors are already in the femmat structure. Furthermore, in this case, the cauchy files will be used (knw and geo). 
So we need something that tells the user if he wants to use the tensor or not?
and use this as a boolean in the code in order to use the catchy files. 

also in this case the scalar values of the conductivity are not required. 

As previously discussed, I will not code and push something wrong or not be useful, so let's discuss  

Here is the current panel  
![image](https://user-images.githubusercontent.com/37831900/84215123-316a3f00-aa7a-11ea-9dac-801245d1f121.png)

I added some code and I change the panel to this

![image](https://user-images.githubusercontent.com/37831900/84215179-578fdf00-aa7a-11ea-9948-aa7c6316c708.png)
 

Where the case 'UseTensor' will load the tensor from the head model (femmat) and uses the Cauchy files (*.knw and *.geo)

when we select the UseTensor, the values of the conductivity should be disabled.

Does it work for you or do you have a better option?

The case 'isotropic' is not useful for the final user, it's just in my local version for more testing. 
FYI, I'm using this to create N times (length(Element)) the isotropic tensors to fille the knw file and then call the duneuro.  

 

Also, I have updated the tutorial with some results that validate the anisotropy FEM computation, check  [here](https://neuroimage.usc.edu/brainstorm/Tutorials/ReviewForward?highlight=%28duneuro%29#Ansiotropy_model)

A+

ftadel:
> when we select the UseTensor, the values of the conductivity should be disabled.

I would go for skipping completely this extra parameter "Use tensors":
- if the tensors are available in the FEM file, we use them and hide the panel "FEM layers and conductivities" (or is just the edit boxes you want to disable, and keep the tissue selection checkboxes?)
- if the tensors are not available in the FEM file, then we keep the panel as it is now.

In most cases, people who have the DTI images would use them, and the ones without would not. So it is not really an option that most users would need to see in this panel. And the fewer options in this panel the better.

[...]

PR #309, opened 2020-06-16
add the tensors to the fem computation with duneuro

tmedani:
Hello 

This PR takes into account the last updates regarding the tensors and FEM anisotropy computation.

This version checks if the Tensors field is present and not empty on the femmat (head model).
if it's the case it will set the UseTesor to true and then the associated temporary files will be used.

@ftadel I have noticed some bugs
- The duneuro_install- is not working, there is this error: "could not get the current online version of bst_Duneuro"
this error is related to the server and I have the same issue as reported here: https://github.com/brainstorm-tools/brainstorm3/issues/308

==>  I have added a **temporary** option to make it work since I have it already on my computer. 
Maybe we can add a similar check before re-installing it and avoid the server errors. 

- The current version of 'out_fem' that writes the geo file is not working, therefore I'm using the previous function "bst_write_cauchy_geometry(cfg) "

- We need to change also the duneuro panel as discussed here https://github.com/brainstorm-tools/brainstorm3/issues/306

A+

### DIFF
--- a/toolbox/forward/bst_duneuro.m
+++ b/toolbox/forward/bst_duneuro.m
@@ -31,6 +31,10 @@ errMsg = '';
 gui_brainstorm('EmptyTempFolder');
 % Install bst_duneuro if needed
 [DuneuroExe, errMsg] = duneuro_install();
+if isempty(DuneuroExe)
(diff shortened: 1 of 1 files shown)


ftadel:
All the issues related here should be fixed in this commit:
https://github.com/brainstorm-tools/brainstorm3/commit/070bbb8a7d4011b8d95bc8302a31ba4861dceca6

PR #313, opened 2020-07-08
PSD: Consider bad segments on raw scout sources

martcous:
The PSD process currently considers bad segments on raw recordings, raw sources, but not when using scouts of raw sources. 
If this was intended or not implemented in the best way please let me know!

To reproduce issue:
1. Add a bad segment to a raw data file
2. Compute sources
3. Run PSD on the sources. If you check "Use scouts", no bad segment are skipped. If you do not, bad segments are skipped (for quick debugging, the user is notified in the command window when bad segments are skipped).

### DIFF
--- a/toolbox/timefreq/bst_timefreq.m
+++ b/toolbox/timefreq/bst_timefreq.m
@@ -444,6 +444,23 @@ for iData = 1:length(Data)
             else
                 nComponents = OPTIONS.nComponents(iData);
             end
+            
+            % PSD: we don't want the bad segments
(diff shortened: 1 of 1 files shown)


martcous:
It seems the process_extract_scout has the same issue for source files on raw data files, which is used in other derivative processes like tPAC.

ftadel:
> 
> 
> It seems the process_extract_scout has the same issue for source files on raw data files, which is used in other derivative processes like tPAC.

Then the tPAC process should be fixed as well?

Issue #302, opened 2020-06-02
FEM source space construction 


tmedani:
Hi, 

In this issue, we will discuss and solve the FEM source space generation for the realistic head model. 

This will be important in the case where we separate the brain into WM and GM. 

For more accurate results, all the dipoles (sources) should be within the GM layer, and ideally at least one element far from the neighboring tissues.  

The cortex generated with CAT is good, however, with the SimNibs FEM mesh, some points can be either on the CSF or on the WM.

We need to solve this for more realistic head models. 

A+

tmedani:
Hello, 

A first solution for this issue is proposed in this PR 
https://github.com/brainstorm-tools/brainstorm3/pull/316#pullrequestreview-455193388

This is faster than the standard approach which checks all the dipoles and all the GM elements and takes between 1-3 hours to build the source space. 
In this pull request, we use the SPM-CAT (0rrany other cortex) cortex and then check the position of the dipoles, then all the dipoles located outside the GM are moved slightly to the nearest centroid of the GM element and it takes only few minutes.

A+

ftadel:
Modified solution proposed in this commit: https://github.com/brainstorm-tools/brainstorm3/commit/29176ee27f3f1f71237e66dbb28eac2f64f0ee86

The new option is available in the DUNEuro option panel (adaptive to the layers available):
![image](https://user-images.githubusercontent.com/6920058/89223185-a8353b80-d5d6-11ea-80a2-6ff30b0739b0.png)
![image](https://user-images.githubusercontent.com/6920058/89223227-b8e5b180-d5d6-11ea-8f62-4ee845c9f3e5.png)

This solution keeps your logic for identifying the closest GM element for dipoles that are outside the GM.
But I think your solution of replacing the position of the dipole directly with the centroid of the element doesn't work, as in my tests it seemed to project a lot of source points into exactly the same location... I'm proposing here a first draft for something more adaptive, that moves the dipoles progressively towards the center of the centroid until it is inside the element, but not necessarily reaching its center. In the Matlab command window, you can see by how many millimeters each dipole was moved.
What do you think?

To evaluate the impact on the shape of the cortex, you can uncomment the two lines with calls to view_surface_matrix. This lets you use the Surface tab in the Brainstorm window to manipulate the figure (transparency, smoothing, edges...) => much more convenient than most surface viewers in Matlab.
[...]

PR #316, opened 2020-07-23
remove the unused vertices, element and tensors from Cauchy files for MEG

tmedani:
Hi, 

This is related to the previous discussion with the introduction of the anisotropy computation, 

This PR contains part of code that removes the unused vertices, elements, and tensors from the femmat head model. 

This part of the code is exactly the same as the one that we have used to resect the neck, https://github.com/brainstorm-tools/brainstorm3/blob/3753b958571b6483e53f803e71d3049124cfbf45/toolbox/anatomy/fem_resect.m#L89

So, @ftadel it's up to you, we can merge this PR as it is or creates a sub-function that can be called by both functions.  

Best,

### DIFF
--- a/toolbox/forward/bst_duneuro.m
+++ b/toolbox/forward/bst_duneuro.m
@@ -116,8 +116,23 @@ if strcmp(dnModality, 'meg')
     % Remove the elements corresponding to the unselected tissues
     iRemove = find(~ismember(FemMat.Tissue, find(cfg.FemSelect)));
     if ~isempty(iRemove)
-        FemMat.Elements(iRemove,:) = [];
-        FemMat.Tissue(iRemove,:) = [];
+        % Remove elements
+        FemMat.Elements(iRemove, :) = [];
+        FemMat.Tissue(iRemove) = [];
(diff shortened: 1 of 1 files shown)


tmedani:
We have discussed this here 
https://github.com/brainstorm-tools/brainstorm3/issues/306#issuecomment-646237134

ftadel:
I will review this later this summer.

close #306

ftadel:
I split your modifications proposed in this PR in two commits:
- One related with the removal of the elements from the Cauchy file (the title of this PR): https://github.com/brainstorm-tools/brainstorm3/commit/090f7796d6e6fef47eca91bccb16f02acbf85fcc
- One related with the new option "Force source space inside GM": https://github.com/brainstorm-tools/brainstorm3/commit/29176ee27f3f1f71237e66dbb28eac2f64f0ee86

Closing this PR now, and keep the discussion going for the second one in issue https://github.com/brainstorm-tools/brainstorm3/issues/302

tmedani:
hi, 
yes, you are right, it was indeed two PR but it appears like just in one.

PR #318, opened 2020-07-30
Revert recent change in mri_histogram that failed with some defaced MRIs

Moo-Marc:
As described [here](https://neuroimage.usc.edu/forums/t/brainstorm-generate-head-surface-fails/20166/13), removed 50% limit for detecting "already removed background" as this failed in tutorials with defaced MRIs.

I also changed the label for the new option as it wasn't clear that the displayed value was the default automatically calculated one.  Fixed to avoid error when this new field is left empty, though it can still give an error on a value to large.

Added "Recompute" option in view histogram function, so that one can easily fix this issue without having to re-import files:
view_mri_histogram(MriFile, true)

image
(Image for updating [the tutorial](https://neuroimage.usc.edu/brainstorm/Tutorials/LabelFreeSurfer) once merged.)


ftadel:
Thanks for investigating this issue.
I fixed it in a slightly different ways than what you propose: since computing the histogram is fast, I changed the viewing function to always recompute the histogram. 
See commit: https://github.com/brainstorm-tools/brainstorm3/commit/25732db0073a343411ea2c88bebb5686873e41f9

PR #319, opened 2020-08-10
stop openmeeg LF computation if dipoles are outside of the innerskull

tmedani:
Hi, 

just a small correction for the bst_openmeeg.m function, 
the message "Error: Some dipoles are outside the BEM layers" is displayed at the end of the computation, 
and sometimes this computation can be long. 
so it's can be much better to display this message before the computation, then the users can correct the cortex by the ** Right-click on the surface file > Force inside the skull ... as proposed by the original version. 

A+


ftadel:
I added a user confirmation when in interactive mode (= not from a process):
https://github.com/brainstorm-tools/brainstorm3/commit/728a5919b1fff1d57c2305b2a7f4ac755d9ca697

We might still be interested in proceeding as is...

PR #321, opened 2020-08-12
add and updates the roast head reconstruction process

tmedani:
Hi 

This 2nd PR is about the roast head model generation, this new code corrects the previous version and aligns the head model to the SCS coordinate system.

The required functions are available on the bst-duneuro repo. They are mainly two functions which isolate the head reconstruction process from the whole roast process, it removes the electrodes from the head mesh and the TDCS process.  

This is as a first step for the integration of some Roast processes and results to Brainstorm.

We may work in the near future on the TMS/TDCS and DBS, and Roast has some similar functionalities, we may discuss all this one day.

A+

### DIFF
--- a/toolbox/process/functions/process_fem_mesh.m
+++ b/toolbox/process/functions/process_fem_mesh.m
@@ -296,7 +296,7 @@ function [isOk, errMsg] = Compute(iSubject, iMris, isInteractive, OPTIONS)
     end
         
     % ===== LOAD/CUT T1 =====
-    if ismember(lower(OPTIONS.Method), {'brain2mesh', 'simnibs'})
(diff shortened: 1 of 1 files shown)


ftadel:
I'm checking this PR and reorganizing "a bit" the code, as usual :-)

Where is **meshByIso2meshWithoutElectrode**?

tmedani:
Hi @ftadel 

you can find it here 
https://github.com/tmedani/duneuro_interface/blob/master/toolbox/meshByIso2meshWithoutElectrode.m

> I'm checking this PR and reorganizing "a bit" the code, as usual :-)
as usual, 👍, I try to avoid you this but I know that always it needs a better touch from your side :)

Issue #142, opened 2018-11-29
Connectivity: Update functions and tutorials


ftadel:
Waiting for updates from Richard and Hossein.

This tutorial was started a few years ago and never actually written:
https://neuroimage.usc.edu/brainstorm/Tutorials/Connectivity

There is also this page where I was trying to centralize the summary of some of the discussions:
https://neuroimage.usc.edu/brainstorm/ConnectivityAdmin
All the information in there should be either deleted (if of no interest anymore), moved to this issue (for topics that still need to be discussed), or moved to the connectivity tutorial (for info relevant for users).

A section "Connectivity" is also available in the general todo-list:
https://neuroimage.usc.edu/brainstorm/Next#Connectivity

TODO:
- Define example dataset and precise results to obtain from them
- How to deal with unconstrained sources? See:
   -  http://neuroimage.usc.edu/forums/showthread.php?2401
   - https://neuroimage.usc.edu/forums/t/functional-connectivity-unconstrained-sources-and-pca/30486/2
- How to assess significance from connectivity matrices?
- All the functions using bandpass filters must be updated to use the new filters

HosseinShahabi:
Hi @martcous @ftadel
I'm trying to add an image to the tutorial. I uploaded the image in different formats and used the script to attach to the body but it doesn't work. I used this code: 

<img alt="FlowChartGeneral3.gif" height="250" src="/brainstorm/Tutorials/Connectivity?action=AttachFile&amp;do=get&amp;target=FlowChartGeneral3.gif" title="attachment:FlowChartGeneral3.gif" width="579" />

Would you please take a look at it? Also, can I upload .png files? 

Thanks

ftadel:
The button to insert images doesn't work in the GUI editor: you have to first upload the file on the page (Menu > Attachments), then use the syntax {{attachment:filename.gif}}

GIF format recommended for interface screen captures (smaller), but you should be able to use any file format supported in web browsers (including png)

ftadel:
New questions on the forum:

- https://neuroimage.usc.edu/forums/t/connectivity-tutorial-and-methods-development-on-bst/12223
- https://neuroimage.usc.edu/forums/t/connectivity-with-imaginary-coherency-and-corrected-with-ft-sourcestatistics-plotting-issues/12158/7

Message from Richard, 20 Jun 2019:
[...]

PR #323, opened 2020-08-16
Updated functions - Part 2 Connectivity

HosseinShahabi:
Hi @ftadel 
I've updated the functions based on your suggestions and group discussion. Now we can easily archive the current Amplitude Envelope Correlation function since the results are similar. I changed to code so it asks Brainstorm to start the parallel processing toolbox. It works but I guess every time you check the box it starts it and then shuts it down after the process ends, immediately. 
Let me know if there is any problem. 
Thank you.


ftadel:
Thanks, this looks good!

I reformatted a few things in these additional commits:
- https://github.com/brainstorm-tools/brainstorm3/commit/f375c25554aff670cebd1ddee318693915e338a2
- https://github.com/brainstorm-tools/brainstorm3/commit/f81bda70c9ebad55d5b5ba8d472f2c054936e5ea
I haven't read the tutorial yet...

Now this PR is closed, let's move the discussion back to the corresponding issue:
https://github.com/brainstorm-tools/brainstorm3/issues/142

PR #325, opened 2020-08-19
Fitting Oscillations and One-Over-F (FOOOF) support for Brainstorm (really this time)

lucwilson:
I think second time's the charm! Finally, the only files left with changes are the ones which _actually_ have changes. You were right in how to solve the issue of being behind the current build (although I have no idea where the older files were as I had manually replaced them with the newest...?), so thank you for helping with that. Again, here is the description for FOOOF and its required additions/alterations:

The Fitting Oscillations and One-Over-F (FOOOF) algorithm is designed to identify and parametrize features of the neural power spectrum through nonlinear modelling.

New functions:
toolbox/process/functions/...
* process_fooof.m: Main FOOOF process file.
* process_fooof_analysis: FOOOF parameter/statistics extraction/organization function.

toolbox/timefreq/...
* process_fooof_bands.m: support function for frequency band organization in process_fooof_analysis.
* panel_fooof_options.m: advanced options panel for process_fooof.
* panel_fooof_analysis_options.m: advanced options panel for process_fooof_analysis.

Updated functions:
toolbox/gui/...
* figure_spectrum.m: Support for FOOOF model PSD.
* panel_display.m: FOOOF visualization tools.

toolbox/db/...
* db_template.m: Added FOOOF-related fields to loadedtimefreq.

toolbox/core/...
* bst_get.m: Added FOOOF presets to library.
* bst_set.m: Added FOOOF processes to library of options.
* bst_memory.m: FOOOF field support for LoadedTimefreqFile.


ftadel:
This looks much better, thanks!

This is quite a lot of code and documentation, I will need some time to review it. 
I'll try to do it by end of next week.

ftadel:
cc: @sbaillet @martcous 

I reviewed the online tutorial : congratulations for this great work!
It reads very well, it is compact and easy to follow, and includes all the external references it needs. I only modified a few minor things in the layout and added some screen captures.
I gave a quick look at the code: it looks very clean, well documented, following strictly all the conventions in use in the rest of the software and therefore easy to read for other developers.

Before moving forward with the final merge in the Brainstorm distribution, I have some comments and modifications to propose.

### 1. Do we really need all these options?

I didn't test all the parameters, and wouldn't know what to expect from them, therefore whether the combination of all the possible parameters work is up to you. 

The problem with methods with too many parameters is that we don't know how to tune them when we're not experts in the field. Some advanced documentation would be interesting: examples of failure of the default parameters, together with instructions on how to modify the various parameters in order to obtain improved results. But maybe this is more a work for the authors of the method. If it is already done, or if the influence of all the parameters is discussed in the article, it would be good to mention it in the tutorial, at the end of the section "Initializing and running FOOOF".

In general, in Brainstorm, we try to hide as much complexity as possible, for allowing users with limited skills in signal processing and electromagnetic modelling (MDs, psychology students...) to understand what they do and navigate in the interface with limited risks of manipulation errors. All the algorithms we use have a lot more options than what we offer in the interactive interface, to make it easier to use for the general public. Experts use FieldTrip or MNE-Python anyway.

All the parameters you think a Brainstorm user has no clear reason to modify (ie. you wouldn't be able to provide an example dataset and write a tutorial justifying their use) could do more harm than good and could be removed from the interface, and hard coded as fixed parameters at the beginning of the computation functions. Expert users or developers interested in all the possibilities offered by FOOOF would be able to use modify them in the code, use the functions from the command line, or directly in Python.

If there are some parameters you think you might like to hide, please let me know.

### 2. Interface simplification

The panels foof_options and foof_analysis_options are not used except in the FOOOF processes options. All the options they contain can be integrated directly as options of the processes. It would make the future modification and maintenance of these processes much easier. Previous values of all the parameters would be saved automatically, therefore the modifications to bst_get and bst_set would become useless.

### 3. Process grouping

Having a second process "Analyse FOOOF features" for the last steps of feature extraction is maybe not justified. This process is very fast, it does not create a lot of data, it does not represent a lot of code, it overwrites the same file. The initial FOOF process is fast as well, therefore a user interested in trying different feature extractions can simply recompute the whole thing mulitple times.
I understand there is probably some added value to having it split in two locally for this specific research topic, but:
- it adds complexity to the full processing scripts that go from importing anatomy to connectivity analysis in source space, 
- it makes the maintenance of the software heavier (if everything related to one topic is in the place, it is easier to understand and manipulate), 
- it overloads the interface and the list of files.
I tend to prefer more compact solutions, therefore I would prefer to have it grouped together the main FOOOF process.

Regarding the function process_fooof_bands.m:
- The name should not start with "process" (reserved for actual processes)
- Would be better included in the single merged process function.
- I'll probably rewrite it to use the existing frequency bands functions (adapting them to the missing "function" information).

### 4. Python implementation 

There is something missing in the instructions. If I only use pyenv/pyversion to set the python 3.7 environment in Matlab, the FOOOF process returns an error with Matlab 2020a.

```
>> pyenv

ans = 

  PythonEnvironment with properties:

          Version: "3.7"
       Executable: "C:\Users\franc\Anaconda3\envs\mne\python.exe"
          Library: "C:\Users\franc\Anaconda3\envs\mne\python37.dll"
             Home: "C:\Users\franc\Anaconda3\envs\mne"
           Status: Loaded
    ExecutionMode: InProcess
        ProcessID: "66376"
      ProcessName: "MATLAB"

>>

** Error: [process_fooof]  Frequency > FOOOF > Generate FOOOF models
** Line 34:  _distributor_init> (line 34)
** Python Error: ImportError: DLL load failed: Le module spécifié est introuvable.
** 
** Call stack:
** >_distributor_init.m><module> at 34
** >__init__.m><module> at 140
** >fit.m><module> at 54
** >__init__.m><module> at 3
** >__init__.m><module> at 6
** ><frozen importlib.m>_call_with_frames_removed at 219
** ><frozen importlib.m>exec_module at 728
** ><frozen importlib.m>_load_unlocked at 677
** ><frozen importlib.m>_find_and_load_unlocked at 967
** ><frozen importlib.m>_find_and_load at 983
** ><frozen importlib.m>_gcd_import at 1006
** >__init__.m>import_module at 127
** >process_fooof.m at 30
** >bst_process.m>Run at 229
** >bst_process.m at 36
** >panel_process1.m>RunProcess at 151
** >panel_process1.m at 26
** >gui_brainstorm.m>CreateWindow/ProcessRun_Callback at 772
** >bst_call.m at 28
** >gui_brainstorm.m>@(h,ev)bst_call(@ProcessRun_Callback) at 296
** 
** 
** File: Subject01/@rawS01_AEF_20131218_01_600Hz_notch/timefreq_psd_200618_2105.mat
** 
```

If I call `bst_mne_init('Initialize', 1);` first, I don't have any problem. 
(for this to work, the python environment needs to be set in the Brainstorm preferences).
(yes, this function needs to be updated with pyenv to replace pyversion for Matlab >= 2019b)

Another issue:
Calling process_foof a second time throws tons of warnings. It looks like importing multiple times the scipy module is not much appreciated... Some checks should be added not to call import_module repetitively.

```
Warning: The name 'delete' is already in use as a method name.  This will become an error in a future release. 
> In process_fooof>Run (line 99)
  In process_fooof (line 30)
  In bst_process>Run (line 229)
  In bst_process (line 36)
  In panel_process_select>ShowPanel (line 2903)
  In panel_process_select (line 30)
  In bst_report>Recall (line 1227)
  In bst_report (line 58)
  In bst_call (line 28)
  In gui_brainstorm>@(h,ev)bst_call(@bst_report,'Recall','current') (line 135) 
```

### 5. Backward compatibility

The use of tables in process_fooof_analysis.m is OK, but limits the backward compatibility to Matlab 2013b, while almost everything in Brainstorm still works with 2009b (useless in North America, a bit more interesting in South Asia or Africa).

Arrays of structures provide easier and faster access to the data contents, easier to manipulate in Matlab (with the syntax `[struct_array.fieldname]`), and with smaller memory and disk size. Sorting rows is easy with something like `[tmp,iSort]=sort([struct_array.index_to_sort]); struct_array=struct_array(iSort);`.  Whenever advanced features of tables are not required, I recommend using arrays of struct instead.

This is a simple suggestion. I'm not sure yet I would invest time in rewriting this.

### 6. Before merging

- **Starting from tomorrow**, I will work on rewriting the processes (merge processes, linearize options, remove extra panels, remove extra functions), unless you would like to do it yourself, or you think there is a serious issue about these simplifications.
- Don't hesitate to tell me if you disagree with my suggestions: I am not an expert in this field, I can make wrong assumptions and poor decisions.
- Let me know if there are some of these topics you'd like to discuss in a skype/zoom call, I am avaialable this week. 
- If there are parameters you'd like to hide, please let me know ASAP.
- Please look into the Python environment issues

Thank you for the very high quality of this work, it will be a very valuable contibution to the Brainstorm software.

ftadel:
FYI: This is what the options integrated directly in the process will look like (matlab/python).

image
 
image

lucwilson:
Hi Francois,

Thank you for your very detailed and informed suggestions regarding the implementations and their associated options. Sorry I didn't get back to you yesterday, but it looks like you're off to a great start in redesigning the UI for the better.

1. Do we really need all these options?
You are completely right about there being many options for users (perhaps too many). In my experience, the option I have noticed the least importance in modifying has been the **Peak threshold** option, as I believe using Minimum peak height addresses possible concerns about fitting noise more appropriately. Additionally, I believe the MATLAB implementation's **Threshold after fitting** option should also always be selected, as this mirrors the Python FOOOF closest by removing peaks that do not satisfy a user's predetermined conditions for peaks. I agree with you on both of these options being removed from the options pane and hardcoded to a particular value, allowing advanced users to modify it within the code if they would like. That being said, I believe all of the other options are still important to change with respect to each dataset.

2. Interface simplification
Your changes to include the options from process_fooof_options in the initial options pane looks much better than how I had previously designed it, great work! One note from the screenshot you attached is that both Python and MATLAB versions of FOOOF allow for different aperiodic modes, so after that is not controlled by the radio button, everything will be controlled correctly. There are a few other minor things, but I can correct these myself.
The options for process_fooof_analysis_options are controlled in a complex manner, but if you can find a way to simplify them to fit in the initial options pane, then I am also completely happy with this compromise. Even if it is not the _exact_ same, whatever you think is best.

3. Process grouping
The original python FOOOF contains statistics by default, so I don't think including them in this build would be an issue. Merge as you wish!

4. Python implementation
(Darn, I thought I had fixed this!)
I will look into this some more, thank you for bringing it to my attention.

5. Backward compatibility
I can quickly change these data structures, no worries. (In truth, I was reluctant to use tables to begin with.)

Here is the order to which I will address the relevant issues above (should be done the first two today):
- Interface simplification (just changing some wording)
- Backward compatibility
- Python implementation

Again, thank you so much for your help! I think we are on the right track even without a video meeting, but if something major comes up that you would like to change, just send me a message and we can meet to discuss :)

ftadel:
Does this look good to you?
(I also removed the option "all frequencies", as by default the panel will offer the entire spectrum)

image image

ftadel:
Wait... given the test you're doing at the beginning, you seem to force excluding f=0Hz from the input spectrum.
Is it really something that should be enforced? Or can someone include the first frequency bin as well?
If you are sure you never want it: we could also remove f=0Hz even if it is passed in input.

ftadel:
With everything grouped in the same process:
image
image

lucwilson:
> Wait... given the test you're doing at the beginning, you seem to force excluding f=0Hz from the input spectrum.
> Is it really something that should be enforced? Or can someone include the first frequency bin as well?
> If you are sure you never want it: we could also remove f=0Hz even if it is passed in input.

The model will fail if 0 Hz is left as an input, as we log transform the input frequencies (leads to -inf). We can auto-remove 0 Hz, but we should ensure that this is reflected in the options after performing the process (FOOOF_options struct). I think we can just modify the freqBand (fB) lower bound up one index position from the frequency vector.

The combined process options look great! There is no chance I could have been able to produce this result myself, so thank you so much for your involvement.

ftadel:
Done.
I also modified the saved structure. I put the FOOOF structure in the Options field, so that we don't have to modify the structure definitions, or bst_memory. Your old files will not work anymore, unless you rename the structures.

Could you please test that it still works?
And then update the tutorial accordingly?

Let me know when you're done, so I can merge the PR and make add the FOOOF tutorial link to the tutorial page.
Thanks

lucwilson:
Hi Francois, 

I've just finished running about every arrangement of options I can think of and everything appears to be running as expected! The only change I would make would be to the frequency range option to maintain whatever range was previously set in the options UI; having to keep switching from 0-max to a desired range can become tedious. All other options are working perfectly!

I will update the tutorial now and add another comment here when I am finished.

lucwilson:
Hi Francois,

The tutorial is updated and complete. Ready when you are :)

ftadel:
> I've just finished running about every arrangement of options I can think of and everything appears to be running as expected! The only change I would make would be to the frequency range option to maintain whatever range was previously set in the options UI; having to keep switching from 0-max to a desired range can become tedious.

Done. But then there is no easy option to select all the frequencies in the file (I removed the "all frequencies" options because I added this dynamic selection). And this can become annoying too. 
The choice has to be made based on what will be mostly used. If your recommendation is to always run FOOOF only on the lower part of the spectrum, then this new static version that will offer [1-40] by default is fine.

ftadel:
Thank you for double-checking everything, and for fixing all the errors I left in the code. I had not checked that the feature extraction was still working identically, sorry about about that...

Regarding the setting up of the Python environment: it should be mentioned that the python bin and lib must be on the system path, which is not always the case on windows. I move the information about the Python implementation at the end, and added some details about setting the system path. Feel free to edit if needed.
https://neuroimage.usc.edu/brainstorm/Tutorials/Fooof#Python_implementation

sbaillet:
Fantastic work!

Thank you both very much - this is a very significant addition to the Brainstorm toolkit!

Sylvain.

lucwilson:
Thank you for all of your help, Francois! I'm happy to see this make it to the main build :)

tmedani:
Congrats for this great job!

PR #330, opened 2020-08-30
add interactive option to view the tensors

tmedani:
New option to view the tensors,
when the tensors are displayed either on the MRI or the Mesh,
-> Shift + uparrow ==> increase the size of the ellipsoid or the line
-> Shift + downarrow==> decrease the size of the ellipsoid or the line
-> Shift + space ==> switch from ellipsoids view to lines and inversely

### DIFF
--- a/toolbox/gui/figure_3d.m
+++ b/toolbox/gui/figure_3d.m
@@ -1009,7 +1009,49 @@ function FigureKeyPressedCallback(hFig, keyEvent)
                     if isempty(TimeSliderMutex) || ~TimeSliderMutex
                         panel_time('TimeKeyCallback', keyEvent);
                     end
-                    
+           % === 'UPARROW', 'DOWNARROW','SPACE' : change fem tensor view ===
+            case {'uparrow', 'downarrow','space'}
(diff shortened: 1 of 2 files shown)


ftadel:
I added popup menus to configure the display with a right-click on the figure > FEM tensors > ...

I also reorganized a bit the code, to have all the properties saved in the same structure (the figure Handles structure) instead of putting different parameters in different places (I removed the access to UserData).

tmedani:
Super, 
Thanks for the re-organization, 
I will test this version.

tmedani:
Hi @ftadel 

FYI, not important but sometimes, this error appears when reduce/increase the size of the tensors and scroll with the mouse

image

ftadel:
> FYI, not important but sometimes, this error appears when reduce/increase the size of the tensors and scroll with the mouse

I added a check to prevent getting properties from objects that were already deleted:
https://github.com/brainstorm-tools/brainstorm3/commit/f8c0a79b018e39c3bccb7a1093e96b42207587fb

juangpc:
Hey, this is just great! 👍

PR #332, opened 2020-09-17
minor updates and useful options for the bst-duneuro

tmedani:
This mainly contains the following
- convert the format of the mesh element to uint64 (it's in this patch but it seems not required)
- add the possibility to check the source space when the keywords wm and gm are not used (we add the inner and scf key words)
- reduce the number of tests nFix  from 10 to 3, we will work in a better option soon, further is better, 
- add the possibility to save the real fem source space if it is different from the original as a checkbox from the duneuro panel, this can be useful for further investigation
- add the options cache.enable in the case of MEG, this is important since the MEG with high-density mesh and sensors require lot memory,  but this will slow the computation. Ideally, we may keep it for regular mesh and use it for high mesh density or when the MEG fails. This is also a checkbox option from the duneuro panel

### DIFF
--- a/toolbox/io/out_fopen_spm.m
+++ b/toolbox/io/out_fopen_spm.m
@@ -19,8 +19,10 @@ function sFileOut = out_fopen_spm(OutputFile, sFileIn, ChannelMat)
 % For more information type "brainstorm license" at command prompt.
 % =============================================================================@
 %
-% Authors: Francois Tadel, 2017
+% Authors: Francois Tadel, 2017-2020
 
+% Check SPM installation
+bst_spm_init();
 % Check if SPM is in the path
 if ~exist('file_array', 'file')
     error('SPM must be in the Matlab path to use this feature.');


toolbox/forward/bst_duneuro.m
```diff
@@ -154,7 +154,9 @@
 end
 % Write mesh model
 MeshFile = fullfile(TmpDir, MeshFile);
-out_fem(FemMat, MeshFile);
+FemMatx = FemMat;
```
ftadel, line 157: All the file formats saved by out_fem are text-based.
I don't think converting to uint64 can make any difference.
Please revert this change.

toolbox/forward/bst_duneuro.m
```diff
@@ -179,6 +181,13 @@
         % Force all the dipoles within the GM layer
         iGM = find(panel_duneuro('CheckType', FemMat.TissueLabels, 'gray'), 1);
         iWM = find(panel_duneuro('CheckType', FemMat.TissueLabels, 'white'), 1);
+        % apply same test in the case where wm or gm are not found.
+       if (iGM || iWM) == 0
+        iGM = find(panel_duneuro('CheckType', FemMat.TissueLabels, 'inner'), 1);
```
ftadel, line 186: Please pay attention to respecting the indentation of the code.
It's not difficult to do and it makes the code easier to read.
You can simply select the lines you edited in the file and press CTRL+I, Matlab will align the code correctly for you.

toolbox/forward/bst_duneuro.m
```diff
@@ -179,6 +181,13 @@
         % Force all the dipoles within the GM layer
         iGM = find(panel_duneuro('CheckType', FemMat.TissueLabels, 'gray'), 1);
         iWM = find(panel_duneuro('CheckType', FemMat.TissueLabels, 'white'), 1);
+        % apply same test in the case where wm or gm are not found.
+       if (iGM || iWM) == 0
```
ftadel, line 185: iGM and iWM are values that can be empty, therefore this test can crash.
You should test this with:
`if isempty(iGM) && isempty(iWM)`

toolbox/forward/bst_duneuro.m
```diff
@@ -179,6 +181,13 @@
...
+        iGM = find(panel_duneuro('CheckType', FemMat.TissueLabels, 'inner'), 1);
+       end
+       if (iGM || iWM) == 0
```
ftadel, line 188: Same: invalid test.

toolbox/forward/bst_duneuro.m
```diff
@@ -293,7 +302,23 @@
 %                     end
 %                 end
 %             end
-
+           %% save new source space to database/ can be useful for control and display results
+           if cfg.BstSaveSourceSpace
+           % New surface structure
+            NewTess = db_template('surfacemat');
```
ftadel, line 308: Indentation

toolbox/forward/panel_duneuro.m
```diff
@@ -226,10 +232,22 @@
     % ==== PANEL RIGHT: OUTPUT OPTIONS ====
     jPanelOutput = gui_river([1,1], [0,6,6,6], 'Output options');
         % Save transfer matrix
-        jCheckSaveTransfer = gui_component('checkbox', jPanelOutput, '', 'Save transfer matrix', [], '', [], []);
+        jCheckSaveTransfer = gui_component('checkbox', jPanelOutput, 'br', 'Save transfer matrix', [], '', [], []);
```
ftadel, line 235: 'br' is for moving to the next line.
This is the first line: not needed here.

toolbox/forward/panel_duneuro.m
```diff
@@ -226,10 +232,22 @@
...
             jCheckSaveTransfer.setSelected(1);
         end
+        % Save source space
+        jCheckSaveSourceSpace = gui_component('checkbox', jPanelOutput, 'br', 'Save source space', [], '', [], []);
+        if (OPTIONS.BstSaveSourceSpace)
+            jCheckSaveSourceSpace.setSelected(1);
+        end
+        if isMegOnly
+            % Enable cach memory ==> useful only for high mesh density but slower/ not required for basic meshs 
+            jCheckMegCacheEnable = gui_component('checkbox', jPanelOutput, 'br', 'Enable cache memory', [], '', [], []);
```
ftadel, line 246: Is this really an "output option"?
I don't know at which level it impacts the computation, but it might require a new panel "Computation options"?

toolbox/forward/panel_duneuro.m
```diff
@@ -226,10 +232,22 @@
     % ==== PANEL RIGHT: OUTPUT OPTIONS ====
     jPanelOutput = gui_river([1,1], [0,6,6,6], 'Output options');
         % Save transfer matrix
-        jCheckSaveTransfer = gui_component('checkbox', jPanelOutput, '', 'Save transfer matrix', [], '', [], []);
+        jCheckSaveTransfer = gui_component('checkbox', jPanelOutput, 'br', 'Save transfer matrix', [], '', [], []);
         if (OPTIONS.BstSaveTransfer)
             jCheckSaveTransfer.setSelected(1);
         end
+        % Save source space
+        jCheckSaveSourceSpace = gui_component('checkbox', jPanelOutput, 'br', 'Save source space', [], '', [], []);
```
ftadel, line 240: This could make more sense in the panel "Source space", right?
And maybe be disabled if both "Shrink source space" and "Force source space inside brain" are disabled?
(I can make this last thing for you if you validate the logic)

ftadel (changes requested):

tmedani:
Hi @ftadel , 
sorry for my low reaction, 
I'm working on it, and some other improvements, 
I will push all the changes asap.

PR #335, opened 2020-09-24
Error fix (that I caused) for FOOOF

lucwilson:
You can never be too cautious when making even the simplest of patches. Always try running your code before you publish it!

- Referenced a data structure "opts" when none exists (should have been "opt"). Instantly crashes the process. This has been fixed.

Consider my lesson on publishing quick fixes learned.


ftadel:
I didn't test this fix, or the previous one, sorry... 
I review pull requests mostly to make sure they don't damage anything else in the software. But I won't prevent you from damaging your own functions :)

PR #297, opened 2020-05-21
Support for Phase Locking - Circular Plots

mpompolas:
CircStat toolbox added.
The process_spiking_phase_locking allows visualization of the preferred phase neurons fire based on the oscillations of the selected channels, on selected frequencies

PR #338, opened 2020-10-06
Added Phase locking circular plots

mpompolas:
**Users can use the function Spiking Phase Locking to see the phase selectivity of individual neurons to ongoing oscillations of selected channels.**
The lfps from the selected channels can be filtered in selected frequencies to reveal spectral phase selectivity.

This is a clean pull request that is based on: https://github.com/brainstorm-tools/brainstorm3/pull/297#issue-421621488 addressing the changes requested there.

Usage:
image

Visualization:
The dropdown is used to select the combination of the existing neurons in the trials and the selected channels (they are taken in a combinatorial way as shown on the image below nNeurons x nChannels):

image


ftadel:
This looks great, but we would need a proper tutorial for this.
The button "Online tutorial" of the process links to the CircStat toolbox article (https://www.jstatsoft.org/article/view/v031i10), which doesn't doesn't tell us much about how to use it in Brainstorm.
Is it in your plans to write it?

I reviewed the code and the changes look harmless to the rest of the application, I could merge it.
But I don't know how to test this. 
Do you want me to merge without testing?
@martcous ?

mpompolas:
The tutorial is a great idea. I'm sure it will help a lot the users.
I'll add it on my todo list but things are a bit hectic right now with work, so I'll do it in the near future.

**If you want to test:**
You need spiking events - Spikes Channel "channelLabel" (just add some timestamps randomly throughout a trial)
I have a commented-out section in process_spiking_phase_locking (line 204-211) that forces the spikes to be on selected phase bins.

I used this as a check to the final visualization. You can try it out as well and give me feedback.

martcous:
Great job, Kostas! It seems to work OK on my end, except when a channel has no group (ChannelMat.Channel(iChannel).Group is empty), your process_spiking_phase_locking's select_channels() function crashes. If this is not a state that can happen normally then you can disregard my comment.

mpompolas:
I have the feeling that this would occur only when the users **specify a Group** that doesn't exist at all, not when the channels don't have a Group assigned to them.
Is this what you're getting?

image

martcous:
No, I meant when a specific channel has an unassigned group. However, now that I tried to get specific steps to reproduce the issue, I can't seem to reproduce it anymore. My channel file must have been in an abnormal state (this was an old protocol), please disregard my comment, all good on my end!

ftadel:
Thanks!

PR #341, opened 2020-10-15
Improve the bst-iso2mesh version for fem mesh and add new option to refine the fem mesh

tmedani:
This first version of this PR is related to the FEM mesh generation with **iso2mesh** option and the mesh refinement,
 
1- Remove most of the previous version and implement a new version, which I think is more stable
=> possibility to generate mesh from any merged surfaces with non-predifined names and the labels will be identified
=> All the unknown element labels from surf2mesh are identified by computing the distance to the center of the brain and then check with the input surfaces ID.
This can be useful in the case of high mesh densities, and  where only the inner layers are required, 
for example, wm, gm and csf for sEEG

2- Add a new option to re-fine a specific area of the FEM mesh (new function by the right click on the FEM mesh => Refine FEM mesh)
=>The refinement or the adaptative mesh area can be defined by a set of 3D points where the mesh can be refined, by dividing the mean edge length by a factor that the user can change. 
I have added the following options, that I think are useful : 
-a- areas surrounding the cortex,
-b- area surrounding the position of the electrodes (eeg/ecog and seeg)  
-c- both a and b 
-d- load any 3D points from file or Matlab (TODO)

The following example shows a case where the refined area is around the cortex surface. 
image

and this one around an sEEG electrodes
image

@ftadel the code is working correctly however it's not clean, 
I left some commented codes that I'm using for visualization and testing. 
If you agreed with this new version and its results I can clean it, and improve it.
And also optimize the code since either the refinement and the fem mesh generation uses ~the same process.


toolbox/process/functions/process_fem_mesh.m
```diff
@@ -356,34 +356,45 @@
...
+                end
+                faceList = unique(BemMat.Faces);
+                nodeList = BemMat.Vertices(faceList,:);
+                % Find the largest distance between two point
+                maxYcoor = max(nodeList(:,2));
+                minYcoor = min(nodeList(:,2));
+                maxYpoint = [center_inner(1) maxYcoor center_inner(3)];
+                minYpoint = [center_inner(1) minYcoor center_inner(3)];
+                % Find the nearest node on the mesh and update
+                k = dsearchn(nodeList,[maxYpoint;minYpoint]);
```
ftadel, line 385: If you need this to be fast, in some configurations bst_nearest can be much faster.

toolbox/process/functions/process_fem_mesh.m
```diff
@@ -356,34 +356,45 @@
...
+                maxYpoint = nodeList(k(1),:); minYpoint = nodeList(k(2),:);
+                distance_in(iBem) = norm(maxYpoint-minYpoint);
+                listPointasSeed(iBem,:) = maxYpoint;
             end
+            %  Sort from inner to outer
```
ftadel, line 390: I thought that the detection of the layer type based on the distance was NOT working? 
Is it something very different that you did here, than the version that we deleted before?

What was wrong with the idea of identifying the layers with their labels? (which you delete here)

toolbox/process/functions/process_fem_mesh.m
```diff
@@ -406,30 +417,14 @@
...
-            regions = repmat(orig, nBem, 1) + distSeed * v0;
-            
+            regions = listPointasSeed;
             % Create tetrahedral mesh
             bst_progress('text', 'Creating 3D mesh (Iso2mesh/surf2mesh)...');
             factor_bst = 1.e-6;
             [node,elem] = surf2mesh(newnode, newelem, min(newnode), max(newnode),...
-                OPTIONS.KeepRatio, factor_bst .* OPTIONS.MaxVol, regions, [], [], 'tetgen1.5');
-            
+                OPTIONS.KeepRatio, factor_bst .* OPTIONS.MaxVol, (regions), [], [], 'tetgen1.5');
```
ftadel, line 427: What is the point of adding parenthesis to "(regions)" ?

ftadel: Since the variable regions is not used anywhere else, use directly listPointasSeed instead.

toolbox/process/functions/process_fem_mesh.m
```diff
@@ -438,12 +433,45 @@
...
+           end
+            % replace with the correct ID
+            tmp = elem;
+            tmp(:,5) = tmp(:,5) +10; % translation to avoind overlapping
+            allLabels = unique(tmp(:,5));
+            for  ind = 1: length(allLabels) 
+                tmp(tmp(:,5) == allLabels(ind),5) = distOut_tmp(ind);
+            end        
+            % check again just in case
+             allLabels = unique(tmp(:,5));
```
ftadel, line 468: Please fix indentation

toolbox/process/functions/process_fem_mesh.m
```diff
@@ -438,12 +433,45 @@
...
+            % sort weired elements, works in all case even when there are
+            % more output tissues than inputs
+           distOut_tmp = distance_out;
+           for ind = 1 : length(distance_in)
+               tmp = find(round(distOut_tmp,3)<=round(distance_in(ind),3));
+               distOut_tmp(tmp) = ind;
+           end
+            % replace with the correct ID
+            tmp = elem;
+            tmp(:,5) = tmp(:,5) +10; % translation to avoind overlapping
```
ftadel, line 462: avoid

toolbox/process/functions/process_fem_mesh.m
```diff
@@ -438,12 +433,45 @@
...
+            end
+            % sort weired elements, works in all case even when there are
+            % more output tissues than inputs
+           distOut_tmp = distance_out;
+           for ind = 1 : length(distance_in)
+               tmp = find(round(distOut_tmp,3)<=round(distance_in(ind),3));
+               distOut_tmp(tmp) = ind;
+           end
+            % replace with the correct ID
+            tmp = elem;
```
ftadel, line 461: Variable is not used between `tmp = elem;` and `elem = tmp;`
Just work directly on `elem`.
Less variables = higher readability.

toolbox/process/functions/process_fem_mesh.m
```diff
@@ -1297,3 +1325,237 @@ function ComputeInteractive(iSubject, iMris, BemFiles) %#ok<DEFNU>
         NewFemFile = fem_tetra2hexa(FemFullFile);
     end
 end
+
+%% ===== Refine Mesh =====
```
ftadel, line 1329: This is a completely independent process, and should be handled as such.
Ideally we would like to be able to call from a script as well.
=> Create a new process_fem_refine.m

toolbox/process/functions/process_fem_mesh.m
```diff
@@ -1297,3 +1325,237 @@ function ComputeInteractive(iSubject, iMris, BemFiles) %#ok<DEFNU>
...
+        nodeList = node(elemList,:);
+        maxYcoor = max(nodeList(:,2));
+        minYcoor = min(nodeList(:,2));
+        maxYpoint = [center_inner(1) maxYcoor center_inner(3)];
+        minYpoint = [center_inner(1) minYcoor center_inner(3)];
+        k = dsearchn(nodeList,[maxYpoint;minYpoint]);
+        maxYpoint = nodeList(k(1),:); minYpoint = nodeList(k(2),:);
+        distance_out(ind) = norm(maxYpoint-minYpoint);
+    end
+    % sort weired elements, works in all case even when there are
```
ftadel, line 1515: There is a lot of copy-paste between this function and the iso2mesh mesh generation.
Could try factorizing a bit the code? 
(= add a new function, either stand-alone fem_...m if it makes sense, or included in process_fem_mesh otherwise)

toolbox/tree/tree_callbacks.m
```diff
@@ -1206,6 +1206,7 @@
                     gui_component('MenuItem', jPopup, [], 'Convert tetra/hexa', IconLoader.ICON_FEM, [], @(h,ev)bst_call(@process_fem_mesh, 'SwitchHexaTetra', filenameRelative));
                     AddSeparator(jPopup);
                     gui_component('MenuItem', jPopup, [], 'Resect neck', IconLoader.ICON_FEM, [], @(h,ev)bst_call(@fem_resect, filenameFull));
+                    gui_component('MenuItem', jPopup, [], 'Refine FEM mesh', IconLoader.ICON_FEM, [], @(h,ev)bst_call(@process_fem_mesh, 'RefineMesh', filenameRelative));
```
ftadel, line 1209: Call process_fem_refine('Compute', ...) instead

toolbox/process/functions/process_fem_mesh.m
ftadel: When moving this to a new process: 
Separate this whole thing in two functions: Compute and ComputeInteractive, as in many other processes, in order to have both an interactive and a command line call.

ftadel (changes requested):

toolbox/process/functions/process_fem_mesh.m
tmedani: ok, I don't know this function. 
I will check it. Thanks

toolbox/process/functions/process_fem_mesh.m
tmedani: >> I thought that the detection of the layer type based on the distance was NOT working?
I'm using another approach, where I use the triangular mesh as a reference for the distances,
which is not used before. 

>>Is it something very different that you did here, than the version that we deleted before?

Yes, this is totally different, I think this method is more "robust" it can define the unknown labels (outputs of surf2mesh), and also reorder them according to the reference distances.
 

>>Is it something very different that you did here, than the version that we deleted before?

in most case, this error appears if we use different surfaces than (inner, outer and scalp) 
 errMsg = ['Problem with Tetget: Brainstorm cannot understand the output labels (' num2str(allLabels(:)') ').'];

and if the output labels from surf2mesh are different from [1 2 3] the process stops, even if the tetgen works fine.

for example, here is the error if I use white, csf, and grey 👍 
image

but with the new method, this is working, even if the outputs labeled (1 2 3 4) are weird
image

ftadel: > but with the new method, this is working, even if the outputs labeled (1 2 3 4) are weird

But they are not labelled correctly, and the predefined colors are incorrect?
Why not keep the identification by name AND use your new code when the layers do not have standard names?

The standard case will be that people will use fully automated pipelines, like simnibs or brain2mesh, which will produce layers that we can clearly identify. Why not benefit from this?
The case where people mix and match surfaces like you do will be very marginal.

tmedani:
This PR needs more work and changes. 

I will close it, and create a new PR asap.

PR #342, opened 2020-10-15
Adding constrained optimization for Matlab FOOOF and visualization tools in Brainstorm for frequency-wise error

lucwilson:
Once these changes are approved, I will also add new sections to the tutorial to reflect them.

For visualizing frequency-wise error, it makes the most sense (based on how the values are stored) to view frequency-wise error in log10 space; it also makes no sense in other scales, so all other scaling options are disabled for figures currently displaying frequency-wise error.

Happy to answer any questions related to the purpose of each line change/addition!


toolbox/process/functions/process_fooof.m
```diff
@@ -147,6 +147,12 @@
         return
     end
     
+    hasOptimToolbox = 0;
+    if exist('fir2','file') == 2 && strcmp(implementation,'matlab')
```
martcous, line 151: This function is part of the signal processing toolbox, not the optimization toolbox. You could check for the existence of 'fmincon' instead.

martcous:
Looks great on my end!

martcous (approved):

PR #337, opened 2020-10-01
ENH Add log scale to view_leadfields figure

juangpc:
Hi, 

Sometimes the lead field vectors decay pretty fast. This adds the option to log scale their norm.

If anybody thinks this is unneeded or wrong. Feel free to drop the pr. Only trying to help.

@tmedani, perhaps you can rebase from here the changes you're adding to this function. If the rebase doesn't work automatically, feel free to add manually the changes you see in here or and then drop it yourself.

Thanks,

PR #343, opened 2020-10-16
Enhance the LF viewer and adding more options from the keyboard

tmedani:
Hi @ftadel, 

This PR is to complete the PR that @juangpc proposed (https://github.com/brainstorm-tools/brainstorm3/pull/337#issue-496360490), and it includes also the proposed option.

It includes useful options to visualize and manipulate the LF vectors, all these options are managed from keyboard shortcuts, they can be displayed with the 'H' keyboard key, as in this figure : 

image

With this version, the average reference option can be also selected from the same list as the channels (R) instead to have multiple popup panels, the "target channels (T) can be also selected from a dedicated list, which is useful in the case where users want to see the LF between two specific channels. 

There are also functions that change the scale, the length, and the width of the vectors. Furthermore, there is a way to threshold the displayed vectors by fixing the Amplitude threshold as shown in the figure. All this interactively with the figure & the keyboard (as usual with Branstom figure 👍  ).

For the standard view keyboard (1-9), I have used the same subfunctions "SetStandardView" that I have copy-pasted to this function. 

All these options are for visualization purposes, the next dev should focus on functions that can 
quantitatively compare these differents LF models. 

Also, the idea behind these new options is not only for the LF, but also to extend the use for the optical flow (https://github.com/brainstorm-tools/brainstorm3/pull/311), as previously discussed with @LeoNouvelle  @martcous & @sbaillet

I hope this will be useful :)

A+
Takfarinas


toolbox/gui/view_leadfields.m
```diff
@@ -149,7 +149,13 @@
 
 %% ===== DISPLAY LEADFIELD =====
 % Current sensor
+useLogScale = false;
+useLogScaleLegendMsg = 'Off';
 iChannel = 1;
+iQuiverSize = 1;
```
ftadel, line 155: The convention in the rest of the code is "iVariableName" being an index. 
https://neuroimage.usc.edu/brainstorm/Tutorials/Scripting#Naming_conventions

These are not indices.
Can you please rename the variables?

toolbox/gui/view_leadfields.m
```diff
@@ -201,16 +245,81 @@ function KeyPress_Callback(hFig, keyEvent)
...
-                    '<TR><TD><B>Right arrow</B></TD><TD>Next channel</TD></TR>'....
-                    '<TR><TD><B>Page up</B></TD><TD>Previous 10th channel</TD></TR>'....
-                    '<TR><TD><B>Page down</B></TD><TD>Next 10th channel</TD></TR>'....
+                    '<TR><TD><B>Left arrow</B></TD><TD>Previous target channel (red color)</TD></TR>' ....
+                    '<TR><TD><B>Right arrow</B></TD><TD>Next target channel (red color)</TD></TR>'....
+                    '<TR><TD><B>Up arrow</B></TD><TD>Previous ref channel (green color)</TD></TR>'....
+                    '<TR><TD><B>Down arrow</B></TD><TD>Next ref channel (green color)</TD></TR>'....
+                    '<TR><TD><B>Shift + uparrow</B></TD><TD>Increase the vector length</TD></TR>'...
+                    '<TR><TD><B>Shift + downarrow</B></TD><TD>Decrease the vector length</TD></TR>'...                   
+                    '<TR><TD><B>Shift + L</B></TD><TD>Toggle on/off logarithmic scale</TD></TR>'...
```
ftadel, line 280: Looks like this is "Shift + Enter" instead

toolbox/gui/view_leadfields.m
```diff
@@ -201,16 +245,81 @@ function KeyPress_Callback(hFig, keyEvent)
...
+                    '<TR><TD><B>R</B></TD><TD>Select the <B>R</B>eference channel</TD></TR>'....
+                    '<TR><TD><B>T</B></TD><TD>Select the <B>T</B>arget channel</TD></TR>'....
                     '<TR><TD><B>S</B></TD><TD>Show/hide the source grid</TR>'....
-                    '<TR><TD><B>E</B></TD><TD>Show/hide the sensors</TD></TR></TABLE>'], 'Keyboard shortcuts');
+                    '<TR><TD><B>E</B></TD><TD>Show/hide the sensors</TD></TR>'...
+                    '<TR><TD><B>0 to 9</B></TD><TD>Change view</TD></TR>'...
+                    '</TABLE>'], 'Keyboard shortcuts');           
+                  
+                % === NUMBERS : VIEW SHORTCUTS ===
+            case '0'
```
ftadel, line 295: Delegate all this to figure_3d instead.
In the otherwise block, it should call figure_3d/FigureKeyPressedCallback and set the orientation.
Or am I missing something?

toolbox/gui/view_leadfields.m
```diff
@@ -351,32 +497,180 @@ function DrawArrows()
...
+    %% ===== GET LEADFIELD =====
     function GetLeadField       
         % Update the LF according to the selected channels only
         for iLF = 1:length(HeadmodelFiles)
             LF_finale{iLF} = HeadmodelMat{iLF}.Gain(iChannels,:);
         end
     end
+
+%% ===== LEADFIELD TO LOG SPACE =====
+    function lf_log = logScaledLeadField(lf)
```
ftadel, line 553: As this is a function, it would be better if it were named as an action.
For example: ScaleLeadfield

But since this is called only once and not very long, I suggest you integrate it directly into where it is required.
(and not keep it as a separate function)

toolbox/gui/view_leadfields.m
```diff
@@ -351,32 +497,180 @@ function DrawArrows()
...
+        t = atan2(rho,lf(:,3));
+        f = atan2(lf(:,2),lf(:,1));
+        lf_log = [ log10(r) .* sin(t) .* cos(f) ...
+                   log10(r) .* sin(t) .* sin(f) ...
+                   log10(r) .* cos(t)];
+    end
+
+    %= = From this line, these functions are part of the figure3d
+    %% ===== SET STANDARD VIEW =====
+    function SetStandardView(hFig, viewNames)
```
ftadel, line 566: I don't understand why you copy-pasted a function from figure_3d here.
If you need to call a function in a different file, use the syntax figure_3d('FunctionName', param1, param2...)

ftadel (changes requested):

toolbox/gui/view_leadfields.m
tmedani: ok I will correct it,

toolbox/gui/view_leadfields.m
tmedani: I think both are working, 
"Shift + Enter" and "Shift + L"

toolbox/gui/view_leadfields.m
tmedani: You're right, 
I did not manage to link the current function  with the figure_3d

toolbox/gui/view_leadfields.m
tmedani: ok, I will correct it

toolbox/gui/view_leadfields.m
ftadel: The block `otherwise` in the switch calls `KeyPressFcn_bak(hQuiver, keyEvent)`, which actually calls the original handler from figure_3d (`figure_3d/FigureKeyPressedCallback`). You don't have to do add anything so that figure_3d processes the orientation keys. 
If you have any trouble, add a breakpoint at the beginning of `figure_3d/FigureKeyPressedCallback` to check what is wrong.

In general, if you need to call a function in a different file, use the syntax `figure_3d('FunctionName', param1, param2...)`

toolbox/gui/view_leadfields.m
juangpc: Hi, This is just an open comment. Please don't modify your intentions. 
- the integration of the scaling function in the main flow. I personally wouldn't integrate it. Keeping simple, atomic, operations in a codebase, abstracted of everything else, as much as possible, I think makes the code easier to follow, debug and, eventually, reuse code, etc... This might be a bad example because of its simplicity, but still. 
- Totally agree with the renaming. Would suggest LogScaleLeadfield, but ScaleLeadfield is just fine.
Thanks for reading.

toolbox/gui/view_leadfields.m
tmedani: I did not manage to use figure_3d in this function, but if you think it's possible I will try to do it.

toolbox/gui/view_leadfields.m
tmedani: Hi, I have corrected the PR according to the discussion, and regarding this comment I changed only the name of the sub-function "ScaleLeadfield", I think also it's more convenient to have as a subfunction, but this is also my comment.
so, @ftadel if you don't like it I will change it :)

toolbox/gui/view_leadfields.m
tmedani: Ok, I got it, the problem was in the figure handles, 
image

Now it's working. 

Thanks :)

PR #346, opened 2020-10-21
Psd and fft amplitude normalization / scaling / units fix

Moo-Marc:
Fixed psd and fft so that power spectrum has correct amplitude in "physical" units (U^2/Hz).  Previous formula had a few errors that changed the global spectrum scale.  It also depended on the window size.
Both psd and fft processes now have an option to use "physical", "normalized" or "old" units.  The last two are greyed as probably rarely needed.  
Figure axis labels are also updated to reflect the actual units, with the "old" method now labeled U^2/"bin", although that's not exactly accurate.
Verified that the "physical" and "normalized" options give the correct amplitude with random N(0,1) noise @ 1kHz sampling rate.
See [this forum post](https://neuroimage.usc.edu/forums/t/psd-amplitude-factor/22258) for figures.

Although the fft process works, I'm not sure if it's only used to give a power spectrum?  If not, maybe this needs to be tested further.


ftadel:
> Although the fft process works, I'm not sure if it's only used to give a power spectrum?

The only use I know for process_fft is to compute a PSD from imported epochs : average(power(FFT(epochs)))

toolbox/timefreq/bst_timefreq.m
```diff
@@ -507,30 +508,19 @@
 
         % FFT: Matlab function fft
         case 'fft'
-            % Next power of 2 from length of signal
-            nTime = length(OPTIONS.TimeVector);
-            % NFFT = 2^nextpow2(nTime);    % Function fft() pads the signal with zeros before computing the FT
-            NFFT = nTime;                  % No zero-padding: Nfft = Ntime
-            % Positive frequency bins spanned by FFT
-            OPTIONS.Freqs = sfreq / 2 * linspace(0,1,NFFT/2+1);
+            % Use psd function, single window.
```
ftadel, line 511: Hmmm.... I see what you meant with your question about process_fft.
Before the two implementations where more different than what they are now.
It makes process_fft quite useless... but let's keep it for compatibility purposes.
It might be easier for some people to understand.

ftadel (approved):
Could you please edit the tutorial that goes with this process as well?
https://neuroimage.usc.edu/brainstorm/Tutorials/ArtifactsFilter#Evaluation_of_the_noise_level

1) Redo the screen captures (with the correct time windows, etc... exactly how it would look like for someone following the tutorials)
2) Add the description to your new option (no screen captures, as it would be confusing here - but you could add a link "(see github issue)" at the end of the description for option "Before Nov 2020", which includes the screen capture)

ftadel:
Restored previous behavior of bst_psd, regarding the DC offset: 
https://github.com/brainstorm-tools/brainstorm3/commit/b23e0fbafb7357d053c4d4598752be2f9a9fd08e#diff-f005e986e2828474e24c059bb16a01ef0466fdeb48dd7b45563b8082cd34bd6b
Otherwise, it is not possible to replicate previous results.

With the DC correction (subtracting the average of each window): 
image

Without DC correction:
image

@Moo-Marc 
Since we don't care about f=0 Hz,  let's make it as little disruptive as possible.
Do you have a strong preference for the version without this DC correction? 
If so: please disable it at least from the "old" option, so that we can reproduce old results.
Or maybe we should completely remove the first bin (0 Hz)?

@sbaillet @martcous @richardmleahy

ftadel:
@Moo-Marc When done with editing the tutorial, could you please also add an entry to the "What's new" page to indicate that something important changed with the PSD calculation?  (look for previous "Major bug fixes" for examples)
Here you could add two screen captures: before ("old")/after ("physical") to illustrate the change.

https://neuroimage.usc.edu/brainstorm/News

Moo-Marc:
Your figures gave me pause because actually, f=0 is already removed in the the spectrum figure
https://github.com/brainstorm-tools/brainstorm3/blob/b23e0fbafb7357d053c4d4598752be2f9a9fd08e/toolbox/gui/figure_spectrum.m#L1152
and also now in ScaleToFitY.

In a log-x plot x=0 is technically at infinity on the left and it's already removed.  With a rectangular window, it would have no effect on the rest of the spectrum, but in this case, the Hamming window with its wider peak also [affects the second bin](https://upload.wikimedia.org/wikipedia/commons/thumb/9/9b/DFT-even_Hann_window_%26_spectral_leakage.png/300px-DFT-even_Hann_window_%26_spectral_leakage.png) (same as Hann). 

Agreed for keeping it the same for the "old" option.  But since offset removal is an available process, I didn't see why we would want to do it quietly here.  Now with the above realization, I'm not sure.

And sure I'll edit the tutorial and What's new.

ftadel:
Can you bring this topic to the meeting next Wednesday?
It is coded in this way maybe for a reason. Richard, John and Dimitrios will tell you what they want in this function.

Following that, I'll let you open another PR to redo the changes that I undid :)
Thanks

Moo-Marc:
Sure, but it's fine to keep this way.  f=0 is already not displayed so might as well avoid its effect on the second freq bin.

PR #349, opened 2020-10-26
Add the seeg/ecog support with DUNEuro FEM 

tmedani:
Hi, 

This PR adds the possibility to use the DUNEuro computation for sEEG/ECoG from Brainstorm GUI. 
The main change is in the bst_duneuro function, where new parameters are included for the dn binaries.
 
The Duneuro computation is compared with the OpenMeeg in similar scenarios, and the results are satisfying. 

Best,
Takfarinas

### DIFF
--- a/toolbox/forward/bst_duneuro.m
+++ b/toolbox/forward/bst_duneuro.m
@@ -64,7 +64,7 @@ end
 if isEeg
     EegLoc = cat(2, cfg.Channel(cfg.iEeg).Loc);
 else
-EegLoc = [];
+    EegLoc = [];
 end
 % Get MEG positions/orientations
 if isMeg
@@ -553,7 +553,7 @@ Gain = NaN * zeros(length(cfg.Channel), 3 * length(cfg.GridLoc));
 if isMeg
     Gain(cfg.iMeg,:) = GainMeg; 
 end 
-if isEeg 
+if isEeg
     Gain(cfg.iEeg,:) = GainEeg; 
 end
 if isEcog 
--- a/toolbox/forward/bst_duneuro.m
+++ b/toolbox/forward/bst_duneuro.m
@@ -53,20 +53,18 @@ if (nnz([isEeg, isMeg, isEcog, isSeeg]) > 2) || ((nnz([isEeg, isMeg, isEcog, isS
     return;
 end
 % Get the modality
-if isEeg && isMeg
+if isMeg
+    dnModality = 'meg';  
+elseif ((isEeg || isEcog || isSeeg) && isMeg)
     dnModality = 'meeg';
-elseif isEeg
-    dnModality = 'eeg';
-elseif isMeg
-    dnModality = 'meg';
-elseif isEcog
-    dnModality = 'ecog';
-elseif isSeeg
-    dnModality = 'seeg';
+elseif  isEeg || isEcog || isSeeg
+    dnModality = 'eeg'; % from DUNEuro side, EEG, sEEG, ECOG  uses the same process
 end
 % Get EEG positions
 if isEeg
     EegLoc = cat(2, cfg.Channel(cfg.iEeg).Loc);
+else
+EegLoc = [];
 end
 % Get MEG positions/orientations
 if isMeg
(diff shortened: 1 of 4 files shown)


toolbox/forward/bst_duneuro.m
```diff
@@ -78,6 +76,27 @@
         end
     end
 end
+% TODO : use the real position of the sensors unstead of the integration
+% points ==> Too much memory for low accuracies advantages ==> discuss with
+% ftadel and create git discussion + PR
+
```
ftadel, line 82: This would be only for MEG. 
And would we really use this super heavy FEM for MEG routinely?...

If you want to replace the MEG integration points with the center of the sensor instead, you can use:
`chan_loc = figure_3d('GetChannelPositions', ChannelMat, iChan/Modality)`

toolbox/forward/bst_duneuro.m
```diff
@@ -53,20 +53,18 @@
...
-elseif isEeg
-    dnModality = 'eeg';
-elseif isMeg
-    dnModality = 'meg';
-elseif isEcog
-    dnModality = 'ecog';
-elseif isSeeg
-    dnModality = 'seeg';
+elseif  isEeg || isEcog || isSeeg
+    dnModality = 'eeg'; % from DUNEuro side, EEG, sEEG, ECOG  uses the same process
```
ftadel, line 61: Then why do you need to handle them in different ways?
If there are no specific parameters for ECOG or SEEG: then EEG+ECOG+SEEG could all be passed to bst_duneuro as `'EEG'`, with only one list of indices `cfg.iEeg`.

toolbox/forward/bst_duneuro.m
```diff
@@ -527,10 +553,16 @@
 if isMeg
     Gain(cfg.iMeg,:) = GainMeg; 
 end 
-if isEeg
+if isEeg 
     Gain(cfg.iEeg,:) = GainEeg; 
 end
-
+if isEcog 
+    Gain(cfg.iEcog,:) = GainEeg;
```
ftadel, line 560: This is weird... It means that you can't compute a mix of ECOG and SEEG?

toolbox/forward/panel_headmodel.m
```diff
@@ -104,6 +104,7 @@
         % Combobox
         jComboMethodECOG = gui_component('ComboBox', jPanelMethod, 'tab hfill', [], [], [], @UpdateComment, []);
         jComboMethodECOG.addItem(BstListItem('openmeeg', '', 'OpenMEEG BEM', []));
+        jComboMethodECOG.addItem(BstListItem('duneuro', '', 'DUNEuro FEM', []));
         jComboMethodECOG.setSelectedIndex(0);
```
ftadel, line 108: You should add the same new entries to process_headmodel as well.

ftadel:
It looks good, but the EEG+ECOG+SEEG part could be handled more gracefully.  Not being able to compute a ECOG+SEEG FEM model will be annoying.
Probably by just grouping them all as EEG at the beginning of the function?

I'll let you decide on this. 
Let me know when you want me to merge to this PR.

toolbox/forward/bst_duneuro.m
tmedani: Yes, only MEG, 

> And would we really use this super heavy FEM for MEG routinely?...

Currently yes, in the case of the finer head model, the DUNEuro MEG will crash since the cash memory will not support the size of the transfer matrix, and if we disable the cache the computation will be very long. 

Therefore, the probable solution is to avoid the integrations point for the FEM, and ask the DUNEuro to improve the memory management for the MEG. 

> If you want to replace the MEG integration points with the center of the sensor instead, you can use:
> chan_loc = figure_3d('GetChannelPositions', ChannelMat, iChan/Modality)
Thanks, this is more than useful, I have implemented a method that computes the mean of the 4 points :P

toolbox/forward/bst_duneuro.m
tmedani: ok, there is just one parameter to change on the in the minifile file, 
to avoid the projection of the electrode to the surface. I will correct it

toolbox/forward/bst_duneuro.m
tmedani: I need to check, I don't have a data set with mixed modalities to check, but I will simulate the behavior.

tmedani:
Hi @ftadel 

Here is the new version with corrections, I have also added some data to the file history that can be useful. 

I will work on a new PR for the MEG without the integration points. 

Thanks 
A+

Issue #166, opened 2019-01-29
Kilosort download fail?


mpompolas:
Very weird behaviour on a clean install of Kilosort.

The installation of Kilosort depends on 4 GitHub zips.
There are checks for each one if the file to be downloaded is present on the website.
However, although the urls are correct, the code fails to pull the files.

But if you try to download each file again and again, at some point it works, without any changes on the code!

In my understanding this is not a Brainstorm issue, but mostly Matlab-GitHub communication problem.
Do you guys have any insights on this?

martcous:
This is probably related to the problem other users have been having with the updater... It's becoming quite an annoying and unpredictable bug. What is your Matlab, Java and OS versions?

In the meantime, it is possible to download each GitHub zip yourself and put them under .brainstorm/kilosort_tmp/ for the installation to work properly.

mpompolas:
Yeah, once it's download it, it's done.
Already did it.

Matlab R2016b
Windows 7
Java Version 8 Update 161 (build 1.9.0_161-b12)

jschmiedt:
I'm having the same problem on Windows 10, Matlab R2018b with both Kilosort and UltraMegaSort:

```
***************************************************************************
** Error: [process_spikesorting_kilosort]  Electrophysiology > Unsupervised Spike Sorting > KiloSort
** Line 564:  process_spikesorting_kilosort>downloadAndInstallKiloSort (line 564)
** Impossible to download KiloSort.
** Error: File is empty.
** 
** Call stack:
** >process_spikesorting_kilosort.m>downloadAndInstallKiloSort at 564
** >process_spikesorting_kilosort.m>Run at 123
** >process_spikesorting_kilosort.m at 33
** >bst_process.m>Run at 229
** >bst_process.m at 36
** >panel_process1.m>RunProcess at 151
** >panel_process1.m at 26
** >gui_brainstorm.m>CreateWindow/ProcessRun_Callback at 759
** >bst_call.m at 28
** >gui_brainstorm.m>@(h,ev)bst_call(@ProcessRun_Callback) at 292
** 
** 
** File: monkeyB/@rawytu288c-01/data_0raw_ytu288c-01.mat
** 
*********
```
```
***************************************************************************
** Error: [process_spikesorting_ultramegasort2000]  Electrophysiology > Unsupervised Spike Sorting > UltraMegaSort2000
[...]

PR #357, opened 2020-11-07
NIRS -  nirstorm installation

Edouard2laire:
Hello François, 

I know we decided that having of simple way for user to install nirstorm was not a priority (#289) but as it was easy to do, and as I think nirstorm should be easily accessible, I did it anyway. :) 

Best regards, 
Edouard


ftadel:
It would be cleaner if everything was handled in the same function, instead of adding extra functions to gui_brainstorm.m (which is already dense enough).
Could you reformat your work to organize it in this way?
- Rename install_nirstorm.m into something with a prefix indicating which program provides it (eg. bst_install_nst.m, bst_install_nirstorm.m, nst_install.m...)
- Add a parameter install/uninstall
- Add a parameter isInteractive (to decide whehter you should display the questions and messages you put in gui_brainstorm)

Edouard2laire:
Hello François, 
Thanks for the comment. The code should be better now :) 

One point, you might want to discuss is the fact that I used unzip to download the archive from GitHub.  I tried to use the DownloadFile function from gui_brainstorm but it was not working (it was telling me that the file was empty). Note: this bug only happens in interactive mode, web save and . This might be related to #166 

A fix used in Brainentropy toolbox seems to call the Download file function multiple times : https://github.com/brainstorm-tools/brainstorm3/blob/master/external/brainentropy/be_install.m#L165-L180


Best regards, 
Edouard

ftadel:
> One point, you might want to discuss is the fact that I used unzip to download the archive from GitHub. I tried to use the DownloadFile function from gui_brainstorm but it was not working (it was telling me that the file was empty). 
> A fix used in Brainentropy toolbox seems to call the Download file function multiple times

This is indeed a problem that was reported multiple times: downloading from github with our downloader does not work correctly. Restarting the download multiple times indeed seemed to solved the problem.
The way we solved this problem in other places of the software was simply to never download directly from github, but copying the archives on the neuroimage server. This was also a way to be more robust to changes in external toolboxes for which the developers do not check the compatibility with Brainstorm before making changes (which won't happen with nirstorm :)

If unzip works directly, this is great!
It would be good to test it with older versions of Matlab though, this seems like a feature that could have changed over the years.

> Can you also confirm me that the nirstorm process have to be put in .brainstorm/process folder ?

Yes, otherwise it won't be made accessible as a plugin in Brainstorm:
https://neuroimage.usc.edu/brainstorm/Tutorials/TutUserProcess#Process_folders

ftadel:
> It would be good to test it with older versions of Matlab though, this seems like a feature that could have changed over the years.

It works after Matlab 2018a, but crashes with all the previous versions of Matlab.
Brainstorm works with any version >= 2009b.
If you keep this download mechanism, you need to add in your documentation that NIRSTORM requires Matlab 2018a.

ftadel:
@thomas-vincent FYI

Edouard2laire:
> > It would be good to test it with older versions of Matlab though, this seems like a feature that could have changed over the years.
> 
> It works after Matlab 2018a, but crashes with all the previous versions of Matlab.
> Brainstorm works with any version >= 2009b.
> If you keep this download mechanism, you need to add in your documentation that NIRSTORM requires Matlab 2018a.

thanks. Then I will change the code to use websave/urlwrite and then unzip. I'll make the PR latter this week.

ftadel:
> thanks. Then I will change the code to use websave/urlwrite and then unzip. I'll make the PR latter this week.

Better suggestion: use gui_brainstorm('DownloadFile'), like all the other downloaders. 
I added a switch in there: when downloading from github, it will use websave or urlwrite instead. Compatible with all Matlab versions. See this commit:
https://github.com/brainstorm-tools/brainstorm3/commit/7cd0f62f684f7aab1dc73e341cda68727b79af13

> edit1: I am using matlab2017b and it's working so maybe it changed with Matlab 2017

Maybe it depends on the OS... It crashed with 2017b on my computer.

> I don't know how known is NIRSTORM in the brainstorm community, but do you think it could be interesting to announce in the Brainstorm Fb page that user can now download nirstorm directly from Brainstorm ?

@sbaillet ?

PR #360, opened 2020-11-12
Spec units

Moo-Marc:
**1/f compensation change**: Something we can discuss.
It was compensating by f^2 for power spectrum.  While the power-law exponent of brain activity varies (and may depend on modality?) I believe in general it is the power spectrum that is said to be 1/f.  At least for my MEG data multiplying by f (right) seems better than f^2 (left):
 
image

**Relative power**: Not sure if data was ever saved in "magnitude" measure, but I changed it so that it is still relative to the (sqrt) total power and not to the sum of magnitudes.  This fits better with the process description and makes it simpler to describe the units.

**Units adjusted for normalized or relative psd**:  For 1/f normalization, the Hz units cancel, though not exactly in a meaningful way.  Still, I think it's a little useful to see the physical units.  For relative power, it shows "Relative power (no units)" or "Sqrt relative power (no units)" for magnitude.


ftadel:
These are all useful improvements and fixes, thanks!

However, the changes in process_tf_norm would break the compatibility with existing processing scripts (I think the other changes do not actually modify the values, but just the units in the display).

What about adding an extra option to the spectrum normalization process?
=> 'multiply' => Description "1/f^2  compensation" (greyed out, with an extra line mentioning that it was the old default in Brainstorm)
=> 'multiply2020' => Description "1/f compensation"
=> 'relative' => Unchanged

In you tests, please consider that is probably mostly used for time-frequency files, including directly by the time-frequency processes (Morlet and Hilbert). And this probably why this squaring issue was never reported...
Changing process_tf_norm changes in cascade the behavior of process_timefreq and process_hilbert. We need to pay attention to keeping the existing results stable, while adding this improvement for new processes. I think one possible solution would be to the following:
=> Adding new checkbox options "normalize2020" to process_timefreq and process_hilbert.
=> Keeping the old options "normalize" (so that the existing scripts keep on running in the same way), but adding the property .Hidden=1 (so that they are not visible in the interface).
=> When getting the options, the processes should use option 'normalize2020' if available, and otherwise 'normalize'

Does it make sense?

Moo-Marc:
> I think the other changes do not actually modify the values, but just the units in the display

Except if files are sometimes saved with measure 'magnitude'.  Is this the case?  If so, I guess I should also make a 'relative2020' option.

Moo-Marc:
I did as you suggested.  However, I then realized real tf decomposition did not have "multiply" or "relative" in their file names and no way to detect if they were normalized that I could see.  So I added a TfMat.Normalized field.  I didn't see a simple way to add this info in process_tf_norm since it's a "Filter" process.  Any suggestion?

I also noted that when changing which sensor is displayed in the TF figure or its associated spectrum figure, only the selected figure updates.  I thought they used to stay in sync.  Is this something I broke?

Thanks!

ftadel:
> Except if files are sometimes saved with measure 'magnitude'. Is this the case? (edit: I think I found the answer is yes. Testing.) If so, I guess I should also make a 'relative2020' option.

Yes, I tried to save everything with the same measure at some point, but it was a request from @dimitriospantazis to keep the values directly in the format they are used, eg. for computing statistics without having to re-write all the files.
Apparently the `abs(hilbert)` is more frequently used than the `power(hilbert)`, while the Morlet coefficients are more commonly reported squared (`power(morlet)`).

> So I added a TfMat.Normalized field.

Could you move it to TfMat.Options.Normalized? This is a better place to store "optional" parameters.

> I didn't see a simple way to add this info in process_tf_norm since it's a "Filter" process. Any suggestion?<

I could modify the code of bst_process so that TfMat.Options gets carried to the Run function, and the modifications saved after it returns. Would that help?

> I also noted that when changing which sensor is displayed in the TF figure or its associated spectrum figure, only the selected figure updates. I thought they used to stay in sync. Is this something I broke?

Yes, I guess by removing a call to SetSelectedRows in one of your commits.
This calls FireSelectedRowChanged(), which distributes the information to all the figures.
It would be good to restore this behavior, indeed.

Moo-Marc:
> Would that help?

Yes, that makes sense.  I think it's straightforward enough so I implemented it.

> I guess by removing a call to SetSelectedRows

I searched for a while, and then I checked on an older version and I believe it was actually always like this.  Maybe to allow displaying different spectra or TF figures for different sensors at the same time...  So I'll leave it like that.

ftadel:
These 2020 options makes things a big heavy, and difficult to review...
But it's much better with this fixed 1/f normalization anyway! Let's try it this way

ftadel:
Regarding the change of channels in multiple figures: it was working when changing the channel from the panel Display, but not when using the shortcuts arrowup/arrowdown. I changed this, so they use the same function.
https://github.com/brainstorm-tools/brainstorm3/commit/ae7af2b208d3579fcf6e4c08009c1995cd1ed08e
Could you please check if this working as expected?

I removed the `clf` call `figure_spectrum/UpdateFigurePlot`, as it was making the figure flicker.
But maybe this is going to cause some other unwanted side effects... please try to identify any unwanted behavior requiring this call to `clf` - If you find any, I'll try to find an in-between solution.

I noticed that there are some new issues when displaying one sensor of a timefreq file, ERSD-normalized, as time-series or spectrum. See the cuts in the lines and the Matlab warning below:
image

Moo-Marc:
Thanks Francois.  I'll look into it.  But of course negative data won't work with a log scale.  (It would stay positive if it was represented as a factor and not a % change, but it will be fine if displayed in linear scale.)  

On the other hand, I see my changes to units doesn't work for this case, and probably for TF stats which I think also exist.  I'll look into that too.

PR #362, opened 2020-11-19
More spectrum fixes

Moo-Marc:
Small fixes following the last PR: 
Avoid log scale when there is negative data.
Better y axis label and units for baseline normalized spectra.
image

Are there other TF.Measure we should treat differently?  Perhaps stats?  I didn't really find stat processes that apply to TF.


ftadel:
> Are there other TF.Measure we should treat differently? Perhaps stats? I didn't really find stat processes that apply to TF.

All the parametric and permutation tests can be executed on time-frequency maps and PSD.

Do you want to fix this PR for stat, or should I merge it and let your open another one for the stat?

ftadel:
While investigating problems related with the coherence computation (https://neuroimage.usc.edu/forums/t/difference-between-averaging-coherence-and-concatenate/22726/4), I realized that some recent options added for FOOOF or for your improved units and log PSD display are not compatible with connectivity results (coherence, PLV, Granger spectral...)

As it was including some bad blocking bugs, I patched it immediately (adding conflicts to this PR, sorry...)
https://github.com/brainstorm-tools/brainstorm3/commit/d94cce1d6d3ccbc41b9ad84275cca00ba59952ca

Can you please check again all your new display options with all possible sorts of the data types displayed with the Spectrum figures?
- Connectivity NxN between few sensors
- Connectivity 1xN between 1 sensor and all the others
- Connectivity NxN between scouts
- Connectivity 1xN between 1 scout and all the sources
- Connectivity with time (PLVt, time-resolved coherence) and no time (coherence)
- Connectivity results where all values zero, at least for one time point
- All the connectivity measures, just to make sure units and ranges of values are right
- Stat on PSD
- Stat on time-frequency (Hilbert/Morlet)
- Stat on connectivity
- All of the above both in 'columns' and 'butterfly' display mode
- Rechecking the FOOOF display again at the end (maybe with @lucwilson?)

Moo-Marc:
> Can you please check again all your new display options with all possible sorts of the data types displayed with the Spectrum figures?

Sure. Unfortunately, I'll have to read up because I haven't used these before in Bst. And when I put 2 psds in the process boxes, all the stats were greyed out.

And no worries waiting for all the Measures to be sorted out before merging.

ftadel:
> Unfortunately, I'll have to read up because I haven't used these before in Bst.  And when I put 2 psds in the process boxes, all the stats were greyed out.

For the coherence, just put any recordings in the Process1 box, and run the coherence process on 3-4 sensors.
For the stat: use Process2, select 3 PSD files on each side (must be different files, if you want some output), run process Test > Parametric test: independent. To see some values in the output, set the p-threshold closer to 1 in the Stat tab.
https://neuroimage.usc.edu/brainstorm/Tutorials/Statistics#Example_1:_Parametric_t-test_on_recordings

Moo-Marc:
Alright, ready for another review.

I tested stats and connectivity spectrum figures, butterfly and column display modes, disabled y-log scale where appropriate and fixed a lot of little bugs I had introduced, e.g. with zooming in column display.

One thing I changed: when doing a "power" statistic test on PSDs, I changed the test to sqrt the "power" measure instead of warning that the test is usually done on magnitudes.  It seemed to make more sense.  For t-tests and the like, it's not as obvious, so I left the warning as is.

toolbox/gui/figure_timeseries.m
```diff
@@ -1723,11 +1735,9 @@ function UpdateTimeSeriesFactor(hFig, changeFactor, isSave)
         % Butterfly: Zoom/unzoom vertically in the graph
         else
             FigureZoom(hFig, 'vertical', changeFactor);
-            % If auto-scale is disabled: Update DataMinMax to keep it hen scrolling
+            % If auto-scale is disabled: Update DataMinMax to keep it when scrolling
             if ~TsInfo.AutoScaleY
-                for iAxe = 1:length(GlobalData.DataSet(iDS).Figure(iFig).Handles)
```
ftadel, line 1728: iAxe is not defined if you remove this loop.
Please test and fix.

toolbox/gui/figure_spectrum.m
```diff
@@ -1,1745 +1,1839 @@
-function varargout = figure_spectrum( varargin )
```
ftadel, line 1: Why is this file entirely shown as edited?
This makes it impossible to review on the website now, and it will make it make hard to track the changes in the future in case of new bugs.
Can you please try to fix this?
Thanks

Moo-Marc: @martcous Any idea what's going on?
I tried verifying the line endings and I can't see any difference in any viewer (e.g. meld, notepad++).  I tried changing my local repo settings to only do "lf" endings, and recommit this file, but no change.  The culprit commit is d09de715, which is a merge of Bst master branch, so I'm really confused.  I guess I'll try undoing that commit somehow...

Moo-Marc:
Took me a long time, but figured out the line endings.  Ready for review.  But this highlights that we should have "normalized" (lf) line endings in the repo.  We can discuss at our next meeting.

toolbox/gui/figure_timeseries.m
Moo-Marc: (Fixed iAxe, thanks.) 
This line seems unnecessary to me, but I didn't test actual timeseries so I left it there.  The loop however was duplicated so I removed it.

ftadel:
Bit after bit, it became an impressive piece of work :)
It's difficult to test formally all the possible cases, so I trust you with the tests.
Let's keep on testing actively the spectrum figures whenever we use them.

Moo-Marc:
Thanks François! 

One thing I forgot to ask, not sure if you saw, is how to test for connectivity files.  You had checked if RowNames was non-empty, which I changed to RefRowNames since RowNames is used in other cases.  But I'm not sure if all connectivity would have RefRowNames, though it would make sense.

ftadel:
Yes, I think the connectivity files are all generated by `bst_connectivity`, and should all have a non-empty `RefRowNames`.
Do you have any example where it's not the case?

https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/connectivity/bst_connectivity.m#L743

Moo-Marc:
Sounds good.  Lets consider it settled then (until users start reporting more bugs) ;)

PR #364, opened 2020-11-25
move deeper in the GM element the dipoles which are outside the GM, as option3

tmedani:
Hi @ftadel ,
this PR is about the source space correction, and to complete/answer this issue https://github.com/brainstorm-tools/brainstorm3/issues/302#issuecomment-669706648 

The attached figures show how the dipoles are moved inside the nearest element, the code will first move the dipole inside the element (TempVertex ), end then compute its image (FinalDipole) to the centroid.
image
 
if the image(FinalDipole) is outside, then it keeps the TempVertex point, and if the TempVertex is outside it will use the previous algorithm (Option2 in the code).

With this version, we can ensure to fulfill "better" the "Venant condition" for the FEM dipole model.

### DIFF
--- a/toolbox/forward/bst_duneuro.m
+++ b/toolbox/forward/bst_duneuro.m
@@ -230,14 +230,50 @@ switch (cfg.HeadModelType)
                 
                 % OPTION #2: Gradually move the vertex towards the center of the centroid, until it is located inside the element
                 % Move the vertex towards the center until it is inside the element
-                nFix = 10;
-                for iFix = 1:nFix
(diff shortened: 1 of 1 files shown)


tmedani:
piece of code in order to check the patch
%%
close all;
figure;
plotmesh(ElemCenter(iTarget,:),'ko')
hold on
plotmesh(cfg.GridLoc(iVertOut(i),:) ,'bd')
hold on;
plotmesh(tmpVert,'rs')
hold on;
plotmesh(newPoint,'ks')
legend({'ElemCenter','OriginalDipole','TempVertex', 'FinalDipole'},'AutoUpdate', 'off')
hold on;
plotmesh(gmVert,targetFaces, 'facealpha',0.2)
hold on;
plotmesh(newPoint,'r.')

Issue #350, opened 2020-10-28
MEG computation with FEM : memory, time, sensors, integration points & computation per block 


tmedani:
Hi @ftadel @juangpc @jcmosher, et al,

I want just to discuss with you one of the issues that we observe when we compute the MEG forward with the FEM with high mesh densities.

With about 300 sensors and 8 integrations points per sensor and a high mesh density model (~>1 M nodes) 
the transfer matrix then is around ~2.5k xs 1M, which in most case crash the computation. 

here is a user experience feedback: https://neuroimage.usc.edu/forums/t/duneuro-fem-doesnt-finish/21977/24?u=tmedani

Here are the possible solutions : 

-1- Possible solution is to disable the use of the cache memory, however, in this case, the computation can be very long (~ 3 days) on a regular computer (see Yvonne response in the forum with the link above).  
This solution is already implemented (locally), but it's too long on a regular computer 

-2- Reduce the computation of the MEG by considering only the inner FEM layers, then we can reduce the number of the node by ~ 30-50%. This solution can work but also may crash in some cases if the memory is low, and can take a long time if we disable the cache memory (~ 1 day).  
This solution is also possible with the current implementation (master).

-3- Avoid the use of the integrations points and use one point per sensor, which will reduce the size of the sensor from ~ 2.5k to ~600 (gradiometers, as the ctf MEG) 
I have implemented this approach and the computation is much faster, in a basic example (10k nodes and 300 sensors),
 the time passes from ~8 minutes to ~ 2min, and the solution is still correct. 
@jcmosher do you have an estimation of the error on the solution (inverse problem) 
between these two configurations?   (sensors with or without the integration points) 

-4- Use all the integration points, then divides these ~2500 sensors into blocks of ~250, then call ~10 times the DUNEuro app, and finally concatenate the sub-results to build the full final LF. 
With this option, we can use the cache memory without error and the computation "can be faster". 
This last solution is not implemented yet.

I will try to implement it and test it, with this option we may need to read/write multiple times the i/o files.  

There is also the solution (-5-) but it requires further dev on the CPP duneuro code and this will be for the next version of the bst-duneuro, 

@juangpc what do you think? 

@ftadel we may need to add some options to the duneuro_panel to allows the activation of these options for advanced users.
and also decide if we will include all of these options or just select some of them
( also some options can be mixed and then can be faster). 

Thanks in advance for your help & feedback. 

A+, 
Takfarinas

[...]

PR #365, opened 2020-11-26
Solutions for the high density mesh for MEG with DUNEuro:  integration points and memory 

tmedani:
Hello all, 

@ftadel this PR is for the discussion started in this here: https://github.com/brainstorm-tools/brainstorm3/issues/350,

We have observed this issue from different testings and [users' experiences](https://neuroimage.usc.edu/forums/t/duneuro-fem-doesnt-finish/21977/23?u=tmedani).

The FEM computation of the MEG with high mesh density + the sensors integration points requires very high memory resources and time, which is not possible to perform in most user computers.

The following PR add a subpanel to the advanced Duneuro panel, and it proposes to the user the following options :
-1- Compute the MEG with or without the integration points 
-2- Disable/Enable the memory cache for the FEM computation 
-3- ToDo : compute per block of sensors (under investigation)

image

The first option is self-explained, 
The second, users can disable the usage of the memory cache, then could perform computation with the integration points with high models==> but it's time-consuming,

The third option is under investigation. 

Here is a view of the LF vectors when using or not the integration points:
image

And as @Moo-Marc mentioned, https://github.com/brainstorm-tools/brainstorm3/issues/350#issuecomment-717674309
the error can be in the range of 2% between the two models (in the spherical model).

Please let me know if there any change required in the code.
Best,
Takfarinas

### DIFF
--- a/toolbox/forward/bst_duneuro.m
+++ b/toolbox/forward/bst_duneuro.m
@@ -47,27 +47,26 @@ isEeg  = strcmpi(cfg.EEGMethod, 'duneuro')  && ~isempty(cfg.iEeg);
 isMeg  = strcmpi(cfg.MEGMethod, 'duneuro')  && ~isempty(cfg.iMeg);
 isEcog = strcmpi(cfg.ECOGMethod, 'duneuro') && ~isempty(cfg.iEcog);
 isSeeg = strcmpi(cfg.SEEGMethod, 'duneuro') && ~isempty(cfg.iSeeg);
-% Error: cannot combine modalities other than MEG+EEG
-if (nnz([isEeg, isMeg, isEcog, isSeeg]) > 2) || ((nnz([isEeg, isMeg, isEcog, isSeeg]) == 2) && (isEcog || isSeeg))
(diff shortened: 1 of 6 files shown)


ftadel:
> The third option is under investigation.

If you keep in it the interface, I'll wait until the code is here.
Otherwise, and if you want me to merge this now, please remove it from the interface.

ftadel:
I posted some code suggestions that I have edited directly in the github editor, and that I haven't tested.
Please excuse if there are any typos in there...

tmedani:
> > The third option is under investigation.
> 
> If you keep in it the interface, I'll wait until the code is here.
> Otherwise, and if you want me to merge this now, please remove it from the interface.

I will remove it from the GUI for this version,
and work on it, it may take some time before the next PR.

tmedani:
hi @ftadel, your suggestions, and comments are included in the last commit. 
the 3rd option will not be displayed on the gui,  
I will work on it and push as soon as I validate its results.

ftadel:
Should I merge this PR now?

tmedani:
Yes please, then it can be tested by the current users (Carsten's students). 
I'm on the third option that may take some time.

PR #359, opened 2020-11-10
Installation nirstorm - fix compatibility with Matlab < 2017b

Edouard2laire:
As discussed in #357, the installation script now uses gui_brainstorm('DownloadFile') to download nirstorm from GitHub. 

hs: is that normal that all not only my last commit appears on the commit list ?

PR #366, opened 2020-11-26
Update nirstorm installation script

Edouard2laire:
This PR is the continuation to our previous conversation here: #359 

I created the process to install nirstorm, and uploaded the GUI. I still need to figure out how to install the process and the functions in a different location.I am still working on it; but I have some idea. Unfortunately we might loose the ability to uninstall nirstorm as easily. Would that be a problem ? 

Also, i don't know what to do with the error management. (see ligne 99 of process_nst_install), if the script is launched in GUI, it's ok the installation function will display either the success message or the error, but if the script is executed in batch, how do i return the error ? 

Edouard


ftadel:
> I created the process to install nirstorm, and uploaded the GUI. I still need to figure out how to install the process and the functions in a different location.I am still working on it; but I have some idea. Unfortunately we might loose the ability to uninstall nirstorm as easily. Would that be a problem ?

Why is it a problem to move some files from $HOME/.brainstorm/nirstorm to $HOME/.brainstorm/processes?
And why is it complicated to remove files from $HOME/.brainstorm/processes?

> if the script is executed in batch, how do i return the error ?

Add an error to the execution report, and it would even be displayed in interactive mode.
Look for calls to `bst_report('Error', ...)`, you can find examples in almost all the processes. 

One additional comments:
=> Prefer using `bst_error(errMsg, errTitle, isStack)` instead of `java_dialog('error', ...)`

Edouard2laire:
> Why is it a problem to move some files from $HOME/.brainstorm/nirstorm to $HOME/.brainstorm/processes?
> And why is it complicated to remove files from $HOME/.brainstorm/processes?

yes you are right. I was over complicating the process because of the automatic generation of the uninstall script. 
I made the change in the nirstorm folder : https://github.com/Edouard2laire/nirstorm/tree/bst_install 
i also moved the mex files to the mex folder


Could you please help me test the installation script ? 

You can test by changing the following line in process_nst_install :
line 152:  ->     nistorm_folder = fullfile(tmp_folder,'nirstorm-bst_install'); 
line 155 -> nistorm_url = 'https://github.com/Edouard2laire/nirstorm/archive/bst_install.zip';

Thanks a lot :)

toolbox/process/functions/process_nst_install.m
```diff
@@ -0,0 +1,279 @@
...
+eval(macro_method);
+end
+
+
+function sProcess = GetDescription() %#ok<DEFNU>
+    % Description the process
+    sProcess.Comment     = getComment();
+    sProcess.Category    = 'Custom';
+    sProcess.SubGroup    = 'NIRS';
+    sProcess.Index       =  2000 * isnotdeployed(); % Hide in the deployed versiom
```
ftadel, line 32: Do you really need a separate function for this?
I tend to prefer avoiding adding functions that are called only once, but it does make the code slightly more readable. So entirely up to you, this is your code :)

toolbox/process/functions/process_nst_install.m
```diff
@@ -0,0 +1,279 @@
...
+    end
+    
+    sProcess.options.action.Type    = 'radio_linelabel';
+    sProcess.options.action.Value   = 'installation';
+    sProcess.options.action.Controller   = struct('install','installation','uninstall','uninstallation' );
+    
+    sProcess.options.text1.Comment = [ '<b>Download Nirstorm anatomical template:</b> '];    
+    sProcess.options.text1.Type='label';
+
+    sProcess.options.option_colin27_2019.Comment = 'Colin27 (2019)' ;
```
ftadel, line 55: Colin27?
We stopped recommending this template about 5 years ago...
The recommended Brainstorm template is the MNI ICBM152.

toolbox/process/functions/process_nst_install.m
```diff
@@ -0,0 +1,279 @@
...
+
+eval(macro_method);
+end
+
+
+function sProcess = GetDescription() %#ok<DEFNU>
+    % Description the process
+    sProcess.Comment     = getComment();
+    sProcess.Category    = 'Custom';
+    sProcess.SubGroup    = 'NIRS';
```
ftadel, line 31: Could you move ALL the nirstorm processes (process_nst_*) in the same "NIRS" group?
I found it quite unexpected to see new processes appear in the Events menu, which was too large to be displayed on my screen... Try to avoid having more than 10-15 processes per sub-menu, it makes it easier to navigate. (I know, I have some cleaning to do in the 'Events' menu... :)

`'NIRS - wip'` => `{'NIRS', 'Work in progress'}`
`'Events'` => `{'NIRS', 'Import'}`

toolbox/process/functions/process_nst_install.m
```diff
@@ -0,0 +1,279 @@
...
+    sProcess.options.option_colin27.Value   = 0;
+    sProcess.options.option_colin27.Class   = 'installation';
+        
+    sProcess.options.text2.Comment = [ '<b>Install additional features:</b> ', ...
+                                        '<p>wip : install work in progress features <br/>',...
+                                            'debug: install developping tool <br />', ...
+                                            'For more information about the work in progress features, please contact us on our Github page : <a href="https://github.com/Nirstorm/nirstorm">https://github.com/Nirstorm/nirstorm</a></p>'];
+    sProcess.options.text2.Type='label';
+    
+    sProcess.options.option_wip.Comment = 'Work in progress (wip)' ;
```
ftadel, line 76: What's the difference between WIP and DEBUG?

If you select "Debug" or "wip": does it change the basic processes, or is it only that your have more functions?

toolbox/process/functions/process_nst_install.m
```diff
@@ -0,0 +1,279 @@
...
+    end    
+end
+
+function status=status()
+    % process_nst_install('status')
+    % return if nirstorm is installed
+    status =  exist('uninstall_nirstorm')==2 && strcmp(fileparts(which('uninstall_nirstorm')),bst_get('UserProcessDir')); 
+end  
+
+function errMsg = install(extra,template,isInteractive,fromProcess)
```
ftadel, line 144: `isInteractive` and `fromProcess` are redundant. 
I think you could get along with only one parameter `isInteractive`.

toolbox/process/functions/process_nst_install.m
```diff
@@ -0,0 +1,279 @@
...
+    
+    nistorm_url = 'https://github.com/Nirstorm/nirstorm/archive/master.zip';
+    
+    errMsg = gui_brainstorm('DownloadFile',nistorm_url, tmp_file, 'Download NIRSTORM');
+    if ~isempty(errMsg)
+        errMsg = ['Impossible to download NIRSTORM: ' 10 errMsg];
+        return;
+    end    
+    
+    bst_progress('start', 'NIRSTORM', 'Installing NIRSTORM...');
```
ftadel, line 195: If you have a logo, you could add it to this progress bar (and to all the progress bars used in nirstorm).
Example:
```
bst_progress('setimage', 'logo_openmeeg.gif');
bst_progress('setlink', 'http://openmeeg.github.io');
```

The logo would go to the folder `brainstorm3/doc`.
Look at existing examples in this folder.

toolbox/process/functions/process_nst_install.m
```diff
@@ -0,0 +1,279 @@
...
+        delete(tmp_file);
+        return;
+    end
+    
+    % Install nistorm
+    addpath(nistorm_folder);
+    try 
+        nst_install(mode,extra);
+    catch ME
+        errMsg = 'Unable to install nirstorm';
```
ftadel, line 211: Maybe add the actual error message to the returned error, it could help for debugging.
`errMsg = ['Unable to install nirstorm: ' ME.message];`

toolbox/process/functions/process_nst_install.m
```diff
@@ -0,0 +1,279 @@
...
+        return;
+    end
+    
+    % Remove temporary files
+    if strcmp(mode,'copy')
+        rmpath(nistorm_folder)        
+        delete(tmp_file);
```
ftadel, line 218: Could you try using `file_delete` instead?
(both instead of `delete` and `rmdir`)

toolbox/process/functions/process_nst_install.m
```diff
@@ -0,0 +1,279 @@
...
+            end    
+            
+            template_bfn = [template_names{i_template} '.zip'];
+            template_tmp_fn = nst_request_files({{'template', template_bfn}}, ...
+                                        isInteractive && ~fromProcess, ...
+                                        nst_get_repository_url(), 18e6);
+                                    
+                                    
+            % Copy to .brainstorm/defaults/anatomy
+            copyfile(template_tmp_fn{1}, fullfile(bst_get('BrainstormUserDir'), 'defaults', 'anatomy'));
```
ftadel, line 241: `file_copy` offers more options than `copyfile`.
See the code for information

toolbox/process/functions/process_nst_install.m
```diff
@@ -0,0 +1,279 @@
...
+end
+
+function uninstall()
+    % process_nst_install('uninstall')
+    % 'uninstall': uninstall nirstorm   
+    bst_progress('start', 'NIRSTORM', 'Uninstalling NIRSTORM...');
+    
+    cur_dir=pwd;
+    cd(bst_get('UserProcessDir'));
+    uninstall_nirstorm();
```
ftadel, line 258: The install function is `nst_install`.
`uninstall_nirstorm` => `nst_uninstall` ?

ftadel (commented):
This looks good. 
I haven't looked at the actual nst_install and uninstall functions yet.
(which is maybe what you wanted me to review?)

I made a few comments here. 
Most of it is optional, except the SubGroup of the processes.

Let me know when I should try the nst_install function.

toolbox/process/functions/process_nst_install.m
Edouard2laire: We could maybe. I added the fromProcess parameter to be able to distinguish the case where extra is empty. Is it called from the menu and then we need to ask the user, or is it called from the process, and then we don't ask.

toolbox/process/functions/process_nst_install.m
Edouard2laire: i don't think we have a specific logo. I will ask

Edouard2laire:
Hello. 
I am working on your comments; Indeed, I would appreciate that you have a look at the PR in the nirstorm folder too if you have time :)

I noticed an issue with putting the nirstorm function in ./brainstorm/nirstorm is that when we start brainstorm, this folder is not added in the Matlab path.

toolbox/process/functions/process_nst_install.m
Edouard2laire: Unfortunately, I can't change it easily.   When installing nirstorm, the process created by Thomas automatically creates the matlab script to remove the file. This script is called "uninstall_nirstorm" (https://github.com/Nirstorm/nirstorm/blob/master/dist_tools/install_package.m#L78) 

Even if i change the name of the script now, we will still need to handle the case where the script was called "uninstall_nirstorm". 

There is a function in the nirstorm GitHub called nst_uninstall (https://github.com/Nirstorm/nirstorm/blob/master/nst_uninstall.m) but we can't use it as it is using the functions present in the dist_tools folder (https://github.com/Nirstorm/nirstorm/tree/master/dist_tools) that we don't install.

toolbox/process/functions/process_nst_install.m
Edouard2laire: The colin27 template contains two things: the segmentation, Voronoi map, and has the pre-computed fluences (http://thomasvincent.xyz/nst_data/fluence/) 

It could indeed be a good idea to also have a similar version for the ICBM152 template (that I also used in my own study)  The only problem is that I don't have access to Thomas website to upload the fluences and corresponding template...

toolbox/process/functions/process_nst_install.m
Edouard2laire: WIP contains the 'in developpement' process such as deconvolution (https://github.com/Nirstorm/nirstorm/blob/master/bst_plugin/GLM/process_nst_spree.m) and is adding new process to Brainstorm. Note that except for the deconvolution process that I am currently investigating, most of the processes in wip are just frozen and no work is being done on them. 

Debug is a little bit different; it is overwriting some of the main functions of Brainstorm (eg. bst_call, bst_error, and bst_process) with the aim of handling error differently that brainstorm does.  That's why I marked it as 'not recommended' as it affects mains functions of Brainstorm and is not up-to-date with the recent changes made to them (like the recent changes made by Marc Lalancette)

toolbox/process/functions/process_nst_install.m
Edouard2laire: Done here: https://github.com/Nirstorm/nirstorm/pull/151/commits/e5ecbbb6bff41cdc3574dd0baae051905ac23b95

toolbox/process/functions/process_nst_install.m
thomas-vincent: I think there are two different usages:
- the automatic installation via brainstorm
- the installation from source. In there nst_install and nst_uninstall are functions that are outside the toolbox, they only serve the purpose of installing.

All the changes made in these PR try to maintain both using the same functions but the processes are really different. So I think you should leave the original nst_install / nst_uninstall as tools to install from the source folder, and make another installation function to handle the installation from within brainstorm.

Else you will always stumble upon this kind of issues where you try to make the two scenarios work the same way while the mechanisms really differ. Of course you can factorize some sub-functions but not the top-level ones.

Also, I don't think you need to support an uninstall feature in the brainstorm installation case. If you put everything nirstorm-related in one place, you don't have to track files by files.
I did that because there can be many functions in .brainstorm/process that are not nirstorm-related. So it gave a way to clean just nirstorm and leave all the rest.

thomas-vincent:
> Hello.
> I am working on your comments; Indeed, I would appreciate that you have a look at the PR in the nirstorm folder too if you have time :)
> 
> I noticed an issue with putting the nirstorm function in ./brainstorm/nirstorm is that when we start brainstorm, this folder is not added in the Matlab path.

That's why I put everything in .brainstorm/process. But I guess you will find a better solution ;)

Edouard2laire:
> > I noticed an issue with putting the nirstorm function in ./brainstorm/nirstorm is that when we start brainstorm, this folder is not added in the Matlab path.
> 
> That's why I put everything in .brainstorm/process. But I guess you will find a better solution ;)

https://github.com/brainstorm-tools/brainstorm3/pull/366/commits/26375b16823c195db44ac31fe9ccbf126b2a7321
 seems to work :)

toolbox/process/functions/process_nst_install.m
ftadel: > Debug is a little bit different; it is overwriting some of the main functions of Brainstorm (eg. bst_call, bst_error, and bst_process) with the aim of handling error differently that brainstorm does

Can you describe this further?
You do not want to distribute anything that shadows other Brainstorm functions, this has no chance to work on the long run. Long-term maintenance must be considered in the development choices.
If there are things that are missing in the Brainstorm functions, we could think about extending them with additional parameters: please submit PRs for the core functions if this is essential to your work, but this "debug" option doesn't seem look like something we can keep.

toolbox/process/functions/process_nst_install.m
ftadel: Menu: isInteractive=1
Process: isInteractive = 0
What other case do you need to handle?

toolbox/process/functions/process_nst_install.m
ftadel: > the installation from source. In there nst_install and nst_uninstall are functions that are outside the toolbox, they only serve the purpose of installing.

Then `nst_install_bst`/`nst_uninstall_bst`?

> Even if i change the name of the script now, we will still need to handle the case where the script was called "uninstall_nirstorm".

```
if exist('uninstall_nirstorm')
    uninstall_nirstorm
else
    nst_uninstall_bst
end
```

ftadel:
> I noticed an issue with putting the nirstorm function in ./brainstorm/nirstorm is that when we start brainstorm, this folder is not added in the Matlab path.

The way we handle this in most of the external toolboxes: at the beginning of all the processes, an installation function is called, which checks that the toolbox is installed, checks that it is up-to-date (either by checking the download link, or by reading the version from a URL), and adds the folders to database. This ensures the independence of all the external modules and prevents possible incompatibilities between different toolboxes. Only the modules that are needed are added to the path. SPM and FieldTrip work this way as well.
Example: process_fem_mesh / InstallBrain2mesh

> Also, I don't think you need to support an uninstall feature in the brainstorm installation case. 

I agree: an uninstall function is not needed, at least:
- if all the nirstorm function are not permanently in the Matlab path (added on the fly when needed)
- nirstorm can be easily updated (automatically or manually)
If someone ever wants to get rid of it without having to delete the .brainstorm/processes and .brainstorm/nirstorm folders manually, there is always the possibility to reset Brainstorm from the preferences window (delete the entire .brainstorm folder)

> I am working on your comments; Indeed, I would appreciate that you have a look at the PR in the nirstorm folder too if you have time :)

I will do this when we're done with all the changes already discussed here.

toolbox/process/functions/process_nst_install.m
Edouard2laire: indeed. now I don't now haha

toolbox/process/functions/process_nst_install.m
Edouard2laire: > Can you describe this further?

i could but would need to dig into the code of those process more. I personally don't use this option.  I removed it from the installation process

toolbox/process/functions/process_nst_install.m
Edouard2laire: > Else you will always stumble upon this kind of issues where you try to make the two scenarios work the same way while the mechanisms really differ. Of course you can factorize some sub-functions but not the top-level ones.

I don't understand. In the end, the two scenarios are exactly the same. In both cases, we just want to transfer the file from the nirstorm folder to the .brainstorm folder. The only difference is that when installing from brainstorm, you need to download the nirstorm folder first. 

> Also, I don't think you need to support an uninstall feature in the brainstorm installation case. If you put everything nirstorm-related in one place, you don't have to track files by files.

no. The two installation scenarios are the same; so even in the brainstorm installation, we are still flooding the .brainstorm/process folder. So we need to keep track, so when uninstalling, the user is not losing the other process he put here.

ftadel: > no. The two installation scenarios are the same; so even in the brainstorm installation, we are still flooding the .brainstorm/process folder. So we need to keep track, so when uninstalling, the user is not losing the other process he put here.

Call them all .brainstorm/process/process_nst_*, and unistalling corresponds simply to:
```
file_delete({bst_fullfile(bst_get('BrainstormUserDir'), 'nirstorm'), bst_fullfile(bst_get('BrainstormUserDir'), 'process', 'process_nst_*.m')}, 0, 3)
```

Edouard2laire: If you ask me to do it; I'll make the change, but I think we are overcomplicating things here. 

There is a function, generated automatically at the time of the installation, that takes care of this for us.  Even if it's not the way I would have designed it like this, this is working well. So I don't see why we wan't to change it and possibly have to think about side effects (what is a user have some process he called "process_nst_blabla"; they will get deleted by your proposed code but not by the current solution. )

Edouard2laire:
> > I noticed an issue with putting the nirstorm function in ./brainstorm/nirstorm is that when we start brainstorm, this folder is not added in the Matlab path.
> 
> The way we handle this in most of the external toolboxes: at the beginning of all the processes, an installation function is called, which checks that the toolbox is installed, checks that it is up-to-date (either by checking the download link, or by reading the version from a URL), and adds the folders to database. This ensures the independence of all the external modules and prevents possible incompatibilities between different toolboxes. Only the modules that are needed are added to the path. SPM and FieldTrip work this way as well.
> Example: process_fem_mesh / InstallBrain2mesh
> 
> > Also, I don't think you need to support an uninstall feature in the brainstorm installation case.
> 
> I agree: an uninstall function is not needed, at least:
> 
> * if all the nirstorm function are not permanently in the Matlab path (added on the fly when needed)
> * nirstorm can be easily updated (automatically or manually)
>   If someone ever wants to get rid of it without having to delete the .brainstorm/processes and .brainstorm/nirstorm folders manually, there is always the possibility to reset Brainstorm from the preferences window (delete the entire .brainstorm folder)
> 
> > I am working on your comments; Indeed, I would appreciate that you have a look at the PR in the nirstorm folder too if you have time :)
> 
> I will do this when we're done with all the changes already discussed here.

I think we should have the uninstallation process and have the nirstorm function added permanently in the Matlab path. The reason for this is that some nirstorm process needs to have access to the function so the 'getDescription()' function works. For example all the processes doing optimal montage computation.

toolbox/process/functions/process_nst_install.m
ftadel: Anything that works for you is fine. As I told you, I don't even think an uninstall process is required :)

ftadel:
> I think we should have the uninstallation process and have the nirstorm function added permanently in the Matlab path. The reason for this is that some nirstorm process needs to have access to the function so the 'getDescription()' function

You can do what you want for the installer and the optional uninstaller, but this I would prefer to avoid.
1) Not overloading the path is the best way to have multiple toolboxes coexisting peacefully on users' computers.
2) Having the GetDescription function depende on other files than the process function itself is painful. Parsing the existing process folders should be made as fast as possible, and this implies not calling external functions (especially if there are I/O operations, test of the existence of files or variables...), because otherwise the Matlab inline compiler cannot optimize the execution of the GetDesription functions as much.
If you slow down the parsing of the process folder, you can dramatically decrease the speed of Brainstorm (it happened a few times before with other external contributions).

If this is something that is absolutely needed in terms of workflow, can you please describe why you need to call a function that is outside the process_nst_*.m function from GetDescription() in order to have it listed properly in the pipeline editor? Maybe we can optimize this in a different way.

toolbox/process/functions/process_nst_install.m
Edouard2laire: Nice, then i'll keep the first way :)

Edouard2laire:
Ok. I understand your point. For now, do you think it would be possible to merge this PR like this as it starting to become a huge PR in nirstorm and brainstom side?  Especially, it is the current behaviour that all the nirstorm are in the Matlab path for users using nirstorm. 

I can then open a new PR to discuss how to solve the issue with the process requiring access to the nirstorm function from GetDescription() ?

ftadel:
This doesn't work...
1) It downloads the master branch of the nirstorm into the tmp folder and unzips it
2) source_dir = the current path, not the install path (eg. for me: C:\Work\Dev\brainstorm3), therefore this line fails:
```
addpath(fullfile(source_dir, 'dist_tools'));
Warning: Name is nonexistent or not a directory: C:\Work\Dev\brainstorm3\dist_tools 
> In path (line 109)
In addpath (line 86)
In nst_install (line 66)
In process_nst_install>install (line 194)
In process_nst_install (line 23)
In gui_brainstorm>@(h,ev)process_nst_install('install',[],[],1) (line 185) 
```
3) install_package is not in the path and the process crashes, but does not close the progress bar, which stays on the screen forever... 
image

A few things to fix...

ftadel:
(sorry, I should have tested before merging... but you'll fix this super quickly, right? :-)

Edouard2laire:
yes, it's because the commit only work with the corresponding PR in nirstorm (https://github.com/Nirstorm/nirstorm/pull/151)

ftadel:
Works on my end too!
Any reason for which `dg_voronoi.*`, `spree_*.m` and uninstall_nirstorm.m` are still in `.brainstorm/process` instead of `.brainstorm/nirstorm`?

Edouard2laire:
it's because they are not respecting the naming scheme of nirstorm (process_nst_* for process and nst_* for functions). 

For spree_*, it will be corrected when I will merge the work I am currently doing on the process (https://github.com/Edouard2laire/nirstorm/tree/deconvolution).  I hope before the winter break.

uninstall_nirstorm, as discussed has to be in the .brainstorm folder as it is the uninstallation script. 

Concerning dg_vornoi, it's true that it should be renamed to nst_voronoi but the problem is that it would require to recompile the mex file or it seems that the compilation of this file is not that easy (https://github.com/Nirstorm/nirstorm/issues/128) This is on my todo list. But i don't think i will have time to look at this soon.

ftadel:
> Concerning dg_vornoi, it's true that it should be renamed to nst_voronoi but the problem is that it would require to recompile the mex file or it seems that the compilation of this file is not that easy (Nirstorm/nirstorm#128) 

Maybe you could simply handle it separately in the installer/uninstaller, if it requires to recompile complicated things?

PR #375, opened 2021-01-18
nirs - update snirf import

Edouard2laire:
This PR fixes some of the recent changes made recently to SNIRF and  is more robust to errors of formating in the data (transposition)

### DIFF
--- a/toolbox/io/in_data_snirf.m
+++ b/toolbox/io/in_data_snirf.m
@@ -82,6 +82,8 @@ for iChan = 1:nChannels
     if ~isempty(src_pos) && ~isempty(det_pos)
         ChannelMat.Channel(iChan).Loc(:,1) = src_pos(channel.sourceIndex, :);
         ChannelMat.Channel(iChan).Loc(:,2) = det_pos(channel.detectorIndex, :);
+        ChannelMat.Channel(iChan).Orient  = [];
+        ChannelMat.Channel(iChan).Comment = [];
     end
 end
 
--- a/toolbox/io/in_data_snirf.m
+++ b/toolbox/io/in_data_snirf.m
@@ -43,8 +43,19 @@ switch strtrim(str_remove_spec_chars(jnirs.nirs.metaDataTags.LengthUnit(:)'))
 end
 % Get 3D positions
 if all(isfield(jnirs.nirs.probe, {'sourcePos3D', 'detectorPos3D'})) && ~isempty(jnirs.nirs.probe.sourcePos3D) && ~isempty(jnirs.nirs.probe.detectorPos3D)
+    
     src_pos = jnirs.nirs.probe.sourcePos3D;
     det_pos = jnirs.nirs.probe.detectorPos3D;
+
+elseif all(isfield(jnirs.nirs.probe, {'sourcePos', 'detectorPos'})) && ~isempty(jnirs.nirs.probe.sourcePos) && ~isempty(jnirs.nirs.probe.detectorPos)
+    
+    src_pos = jnirs.nirs.probe.sourcePos;
+    det_pos = jnirs.nirs.probe.detectorPos;
+    % If src and det are 2D pos, then set z to 1 to avoid issue at (x=0,y=0,z=0)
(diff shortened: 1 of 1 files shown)


Edouard2laire:
Hi @ftadel, 

This PR is almost ready. I only have an issue: when importing a snirf file, it looks like it's working well but we have an issue with the channel file. When we double click on the montage, we can visualize it well, but if we right on it and then try to do 'Edit channel file' then we have the following error: 

```
Reference to non-existent field 'Orient'.

Error in panel_channel_editor>LoadChannelFile (line 360)
        nbOrient = max(cellfun(@(x)size(x,2), {ChannelMat.Channel.Orient}));
```
Or we don't have orientation for nirs

toolbox/io/in_data_snirf.m
```diff
@@ -45,15 +45,18 @@
 if all(isfield(jnirs.nirs.probe, {'sourcePos3D', 'detectorPos3D'})) && ~isempty(jnirs.nirs.probe.sourcePos3D) && ~isempty(jnirs.nirs.probe.detectorPos3D)
     src_pos = jnirs.nirs.probe.sourcePos3D;
     det_pos = jnirs.nirs.probe.detectorPos3D;
-elseif all(isfield(jnirs.nirs.probe, {'sourcePos', 'detectorPos'})) && ~isempty(jnirs.nirs.probe.sourcePos) && ~isempty(jnirs.nirs.probe.detectorPos)
```
ftadel, line 48: Is it the name of the fields that changed in the file format, or is this older code that was not appropriate?
In the first case: maybe you want to keep both versions, so that you can keep on reading older files?

In general, when updating the reader of a file format, always try to keep the backward compatibility.
If your changes won't prevent using Brainstorm with any existing files, please just ignore this comment.

toolbox/io/in_data_snirf.m
```diff
@@ -80,16 +82,25 @@
...
+        ChannelMat.Channel(iChan).Orient  = [];
+        ChannelMat.Channel(iChan).Comment = [];
     end
 end
+
 % AUX channels
+aux_index = false(1,nAux);
 for iAux = 1:nAux
+     if ~isempty(jnirs.nirs.data.dataTimeSeries) && ~isempty(jnirs.nirs.aux(iAux).dataTimeSeries) && size(jnirs.nirs.data.time,1) ~= size(jnirs.nirs.aux(iAux).time,1)
+        warning('TODO: Resample AUX channels to the NIRS sampling frequency.');
```
ftadel, line 94: If you want example code for this:
https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/io/in_fread_smr.m#L69

ftadel (commented):
Please let me know when I should merge this.

toolbox/io/in_data_snirf.m
Edouard2laire: The name got changed recently (https://github.com/fNIRS/snirf/commit/cde85639c105e4fc1d9309b2bbd3d65cad3ca343) . ok, I will restore backward compatibility.

toolbox/io/in_data_snirf.m
Edouard2laire: oh ok. thanks. I will add that and you will be able to push the PR. Unfortunately i don't have any file to test that. so will need to assume the code i correct :)

toolbox/io/in_data_snirf.m
ftadel: @fangq does it make sense to try and keep reading different versions of the snirf format?

toolbox/io/in_data_snirf.m
ftadel: If it doesn't correspond to anything you need, you can just ignore this comment for now, and we'll see if anybody complains some dat. It was just a pointer in case you were interested in implementing it soon.

Edouard2laire: ok. I'll leave the code in a comment so we have it if someone needs it :)

Edouard2laire:
Ok, you can push it like this :)

PR #378, opened 2021-01-31
Add fine-calibration and cross-talk options.

lkandemir:
Added the options to load fine-calibration and cross-talk files.


ftadel:
Now in the main Brainstorm distribution.
I can't test it because I don't have any dataset with these files, but since it's optional there is no risk of breaking anything :)
Thanks for the contribution! Looking forward to the next MNE-Python wrappers.

lkandemir:
Hi @ftadel, thank you. Yes, it is no problem if no files are loaded :) Looking forward to new contributions as well!

PR #386, opened 2021-03-03
Nirs - Detect satured channels from brainsight information

Edouard2laire:
Hello, 

This feature adds the reading of the saturation field created by Brainsight to detect saturation in the file. 

One question I have, is line 137, I associate the event with a specific channel (the one that is saturating) but how can I retrieve this information later in Brainstorm?

### DIFF
--- a/toolbox/io/in_fopen_nirs_brs.m
+++ b/toolbox/io/in_fopen_nirs_brs.m
@@ -122,6 +122,31 @@ if isfield(nirs,'CondNames')
 end
 
 
+% Detect saturation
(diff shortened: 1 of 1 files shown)


ftadel:
> One question I have, is line 137, I associate the event with a specific channel (the one that is saturating) but how can I retrieve this information later in Brainstorm?

It will change the way the event is rendered:
https://neuroimage.usc.edu/brainstorm/Tutorials/EventMarkers#Channel_events

What do you mean by "retreive"?

Btw, there is this other issue related with the import function: https://neuroimage.usc.edu/forums/t/sensor-cap-image/26058/4

Issue #388, opened 2021-03-13
NIRSTORM: Move to the plugin manager?


ftadel:
@multifunkim @thomas-vincent @Edouard2laire

I added a new solution for managing the packages downloaded by Brainstorm:
- Tutorial: https://neuroimage.usc.edu/brainstorm/Tutorials/Plugins
- Commit: https://github.com/brainstorm-tools/brainstorm3/commit/a8e0a54ef97f1658a2d268e211f586b87912f22a

It would be great if we could also move NIRSTORM to this framework, in order to have everything managed in a centralized way. This will also allow to have automated builds including all the dependencies (to keep on having NIRSTORM included in the compiled distribution of Brainstorm).

Who could help me with this task?
Maybe we could organize a short meeting to plan these changes?

ftadel:
I've just recompiled Brainstorm with the updated BrainEntropy and Nirstorm code.
Nirstorm should now work with both with the Matlab and the compiled distributions of Brainstorm.
Please test it and confirm that we can close this issue.

ftadel:
At this moment, Nirstorm is still excluded from the compiled version of Brainstorm.
It will be the case until this issue gets resolved: https://github.com/Nirstorm/nirstorm/issues/155

Let me know when I should try include to recompile Brainstorm with Nirstorm.

Edouard2laire:
Yes, i am still working on making the replacement process for the problematic ones. I should be able to merge them soon with the main branch of Nirstorm. I will let you know when its done :)

ftadel:
Nirstorm repo issues were closed: https://github.com/Nirstorm/nirstorm/issues/155

Nirstorm is restored as a compiled plugin: https://github.com/brainstorm-tools/brainstorm3/commit/0428eb6da54d4a4a7c575aaca4bb88d2fd5c0f5e

PR #391, opened 2021-03-13
Nirstorm plugin

Edouard2laire:
Hello Francois, 

This is the first try to add NIRSTORM in the plugin manager system. (#388) Firstly, I have to say that I love this new system! 

This PR is not ready to be pushed as I currently have one issue, and I might need your help. Right now, after installing NIRSTORM; if I launch BRainstorm; I got the following error telling me that some processes of NIRSTORM are invalid : 

> BST> Invalid plug-in function: "/Users/edelaire1/.brainstorm/plugins/NIRSTORM/nirstorm-master/bst_plugin/GLM/process_nst_glm_fit1.m"
> BST> Invalid plug-in function: "/Users/edelaire1/.brainstorm/plugins/NIRSTORM/nirstorm-master/bst_plugin/OM/process_nst_OM_from_cap.m"
> BST> Invalid plug-in function: "/Users/edelaire1/.brainstorm/plugins/NIRSTORM/nirstorm-master/bst_plugin/OM/process_nst_OM_from_cortex.m"
> BST> Invalid plug-in function: "/Users/edelaire1/.brainstorm/plugins/NIRSTORM/nirstorm-master/bst_plugin/OM/process_nst_OM_from_head.m"
> BST> Invalid plug-in function: "/Users/edelaire1/.brainstorm/plugins/NIRSTORM/nirstorm-master/bst_plugin/forward/process_nst_cpt_fluences_from_cortex.m"
> BST> Invalid plug-in function: "/Users/edelaire1/.brainstorm/plugins/NIRSTORM/nirstorm-master/bst_plugin/forward/process_nst_cpt_fluences_from_head.m"
> BST> Invalid plug-in function: "/Users/edelaire1/.brainstorm/plugins/NIRSTORM/nirstorm-master/bst_plugin/io/process_nst_import_head_model.m"
> BST> Invalid plug-in function: "/Users/edelaire1/.brainstorm/plugins/NIRSTORM/nirstorm-master/bst_plugin/mbll/process_nst_mbll.m"
> BST> Invalid plug-in function: "/Users/edelaire1/.brainstorm/plugins/NIRSTORM/nirstorm-master/bst_plugin/mbll/process_nst_mbll_dOD.m"

If I load NIRSTORM, then exit and launch brainstorm again, then I don't have those errors. This might be related to https://github.com/Nirstorm/nirstorm/issues/155 : some process calling external functions in the getDescription function. 

edit:   In GetVersionOnline, GlobalData.Program.isInternet is always 0 for me; even if I am connected to internet.  The reason is that the online connection is not checked if auto-update of brainstorm is deactivated. 

Édouard


ftadel:
Thanks for this incredible reactivity!

> This PR is not ready to be pushed as I currently have one issue, and I might need your help. Right now, after installing NIRSTORM; if I launch BRainstorm; I got the following error telling me that some processes of NIRSTORM are invalid :

Indeed, it calls a function that is not on the path or in the same subfolder, therefore it can't find it and return an error...
Three options:
- You put all the subfunctions needed for the GetDescription function in the same folder as the process
- I add a field AutoLoad, to load a plugin automatically at startup: https://github.com/brainstorm-tools/brainstorm3/commit/95b9f28a31b70afd7eb33926545952e377331ef7
You could make NIRSTORM a plugin loaded at startup, and probably solve this problem
- Make all the GetDescription functions static (but as there is a lot of shared code, this is not very optimal...)

> This might be related to Nirstorm/nirstorm#155 

This issue has to be fixed too!

>  In GetVersionOnline, GlobalData.Program.isInternet is always 0 for me; even if I am connected to internet. The reason is that the online connection is not checked if auto-update of brainstorm is deactivated.

Indeed, this is the expected behavior. Maybe the wording is not so good, but the idea is that when this option is unselected, there is no tentative to connect to the internet. This is very important for people who work in contexts where the internet is filtered by firewalls and proxies that take a lot of time to fail. If there is a 30s timeout to wait for for each plugin, this becomes unmanageable.
I agree that this is a slightly different topic than accepting automatic updates, though.

The logic is the following: when you uncheck "auto-updates", you don't get info related with online versions of anything (neither Brainstorm nor its plugins). If you don't get auto updates for Brainstorm, you don't get auto updates for NIRSTORM either. But you can still update manually to the latest version.

> This is the first try to add NIRSTORM in the plugin manager system. (#388) 

I fixed your fix by using `filesep` instead of {';',':'}, because this doesn't work in Windows (path contains both ':' and ';', the : being used for the drive letter...). I also fixed the missing '<'.
https://github.com/brainstorm-tools/brainstorm3/commit/14bfa15802f48ec298798a0dd87ebe91395c7332

So I messed up completely your PR, sorry.
Maybe it would be easier for you to drop it, fork again and submit a new one.

Edouard2laire:
Closing ths PR; #392 should be ready to be pushed. 

> The logic is the following: when you uncheck "auto-updates", you don't get info related with online versions of anything (neither Brainstorm nor its plugins). If you don't get auto updates for Brainstorm, you don't get auto updates for NIRSTORM either. But you can still update manually to the latest version.
ok; that makes sense. 

> This issue has to be fixed too!
yes, I am working on it. Just found that there is a process option for scout which is what the code you want us to remove is doing; so should be fixed soon. 

Édouard

PR #392, opened 2021-03-14
Plugin manager: add nirstorm support

Edouard2laire:
Continuation of #388. 

Note: I noticed a bug with the custom installation. It seems that, now, when using custom installation the process of nirstorm is no longer added to the process list. it seems that the plugin.mat is not created when using the custom installation.


ftadel:
Indeed, the plugin.mat is created only for the files that are managed automatically by Brainstorm. This menu "Custom installation" do not change anything to the selected processes. It is mostly for letting people use their personal installations of SPM and FieldTrip.

For the moment, the search for the plugins is only done during the initial installation of the process.
For "custom install", we could copy the  few lines that look for the `process_*.m` files in the subfolders somewhere else.

But could we consider this "custom install" a marginal case for the moment, and first make it work from the managed environment of Brainstorm? (since it would represent most of the NIRSTORM users I guess)

I'm sorry, the latest changes I pushed break your PR... (but allow Brainstorm to get compiled with all its plugins!)
Could you please fix the incompatibilities? (an easy solution is to fork again and copy again your changes to it) 
Thanks

ftadel:
(next time I promise I'll merge your PR before pushing all my commits :)

Edouard2laire:
> (next time I promise I'll merge your PR before pushing all my commits :)

done in #401

PR #393, opened 2021-03-19
PL2 fix to accommodate EEG channels

mpompolas:
Removed WB (Wideband) from the list of dataChannels - this might need to be revisited

image


ftadel:
Does this fix the issue that was reported on the forum?
https://neuroimage.usc.edu/forums/t/problems-of-import-a-pl2-file/26107

If so, could you post a message asking that user to update Brainstorm?
Thanks!

PR #397, opened 2021-03-31
synchronize two file

Edouard2laire:
Hello Francois, 

This PR is here to complement the current process called 'process_evt_transfer'. Here the process does not only transfer the event from file A to B but also synchronize them.  

This is useful when recording simultaneous EEG-NIRS where the EEG and nirs recording doesn't start and stop exactly at the same time. After this process, both files have the same time definition. (here we only keep the part of the signal where we have both recodings) 

BTW, it could be great to be able to visualize signals with different sampling frequencies. I think an easy way would be to upsample the signal with the lowest frequency just for visualization purposes. 

Edouard

### DIFF
--- a/toolbox/process/functions/process_synchronise_signals.m
+++ b/toolbox/process/functions/process_synchronise_signals.m
@@ -173,6 +173,6 @@ function OutputFiles = Run(sProcess, sInputsA, sInputsB)
     bst_save(OutputFile, sDataB_aligned, 'v7');
     % Register in database
     db_add_data(sInputsB.iStudy, OutputFile, sDataB_aligned);
-    OutputFiles{1} = OutputFile;
+    OutputFiles{2} = OutputFile;
     
 end    
--- a/toolbox/process/functions/process_synchronise_signals.m
+++ b/toolbox/process/functions/process_synchronise_signals.m
@@ -1,0 +1,178 @@
+function varargout = process_synchronise_signals(varargin)
+% process_synchronise_signals: Synchronise two signal based on common event
+
+% @=============================================================================
+% This function is part of the Brainstorm software:
+% https://neuroimage.usc.edu/brainstorm
+% 
+% Copyright (c)2000-2020 University of Southern California & McGill University
+% This software is distributed under the terms of the GNU General Public License
+% as published by the Free Software Foundation. Further details on the GPLv3
+% license can be found at http://www.gnu.org/copyleft/gpl.html.
+% 
+% FOR RESEARCH PURPOSES ONLY. THE SOFTWARE IS PROVIDED "AS IS," AND THE
+% UNIVERSITY OF SOUTHERN CALIFORNIA AND ITS COLLABORATORS DO NOT MAKE ANY
+% WARRANTY, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO WARRANTIES OF
+% MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE, NOR DO THEY ASSUME ANY
+% LIABILITY OR RESPONSIBILITY FOR THE USE OF THIS SOFTWARE.
+%
+% For more information type "brainstorm license" at command prompt.
+% =============================================================================@
+%
+% Authors: Edouard Delaire, 2021
+eval(macro_method);
+end
+
+%% ===== GET DESCRIPTION =====
+function sProcess = GetDescription()
+    % Description the process
+    sProcess.Comment     = 'Synchronyse files A and B';
+    sProcess.Category    = 'Custom';
+    sProcess.SubGroup    = 'Synchronize';
+    sProcess.Index       = 681;
+    % Definition of the input accepted by this process
+    sProcess.InputTypes  = {'data', 'raw'};
+    sProcess.OutputTypes = {'data', 'raw'};
+    sProcess.nInputs     = 2;
+    sProcess.nMinFiles   = 1;
+    
+    %Description of options
+    sProcess.options.inputs.Comment = ['For synchronization, please choose an<BR>event type which is available in both datasets.<BR><BR>'];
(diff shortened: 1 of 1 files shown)


mpompolas:
in_fread_tdt and in_fread_nwb deal with this exact problem:
Checking which sampling rate is the one to be used (usually the one for electrophysiological, not behavioral data), and then the rest of the signals are upsampled or downsampled to match that sampling rate (I've had behavioral data that were sampled higher than raw signals).
When some signals don't have values, they are filled with NaNs.
However, these importers deal with a single acquisition file.

For your case, a new importer would be needed that "concatenates" the signals from all acquisition systems and presents them with the same sampling rate. The segments before one of the acquisitions has started will be filled with NaNs for those channels.

In order to be mathematically accurate though, the time-constants of the internal filtering of each acquisition system needs to be taken into account in order to sync the signals.

ftadel:
I'm not sure I understand what this process does that cannot be done with basic functions of Brainstorm.
- Extracting blocks of recordings of the same dimensions is an Import operation (=epoching), based on either:
   - Event A + a fixed duration
   - Event AB: Extended event from event A to event B created with process "Events > Combine stim/response"
- Enforcing the same time vector for files with the same number of time samples is done with process "Standardize > Uniform epoch time".

Brainstorm can't handle signals with different sampling rates. If you need to import in the same files signals with different sampling rates, you need to upgrade your in_fopen_* function, as @mpompolas mentioned. 
In function `in_data_snirf.m`, there is a commented call to `interp1`: is this what you need?
You can search for `interp1` calls in other in_fopen* and in_data* functions for additional examples.

Edouard2laire:
Hello, @mpompolas and @ftadel,

I understand your points. However, I don't think that upsampling nirs signal is a good idea.  See there is a factor 100 between the sampling rate of nirs (10Hz) and the one of EEG (1khz) and it would not make really sense to have nirs signal at 1Khz. 

Especially, here I am dealing with all-night recording which already takes time to process at 10Hz. Additionally, I don't necessarily want to have EEG and nirs in the same file.. The reason is that the processing we are applying on the EEG and nirs are quite different. At some point, I hope that we would be able to analyze fMRI data in brainstorm; when doing EEG-fMRI, would you recommend to upsample to fMRI signal at 1kHz?

Having the file synchronized would be really helpful as it would allow our epileptologist to mark the specific time of interest in the EEG, send me the trigger file allowing me, because the file is synchronized, to directly import those events in the nirs file.  When now, I need to import them in the same EEG file, export them along with the synchronization trigger, run a separate script to calculate the timing of the events for the nirs, and then import this new event file in nirs signal. 

Note: I know that a process transfers event exists, but it would not work here as the process is automatically converting extended events into single events but even, then it's not very convenient to have to import the event in the EEG file to then transfer them to nirs. 

Edouard

ftadel:
> I don't think that upsampling nirs signal is a good idea.

No, it's not. And it won't be for fMRI either.
There is not necessarily much interest for displaying these signals simultaneously either.

> Having the file synchronized would be really helpful as it would allow our epileptologist to mark the specific time of interest in the EEG, send me the trigger file. And because the files are synchronized, I could directly import those events in the nirs file. 

This is something you can do with only manipulating the lists of events. Duplicating the entire recordings only for transferring a list of latencies seems very suboptimal. I'd recommend you think of another solution that does not require editing the files themselves. 

> Especially, here I am dealing with all-night recording which already takes time to process at 10Hz

Large files should be handled as continuous files, and not fully imported in the database.
Your process can't handle links to continuous files... therefore it does not seem that it addresses your problem correctly.

I apologize if there is something I don't get in your request. 
However, for the moment, I think what this process addresses is too project-specific to be distributed in Brainstorm. 
If this satisfies exactly the needs of your current work, you can simply distribute the process to your collaborators and ask them to copy it into their $HOME/.brainstorm/process folder.

You can also push it to this other repository:
https://github.com/brainstorm-tools/bst-users

Edouard2laire:
> This is something you can do with only manipulating the lists of events. Duplicating the entire recordings only for transferring a list of latencies seems very suboptimal. I'd recommend you think of another solution that does not require editing the files themselves.

Ok. I'll try to think about a btter solution :)


> 
> > Especially, here I am dealing with all-night recording which already takes time to process at 10Hz
> 
> Large files should be handled as continuous files, and not fully imported in the database.
> Your process can't handle links to continuous files... therefore it does not seem that it addresses your problem correctly.

Oh ok. I always thought it was better to import all the recordings in the database; Is there a way to have the file "physically" in the brainstorm database; but opened as continuous files? (useful to be able to transfer the database from one laptop to another) 

How could i change the script to make it work with continuous files? 

> I apologize if there is something I don't get in your request.

I still think it could be useful for different eeg-nirs projects. I recently some article using the eeg signal as regressors for GLM analysis in nirs (then the two continuous signal needs to be synchronized). But If you prefer I can put it in the Bst_user repo.

ftadel:
> Oh ok. I always thought it was better to import all the recordings in the database; Is there a way to have the file "physically" in the brainstorm database; but opened as continuous files? (useful to be able to transfer the database from one laptop to another)

ftadel:
> Oh ok. I always thought it was better to import all the recordings in the database; Is there a way to have the file "physically" in the brainstorm database; but opened as continuous files? (useful to be able to transfer the database from one laptop to another)

Right-click on the link > File > Copy to database.
Or menu File > Export protocol > Copy raw files to database.

> How could i change the script to make it work with continuous files?

Check the existing process_evt_*.m processes.

> I still think it could be useful for different eeg-nirs projects. I recently some article using the eeg signal as regressors for GLM analysis in nirs (then the two continuous signal needs to be synchronized). But If you prefer I can put it in the Bst_user repo.

I repeat my initial question: why can't this be done with a combination of import (epoching around the sync event - simple or extended) and enforcing the same Time vector for all the files (Standardize > Uniform epoch time)?

PR #398, opened 2021-04-02
bug correction, recompute the MEG location as a mean of the IP

tmedani:
Change the way to compute the position of the MEG sensors without the weights, in some case the weights of the gradiometers 
the error appears with the Neuromag gradiometer, the weights are not simply [0.5 0.5 -0.5 -0.5], but [29.76 29.76 -29.76 -29.76] ,
this correction o does not use the weights.

### DIFF
--- a/toolbox/forward/bst_duneuro.m
+++ b/toolbox/forward/bst_duneuro.m
@@ -91,15 +91,17 @@ if isMeg
         for iChan = 1 : length(cfg.iMeg)
             group = MegChannels(MegChannels(:,1) == iChan,:);
             groupPositive = group(group(:,end)>0,:);
-            groupNegative = group(group(:,end)<0,:);
+            groupNegative = group(group(:,end)<0,:);            
             if ~isempty(groupPositive)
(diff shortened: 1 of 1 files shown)


ftadel:
Oops, maybe I merged this a bit too fast.
Can you please have a look a the following function, that computes the sensor positions from the integration points in a reliable way for all the supported MEG systems:  `figure_3d('GetChannelPositions', ChannelMat, selChannels/Modality)`
https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/gui/figure_3d.m#L3370

tmedani:
> Oops, maybe I merged this a bit too fast.
> Can you please have a look a the following function, that computes the sensor positions from the integration points in a reliable way for all the supported MEG systems: `figure_3d('GetChannelPositions', ChannelMat, selChannels/Modality)`
> https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/gui/figure_3d.m#L3370

Hi Francois, 

Indeed this function computes the positions of the sensors, I have tried it, it's fine if all the sensors are Magnetometers. However, for the gradiometers, we need the two coils' positions with their weights in order to compute the final value of the GRAD.

PR #400, opened 2021-04-07
Fix ignored scouts in process_henv1n

rcassani:
Fix bug in [process_henv1n](../blob/master/toolbox/process/functions/process_henv1n.m) that ignored the selected scouts.
Bug mentioned [here](../issues/142#issuecomment-782818111) (in #142)


ftadel:
FYI: After pushing something, if you want to make the modification immediately accessible by a user, you need to force rebuilding the brainstorm package available on the website by calling this PHP script:
http://neuroimage.usc.edu/bst/update_brainstorm.php?g=p845Jui

If there is no emergency, then it will be rebuilt automatically some during the night (daily cron on neuroimage.usc.edu).

PR #401, opened 2021-04-08
Integration of NIRSTORM in Brainstorm plugin system

Edouard2laire:
> Indeed, the plugin.mat is created only for the files that are managed automatically by Brainstorm. This menu "Custom installation" do not change anything to the selected processes. It is mostly for letting people use their personal installations of SPM and FieldTrip.
> 
> For the moment, the search for the plugins is only done during the initial installation of the process.
> For "custom install", we could copy the few lines that look for the process_*.m files in the subfolders somewhere else.
> But could we consider this "custom install" a marginal case for the moment, and first make it work from the managed environment of Brainstorm? (since it would represent most of the NIRSTORM users I guess)

I think you can still create a folder in the .brainstorm folder and create the .mat file there. I think this use case is even more marginal than you estimated as I think only developers of nirstorm will use the custom installation. 

Also, I found a weird behavior since the plugin system is that when I make a change nirstorm processes, the changes seem to not be taken into account in brainstorm until I restart it.  

Additionally, i think that when updating a plugin, it would be great to be able to display the changelog to the user :) 

if the system is working well, do you think I can remove from the nirstorm Github, all the codes related to the previous installation scheme? Also, I think two or three minor processes of NIRSTORM are using iso2mesh. should I include iso2mesh as dependencies or just add the specific line to load iso2mesh in the specific processes that are using it?


Edouard2laire:
EDIT: wait before mergig this :)

Edouard2laire:
ok. should be good now. Before NIRSTORM was not properly uninstalled at the first lunch of brainstorm.

toolbox/core/bst_startup.m
```diff
@@ -584,8 +585,17 @@ function bst_startup(BrainstormHomeDir, GuiLevel, BrainstormDbDir)
             'Delete the old plugins listed below?' 10 ...
             sprintf(' - %s\n', OldPlugPath{:})], 'Plugin manager');
         if isConfirm
+            
             file_delete(OldPlugPath, 1, 3);
-        end
+            cur_dir=pwd;
```
ftadel, line 590: `1) Shouldn't `uninstall_nirstorm` be executed before deleting `OldPlugPath`?

2) Please move the folder change inside the test for `nirstorm`: no need to switch folder for the other plugins.

toolbox/core/bst_plugin.m
```diff
@@ -284,13 +284,19 @@
...
-%     PlugDesc(end).LoadFolders   = {'*'};
+    PlugDesc(end+1)             = GetStruct('NIRSTORM');
+    PlugDesc(end).Version       = 'github-master';
+    PlugDesc(end).AutoUpdate    = 0;
+    PlugDesc(end).AutoLoad      = 1;
+    PlugDesc(end).CompiledStatus = 2;
+    PlugDesc(end).URLzip        = 'https://github.com/Nirstorm/nirstorm/archive/master.zip';
+    PlugDesc(end).URLinfo       = 'https://github.com/Nirstorm/nirstorm';
+    PlugDesc(end).LoadFolders   = {'bst_plugin/core','bst_plugin/forward','bst_plugin/GLM', 'bst_plugin/inverse' , 'bst_plugin/io','bst_plugin/math' ,'bst_plugin/mbll' ,'bst_plugin/misc', 'bst_plugin/OM', 'bst_plugin/preprocessing', 'bst_plugin/ppl'};
+    PlugDesc(end).TestFile      = 'nst_install.m';
```
ftadel, line 295: Do you still need `nst_install.m`?
Otherwise, maybe it's not a good `TestFile`.

toolbox/core/bst_plugin.m
```diff
@@ -419,6 +425,12 @@ function Configure(PlugDesc)
                 disp(['BST> Checking latest online version for ' PlugName '...']);
                 str = bst_webread('http://neuroimage.usc.edu/bst/getversion_duneuro.php');
                 Version = str(1:6);
+           case 'NIRSTORM'
+                bst_progress('text', ['Checking latest online version for ' PlugName '...']);
+                disp(['BST> Checking latest online version for ' PlugName '...']);
+                str = bst_webread('https://raw.githubusercontent.com/Nirstorm/nirstorm/master/bst_plugin/VERSION');
+                Version = str(9:end);
+                URLzip = 'https://github.com/Nirstorm/nirstorm/archive/master.zip';
```
ftadel, line 433: You can remove this line, as the URL does not depend on the version.

ftadel (changes requested):
Looks good!
Just a few details to fix.

toolbox/core/bst_startup.m
Edouard2laire: 1. it doesn't matter. uninstall_nirstorm is deleting the file in the .process folder 
2. Done

toolbox/core/bst_plugin.m
Edouard2laire: Done

toolbox/core/bst_plugin.m
Edouard2laire: Done

ftadel:
I could not test the uninstall, as I already had removed it. 
But the new installer works well. Thanks!

I just removed the \n at the end of the version string, which was causing the install/update dialog boxes to display in a very ugly way.
https://github.com/brainstorm-tools/brainstorm3/commit/2281ba0545276c6d42242d9780bd75e9f67acb45#diff-dc8f8ced87ae44db6a025ad870e6b46c1c6569c2a857ae2003e2be9ba2ff0599R432

PR #402, opened 2021-04-12
report an error when unable to connect to brainstorm server

Edouard2laire:
This PR is here to display an error if, after the login, Brainstorm is unable to reach the Brainstorm server. 

I reset my brainstorm installation this morning to use another database, and after I was putting my login information, nothing was happening. The reason was that the send request was not working (see error below). I think it should be reported to the user. 

```
Error using matlab.net.http.RequestMessage/send (line 408)
Could not establish a secure connection to "neuroimage.usc.edu". The reason is "error:14090086:SSL
routines:ssl3_get_server_certificate:certificate verify failed". Check your certificate file
(/util/packages/matlab/R2017a/sys/certificates/ca/rootcerts.pem) for expired, missing or invalid certificates.

```

### DIFF
--- a/toolbox/gui/gui_brainstorm.m
+++ b/toolbox/gui/gui_brainstorm.m
@@ -1645,6 +1645,7 @@ function status = CloneProble(bstDir)
                 bst_error('Invalid username or password.', 'Brainstorm login', 0);
             end
         catch
+            bst_error('Could not establish a secure connection to Brainstorm server', 'Brainstorm login', 0);
             status = 0;
         end
         bst_progress('stop');


ftadel:
Is related to this (red message at the top of the homepage of the Brainstorm website)?
https://github.com/brainstorm-tools/brainstorm3/issues/308#issuecomment-646026943

You can skip this "login" by downloading Brainstorm from the download page, and updating it only with the Brainstorm update function.
Or simply delete the license and readme files from the brainstorm3 folder after cloning it from github.

Edouard2laire:
Yes this is the same error.  The authentification is still (and it's good :)) using https to transfer the login and passwords. 

But this might be an issue with my Matlab installation as I am also not able to download plugins from Github directly.

ftadel:
> But this might be an issue with my Matlab installation as I am also not able to download plugins from Github directly.

This login should block only the Brainstorm startup: the idea was to require users to register on the website, even when cloning the Github repository (the numbers of registered users and monthly downloads are important arguments for justifying our NIH funding).

What if you completely delete Brainstorm and get a new version from the website?
What version of Matlab are you using?

PR #403, opened 2021-04-15
Display nirs sensors on surface

Edouard2laire:
Hello Francois, 

This PR is here to allow users to display nirs channels on top of source maps. Before Brainstorm was only showing a dot in the middle of each channel. I think it would be better to actually display the sources and detectors. (see below)
Capture d’écran 2021-04-14 à 23 10 02

Note, i think i didn't broke any other visualization. But i will do more test tomorrow morning to be sure :)


Edouard2laire:
Edit; the only issue with the current PR is when right-clicking on an optode; it still says 'S1D1' : 

Capture d’écran 2021-04-14 à 23 56 08

toolbox/gui/figure_3d.m
```diff
@@ -2130,6 +2154,27 @@ function UpdateFigSelectedRows(iDS, iFig)
     DispChan = {GlobalData.DataSet(iDS).Channel(iDispChan).Name};
     % Remove spaces in channel names
     DispChan = cellfun(@(c)strrep(c,' ',''), DispChan, 'UniformOutput', 0);
+    
+    
+    if strcmpi(GlobalData.DataSet(iDS).Figure(iFig).Id.Type,'3DViz') && ...
```
ftadel, line 2159: Same comments exactly apply here.
Since you are using 3x the same block of code, maybe it would add to the readability to factorize it.

ftadel:
> 
> 
> Edit; the only issue with the current PR is when right-clicking on an optode; it still says 'S1D1' instead of S1 or D14 for exemple. I found that it was due to the value of GlobalData.DataViewer.SelectedRows but havn't found how to fix this.
> 
> Capture d’écran 2021-04-14 à 23 56 08

figure_3d.m, line 1489, beginning of DisplayFigurePopup.
It shows the name of the first selected data channel. 
You may change this part for NIRS as well.
(and maybe you might want to change it for figure_timeseries.m)

Edouard2laire:
Closing the PR as i don't have time to work on it right now.

PR #407, opened 2021-04-28
Fix foof compatibility with matlab < 2019b

Edouard2laire:
The syntax ''' szdim13 = size(A,[1 3])''' appeared in Matlab2019b


ftadel:
@lucwilson FYI

PR #408, opened 2021-05-12
New connectivity visualization tool

Yaqi97:
Engineering capstone design project by Helen Lin and Yaqi Li, under the supervision of Martin Cousineau and Dr. Sylvain Baillet. This project consisted in augmenting the connectivity visualization tool in order to remove OGL and Java dependencies. Connectivity graphs are now created as MATLAB objects. 

The new visualization tool is developed in figure_connect_viz.m and view_connect_viz.m. The old tool is still present in figure_connect.m and view_connect.m, and graphs can currently be opened with both tools.


sbaillet:
Fantastic work, Helen & Yaqi!

We’re very much grateful for your significant and impactful contributions.

Sylvain.

On May 11, 2021, at 9:04 PM, Yaqi Li ***@***.******@***.***>> wrote:


Engineering capstone design project by Helen Lin and Yaqi Li, under the supervision of Martin Cousineau and Dr. Sylvain Baillet. This project consisted in augmenting the connectivity visualization tool in order to remove OGL and Java dependencies. Connectivity graphs are now created as MATLAB objects.

The new visualization tool is developed in figure_connect_viz.m and view_connect_viz.m. The old tool is still present in figure_connect.m and view_connect.m, and graphs can currently be opened with both tools.

________________________________
You can view, comment on, or merge this pull request online at:

  https://github.com/brainstorm-tools/brainstorm3/pull/408

Commit Summary

  *   created base structure and files for new ConnectViz tool
  *   set up circular chord diagram base structure
  *   working prototype circ diagram w/o OGL
  *   remove 3DDisplay (unused) and add lobe grouping
  *   Updated FigureMouseWheelCallback and added comments
  *   Merge remote-tracking branch 'upstream/master' into master
  *   Merge remote-tracking branch 'upstream/master'
  *   completed grouping and node hierarchy, completed label visibility on/off controls
  *   Merge remote-tracking branch 'upstream/master'
  *   implement 1) zoom in/out, 2) camera translation, 3) label display toggle, 4) reset display on double-click, and 5) fix node display bug
  *   Merge remote-tracking branch 'upstream/master' into master
  *   Changed node markers for selected nodes
  *   updated display markers and added comments to adapt SetSelectedNodes
  *   fixed errors displaying original connectivity graphs
  *   connectivity links
  *   merge upstream
  *   Merge remote-tracking branch 'upstream/master' into master
  *   Fix merge on panel_display
  *   Merge pull request #1 from martcous/yaqi_fix
  *   Linked graph to colormap
  *   enabled ability to select link size
  *   Added link transparency
  *   Added ability to adjust brightness of colormap
  *   Update: colormap is done
  *   Implemented ability to display links from nodes selected by user
  *   Fixed Distance Filtering Slider
  *   Enabled functionalities for aggregating nodes
  *   Update Link Size increments
  *   bug fix: fix incorrect links displayed due to bug in UpdateColormap
  *   Update: keyboard events work with only one keypress
  *   Node and link selections as click events
  *   major cleanup of node+link structures, bug fixes
  *   Update node.m to clean-up
  *   SHIFT for multi-select nodes, Snapshot/SaveAs functionalities
  *   Add node and text size sliders in GUI
  *   Implemented directional links
  *   Update region toggling to match measure vs region link mode
  *   popup menu GUI updates
  *   Implemented Region Max/Mean
  *   Merge branch 'master' of https://github.com/Yaqi97/brainstorm3 into master
  *   Fixed region links going outside the unit circle
  *   Region Max/Mean is done
  *   Creation of node structs, integration and removal of node.m class
  *   Added Arrowheads for directional links
  *   Used CoCoMac to split overlapping bidirectional links
  *   Fixed bug with superposition of In/Out and bidirectional links
  *   RegionLinks updates
  *   Please enter the commit message for your changes. Lines starting
  *   Created elliptical arcs for overlapping links
  *   Fixed colormap labels when absolute value is selected
  *   Merge pull request #2 from martcous/colormap_fix
  *   Implemented doublearrows
  *   Removed 1 x N work and adde ButtonDownFcn for links
  *   Fixed douuble-arrowheads for directional links
  *   Added ability to increase link size with mouse click
  *   Fixed link selection mouse clicks for arrowheads
  *   Remove OGL remnants
  *   Updated visibility of clickable links and arrowheads
  *   Node size and label size implemented with 2 sliders
  *   Lobe full tag display and change in lobe position
  *   Clickable NodeLabels to select nodes
  *   Implemented Node size and label size in single slider
  *   Made arorwheads clickable
  *   Fixed bug with ClickedArrowIndex declaration
  *   Updated labels of nodes for clicked links
  *   Added text boxes for clicked links (commented out)
  *   Lobe label full and abbr options
  *   Update node selection display
  *   update region links to correspond with lobe indices
  *   Region/Hierarchy nodes update
  *   Fixed bug with mouse move when clicking links
  *   Fixed arrowhead scale for region links
  *   Removed individual sliders for node size and label size
  *   Optimized BuildLinks architecture and cleaned code
  *   Create both arrowhead in 1 function call
  *   Reduced runtime for arrowheads
  *   Fixed bug with extra arrowheads when moving slider
  *   Tested changes to colorbar (commented out)
  *   Added general fix for bug of double spline in non-directed
  *   Finalized arrowhead function
  *   Completed UpdateFigurePlot for spectral graphs
  *   Fixed various bugs for spectral graphs
  *   Update node, label, link size features
  *   Reset/default display options feature
  *   update logic for reset display
  *   Removed display statements, dead code and comments
  *   code optimization
  *   Fixed arrowhead visibility selection
  *   Removed binary display option from popup menu
  *   Fixed double arrowheads for Region toggling
  *   Fixed bug with double arrowheads when selecting labels
  *   update buildlinks code
  *   Fixed issue clicking links when more than one graph is open
  *   Merge branch 'master' of https://github.com/Yaqi97/brainstorm3
  *   clean code
  *   Auto-save user display preferences
  *   Fixed arrowheads visibility when toggling region links
  *   Removed profiler results folder
  *   Fix bug in hide region nodes
  *   clean code
  *   Remove unused params
  *   Move final connectivity .m files to /gui folder
  *   Update tree_callbacks.m
  *   add 'b' shortcut label to popup menu
  *   update central lobe labels
  *   Rebase with up to date Brainstorm distribution master branch
  *   Restored Display fibers option with new graph connectivity graph display
  *   Merge pull request #4 from martcous/yaqi_master
  *   Removed @todo comments

File Changes

  *   M toolbox/core/bst_colormaps.m<https://github.com/brainstorm-tools/brainstorm3/pull/408/files#diff-7879deadafe834b4ed2d0b4b5e0114d7ae5ea9b344ee2a8bfa952cd898737951> (13)
  *   M toolbox/core/bst_figures.m<https://github.com/brainstorm-tools/brainstorm3/pull/408/files#diff-f51a0d21cd22aa160169ba2b7a7a334fec29d3a486da3a3527691412a8efd0a6> (31)
  *   M toolbox/core/bst_get.m<https://github.com/brainstorm-tools/brainstorm3/pull/408/files#diff-913687240f342d0f3d6c9d93ee061d8f9e82be3518f00e069e394506d424251c> (20)
  *   M toolbox/core/bst_set.m<https://github.com/brainstorm-tools/brainstorm3/pull/408/files#diff-18d6a3b0cc97c4a4301938846020eb9ed674a906da113d2dfe4ad3752c1a3ae6> (2)
  *   M toolbox/gui/figure_3d.m<https://github.com/brainstorm-tools/brainstorm3/pull/408/files#diff-bb6d744e7d51b6955b619c888f623b91aa3d2d896fe02ecd616712da17b6f2cb> (11)
  *   M toolbox/gui/figure_connect.m<https://github.com/brainstorm-tools/brainstorm3/pull/408/files#diff-3e07339d98d155c42d28e3fc6c89f019d7a871146b02f802d5a6d8c0a530e9af> (10)
  *   A toolbox/gui/figure_connect_viz.m<https://github.com/brainstorm-tools/brainstorm3/pull/408/files#diff-7b95e1cf6e926acd0244539726ce5f13295625bead87a443ef7902aac9579fab> (4039)
  *   M toolbox/gui/gui_layout.m<https://github.com/brainstorm-tools/brainstorm3/pull/408/files#diff-9a4b5b01442145ce25f4ee81fe41bdfb849c0fd9d2be6a37b56c0eb1909dfff3> (10)
  *   M toolbox/gui/panel_display.m<https://github.com/brainstorm-tools/brainstorm3/pull/408/files#diff-50f1ba926a8e63e622287effe118d5d63e599a995a724d606b6f78acd2a7c544> (164)
  *   A toolbox/gui/view_connect_viz.m<https://github.com/brainstorm-tools/brainstorm3/pull/408/files#diff-412a29315d19eb17b871bb62b0f414b1aeae9abf08bfc51ba318ce0bdb2adbda> (239)
  *   M toolbox/tree/tree_callbacks.m<https://github.com/brainstorm-tools/brainstorm3/pull/408/files#diff-0f58c25199257166b30995be3635525263396f9a5b136697df058bb35a152c66> (13)

Patch Links:

  *   https://github.com/brainstorm-tools/brainstorm3/pull/408.patch
  *   https://github.com/brainstorm-tools/brainstorm3/pull/408.diff

martcous:
Yes, awesome work!! For more context, Helen and Yaqi have created a tutorial page that documents the new tool: https://neuroimage.usc.edu/brainstorm/Tutorials/ConnectivityGraph
I made sure the new tool is compatible with the fibers display.
Like they mentioned, this PR does not remove the old graph and all its Java dependencies. Perhaps we should have a decommission plan? I'm not sure if it's worth having both simultaneously for a while similarly to source models so users can still reproduce their previous figures. I guess they can always downgrade using Git or request an older version of the software...

ftadel:
It looks beautiful, and the tutorial with the animated views is very cool too!
Thank you very much for this great contribution to the software.

Before merging, there is a number of issues we need to address: 

1) **Lack of reactivity**: When clicking on a region node or changing the display properties, the update of the figure can be quite slow. We should add some element to indicate that the program is already processing the request, maybe calling `bst_progress`, or simply changing the mouse cursor. In the previous version, the updates were almost instantaneous.

1) **Slow display**: Function `BuildLinks` implements a nested `for` loop that executes millions of times the same lines of code. This has to be re-written in a lighter and/or vectorized way. At the moment, the figure creation (granger spectral 68x68) can take up to 20s, which is about 20x slower than the previous implementation.

1) **Graphs with two nodes**: Crashes => Should be handled gracefully

1) **Pan**: Moving within the figure with SHIFT is good. Adding the same convention as the other figures (middle click + move) would be even better.

1) **Node display**: It's difficult to click directly on a node when there are links displayed. Have you considered plotting the markers of the node on top of the links (higher Z coordinate)?

1) **Bidirectional link**: Displayed with two arrows in the same direction: it does not seem very readable, in the middle of the single arrows. Is this really what was expected?

1) **Handle buggy ConnectGraphOptions**: After some crashes (which I don't know how to reproduce but which happened already twice when I was doing these tests), all the fields of structure `ConnectGraphOptions` get set to `[]` and then saved to the user preferences. This create a permanent bug (NodeSize is [], which crashes at line 3401), that can be solved by the user only by resetting the Brainstorm environment.
=> A mechanism to fix buggy ConnectGraphOptions values should be added, maybe directly in bst_get.

1) **Link transparency**: Increasing the transparency doesn't make the other links visible below them, it just makes their color more dull. Any explanation for that? Should the various links be set to be on slightly different Z coordinates?
image

1) **Popup menu**: Reorganize the popup menu so that it looks like the menus for the other figures:
   - Move all the graph options to a submenu
   - Move the Snapshot menu after
   - Add the Figure menu (including the option to change the color freely - copy-paste from figure_3d)

1) Switching to regional mean: causes an important drop in values and therefore most links disappear. Should the threshold slider be updated when switching to this region mode?

1) **Backwards compatibility: ~**: The syntax `~` to skip input or output parameters make the parser of older Matlab versions, and should be avoided everywhere => Replace `~` with `tmp` for the output parameters that can't be simply ignored. Unused input parameters should have their real names.

1) **Backwards compatibility: bsxfun**: Only the most recent versions of Matlab allow the operator `.*` to behave like `bsxfun`, i.e. applying the same multiplication to all the elements in a given direction => The operator `.*` should be replaced with explicit calls to `bst_bsxfun` for all the multiplications that have inputs of different sizes (e.g. figure_connect_viz.m, line 2525)

1) **Older Matlab versions**: In general, all the tools must be tested with older Matlab versions, as we claim Brainstorm works starting with Matlab 2009b. If the connectivity graphs work only with features that appeared later, this minimum required release of Matlab must be documented clearly at the beginning of the tutorial. 

1) **Tutorial / Mindboggle atlas**: 
    a) This atlas was renamed recently to `DKT40`, which is the name of the parcellation (Mindboggle is the name of the dataset that was used initially, and was in practice not retained as the name to designate the atlas)
=> Screen captures and instructions in the tutorial must be update.
    b) Anatomical parcellations such as DK, DKT and Destrieux, are not good canditates for computing connectomes - but @sbaillet will probably decide to keep it as it is for not having to redo all the screen captures.
    c) "Unknown" regions don't look good in this graph - to be fixed in the parcellation import

> Like they mentioned, this PR does not remove the old graph and all its Java dependencies. Perhaps we should have a decommission plan?

Agreed. When this is all fixed, I also vote for removing all the old code, so we can completely get rid of JOGL.

helen-m-lin:
Hi @ftadel, thanks for all the feedback! I'll address each of these issues this week! I'll fix them as commits on our forked repo and then we can continue the pull request :)

martcous:
Thank you @helen-m-lin for looking into this, let me know if you need help.

@ftadel FYI, I can't test right now but it seems that the skipped parameter `~` is supported by Matlab 2009b so we don't have to avoid it anymore: https://www.mathworks.com/matlabcentral/answers/72537-what-does-a-tilde-inside-square-brackets-mean#answer_268214

helen-m-lin:
Hi @ftadel and @martcous, I've implemented several changes but I need some more clarification for a few:

> 2.	Slow display: Function `BuildLinks` implements a nested `for` loop that executes millions of times the same lines of code. This has to be re-written in a lighter and/or vectorized way. At the moment, the figure creation (granger spectral 68x68) can take up to 20s, which is about 20x slower than the previous implementation.

The issue is not with the loop or math computations (though I agree the nested loop can be improved). It is with Graphics handle creation in Matlab. Currently, each link is a **Line Object**, and additionally directional links have 2 extra **Patch** objects as arrowheads. Within `BuildLinks`, 60% of the time is used to create `Arrowhead` patch objects for directional links, while 30% of the time is used for the `Links(i) = line(x, y, ...)` call to create the **Line Objec**t handle.

- I initially tested if removing the `@ButtonDownFunction` of the link and arrows would help (though this would remove the clickability of links). In my example, I used the Granger graph which took 5.74s for `BuildLinks` initially, and **removing clickability only reduced the load time by 0.3s to 5.46s overall. Not enough.**
- To further reduce load time, ideally we would want to reduce the number of graphics objects created:
   - If we did not have arrowheads, 60% of the load time can be reduced. (But then we encounter the problem of how to display directionality. We had originally explored the use of markers at the start/end node on a line to indicate in/out.)
   - Alternatively, multiple links can be combined together as 1 graphics handle, but this does not allow for fine grain differentiation of different colours, clickability of links, and filtering based on grouping/node.
   - **Lastly, another possibility is displaying fewer links on start-up** (currently the max is 5000). For example, we can create the top 500 links on start-up, then once the graph is loaded, we can create the remaining 4500 links in the background. This may be the best option.
      - Display only 500 links on start: reduced load time from 5.74s to 2.624s.
      - Display 100 on start: reduced load time from 5.74s to 0.74s.
  
If this method is feasible, we can look into a speedup this way. This upper-bound of e.g. 500 or 100 links on start up ensures that all graphs load approximately within the same time (1-3s). What do we think? I think this may help otherwise we can only keep what we have now.

>6.	Bidirectional link: Displayed with two arrows in the same direction: it does not seem very readable, in the middle of the single arrows. Is this really what was expected?

Yes, this was what was decided previously, but perhaps instead of two arrows pointing in the same direction when the link is bidirectional, we can have two arrows pointing in opposite directions (e.g. --<-->--). What do we think?

helen-m-lin:
These suggestions have been fixed:
>4.	Pan: Moving within the figure with SHIFT is good. Adding the same convention as the other figures (middle click + move) would be even better.

Fixed. I compared with `figure_3D` and it seems the other figures allow panning with middle click and SHIFT + move, so I kept the original SHIFT+move ability too.

>5.	Node display: It's difficult to click directly on a node when there are links displayed. Have you considered plotting the markers of the node on top of the links (higher Z coordinate)?

Great idea! Fixed. Now the nodes are always on top, and we also had the clickable labels too in case of crowding.

>8.	Link transparency: Increasing the transparency doesn't make the other links visible below them, it just makes their color more dull. Any explanation for that? Should the various links be set to be on slightly different Z coordinates?

No change. We had encountered this issue and decided to keep the link transparency only referring to the transparency value of the colour property of the links. It does not implement colour blending or making links underneath visible.

>10.	Switching to regional mean: causes an important drop in values and therefore most links disappear. Should the threshold slider be updated when switching to this region mode?

Discussed previously with Martin + Sylvain, we had decided not to implement this in case users wanted to toggle back and forth between link types (mean/max/all) with their custom set thresholds.

>11.	Backwards compatibility: skipped parameter: The syntax ~ to skip input or output parameters make the parser of older Matlab versions, and should be avoided everywhere => Replace ~ with tmp for the output parameters that can't be simply ignored. Unused input parameters should have their real names.

Fixed. In my stashed changes for now since Martin suggested we may not need to change this? 

>12.	Backwards compatibility: `bsxfun`: Only the most recent versions of Matlab allow the operator `.*` to behave like `bsxfun`, i.e. applying the same multiplication to all the elements in a given direction => The operator `.*` should be replaced with explicit calls to bst_bsxfun for all the multiplications that have inputs of different sizes (e.g. figure_connect_viz.m, line 2525)

Fixed.

-------------------------------
Lastly, I'll work on your points 1, 3, 7, 9, 13 and 14 and will report back later. Thanks again Francois for suggesting the changes. Let me know if there's anything else or if you want more clarification on Slack! I'll try to change what I can this week

Yaqi97:
Thanks for taking the time to review our work! A few questions regarding the points you've made:

For #3: By "Graphs with two nodes", are you referring to only selecting 2 specific scouts when generating the graph?

For #6: You mention that the double-arrowheads are "not very readable, in the middle of the single arrows". Could you clarify what you mean by this? Do you find it visually difficult to distinguish bidirectional links from the other links? What do you suggest in that case? 
Context: Throughout our project, we had made the design decision to use arrowheads to distinguish bidirectional links from other In or Out only links, rather than the default green color that was used in the previous tool. We agreed that it was important for the color of a link to reflect its connectivity score through the colormap (and not through a default green color for all bidirectional links). 

For #9: The submenus in our popup menu (in particular the Graph options and display options) were re-organized in a way that we thought would be more user-friendly than the design seen in the old tool. We had made the decision to move the graph options to a submenu, and to keep snapshot options in the main menu as it only includes two options. Is the old design more user-friendly from your perspective, or perhaps could some of our submenus remain in their current state?

For #14: We merged the changes from the main distribution into our own repository on May 9th, and are still seeing the Mindboggle atlas as an option when generating connectivity graphs in the Process box. Where/when was it updated to `DKT40`? 

Moreover, could you clarify what you mean by ""Unknown" regions don't look good in this graph - to be fixed in the parcellation import"? Do you mean that the name of the lobe itself in the atlas (Unknown) needs to be updated, or the grouping of the nodes needs to be updated? What is the terminology that should be used? Note that we had simply imported the node groupings and names from the old tool.

ftadel:
Thank you all for your feedback and for the fixes!

@martcous 

> @ftadel FYI, I can't test right now but it seems that the skipped parameter ~ is supported by Matlab 2009b so we don't have to avoid it anymore: https://www.mathworks.com/matlabcentral/answers/72537-what-does-a-tilde-inside-square-brackets-mean#answer_268214

There were still issues with more recent versions of the compiler, but not one will ever compile Brainstorm with 10yr-old versions of Matlab. You're right, this is something from the past, I'll stop bothering people with it :-)


@helen-m-lin 

> If we did not have arrowheads, 60% of the load time can be reduced.

Maybe all the arrow heads could be created as a single patch object with the lower-level function `patch()`?
This would require a lot more work in terms of creating the graphical object, but would be must faster.


> Alternatively, multiple links can be combined together as 1 graphics handle, but this does not allow for fine grain differentiation of different colours, clickability of links, and filtering based on grouping/node.

It would be possible to have all the links inside only one `patch` object.
You can set color and transparency independently for each face or for each vertex, depending on what's the most convenient. This requires some more coding to assemble all the vertices/faces together correctly. Filtering out a link is the equivalent to making the link fully transparent (alpha=0). Example: 
https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/gui/figure_3d.m#L2850

Clickability requires to store an extra matrix of correspondence vertex-link, saved as `UserData` in the graphics handle for instance, and then to grab the click with `select3d` from the generic clicking callback of `figure_3d`. Example:
https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/gui/figure_3d.m#L683


> Yes, this was what was decided previously, but perhaps instead of two arrows pointing in the same direction when the link is bidirectional, we can have two arrows pointing in opposite directions (e.g. --<-->--). What do we think?

That seems more intuitive to me, but maybe I'm not up-to-date with the recent graphics conventions (you saw I'm still stuck with Matlab 2009b :-) 
@sbaillet What do you think?


> Fixed. I compared with figure_3D and it seems the other figures allow panning with middle click and SHIFT + move, so I kept the original SHIFT+move ability too.

This is what I meant indeed, thanks!

> No change. We had encountered this issue and decided to keep the link transparency only referring to the transparency value of the colour property of the links. It does not implement colour blending or making links underneath visible.

Then it's probably useless. I don't know why we would keep an option that only makes the image darker...
Could you remove the option from the menu?
(maybe you can keep in the code, in case this gets fixed in the future versions of Matlab...)

> Discussed previously with Martin + Sylvain, we had decided not to implement this in case users wanted to toggle back and forth between link types (mean/max/all) with their custom set thresholds.

Makes sense.


@Yaqi97

> For 3: By "Graphs with two nodes", are you referring to only selecting 2 specific scouts when generating the graph?

Yes. This is absolutely useless in terms of graph display, but it could make sense to compute these files, for example to compare the intensity of a connection between two regions across different experimental conditions or groups of subjects.
If the files can be in the database, we need to handle their display.
If the display is impossible or does not make sense, it should return a user-readable error instead of simply crashing.

> We had made the decision to move the graph options to a submenu, and to keep snapshot options in the main menu as it only includes two options. Is the old design more user-friendly from your perspective, or perhaps could some of our submenus remain in their current state?

I'd like the popup menu to be similar with the one of all the other figures:
- Colormaps
- Figure-specific submenus: Display options, Graph options
- Snapshots
- Figure (which was not present before, as it was not a Matlab figure - now it is, this menu can be added)
I think having all the graph configuration menus in the same submenu helps understanding that they all relate to the interpretation of the graph.

> For 14: We merged the changes from the main distribution into our own repository on May 9th, and are still seeing the Mindboggle atlas as an option when generating connectivity graphs in the Process box. Where/when was it updated to DKT40?

https://github.com/brainstorm-tools/brainstorm3/commit/9ecf22af97238225f5d017000ebace6aedda8f8e
You need to import again the anatomy.

> Moreover, could you clarify what you mean by ""Unknown" regions don't look good in this graph - to be fixed in the parcellation import"?

This is a note to myself, you can ignore this comment.
I simply wrote all my comments in that review, not knowing who would work on what.


Thank you all again for this amazing work!

Yaqi97:
Thank you for the detailed response, @ftadel ! A quick note regarding the design of double-arrowheads:

The reason why we made the 2 of them point in the same direction was because 2 separate bidirectional links are still drawn between 2 specific nodes, since a link is filtered as "bidirectional" whenever both the link from A -> B and the one from B -> A are both visible in the graph (i.e., when they both satisfy thresholds and conditions selected by the user). Example:
double_arrowheads

Hence, if we make the 2 arrows point in opposite directions, we would not be able to distinguish these 2 links, which may have different connectivity scores in general. Moreover, the 2 arrowheads are angled differently to follow the curve of the link; so, having them point like this: <> would not be visually appealing. @sbaillet We had a discussion regarding this, is there anything else you suggest in this case?

I will indeed look into creating both arrowheads into one single patch object over the next week. @ftadel Do you know if it is possible to have different `Visibility` properties for each face/vertex? So far, we had them as 2 separate patch objects to allow us to toggle their visibility separately (i.e., toggle between having the second arrow visible or not, depending on whether the bidirectional option is selected).

sbaillet:
Thank you all for your hard work on this.

I recommend we keep the double arrows. Now, there might be ways to optimize the execution time of the drawing by duplicating the same patch object and changing its location/position properties, rather than redrawing it from scratch for every instance.

Every Matlab graphical object has properties that can be set, including things like color, transparency and X, Y, Z coordinates etc. The execution of property changes is usually very fast.

You may have looked into this already?

Best,

Sylvain.

On May 20, 2021, at 7:54 AM, Yaqi Li ***@***.******@***.***>> wrote:



Thank you for the detailed response, @ftadel<https://github.com/ftadel> ! A quick note regarding the design of double-arrowheads:

The reason why we made the 2 of them point in the same direction was because 2 separate bidirectional links are still drawn between 2 specific nodes, since a link is filtered as "bidirectional" whenever both the link from A -> B and the one from B -> A are both visible in the graph (i.e., when they both satisfy thresholds and conditions selected by the user). Example:
[double_arrowheads]<https://user-images.githubusercontent.com/55117412/118973409-a529f280-b93f-11eb-8c28-7c5b951e5a5f.PNG>

Hence, if we make the 2 arrows point in opposite directions, we would not be able to distinguish these 2 links, which may have different connectivity scores in general. Moreover, the 2 arrowheads are angled differently to follow the curve of the link; so, having them point like this: <> would not be visually appealing. @sbaillet<https://github.com/sbaillet> We had a discussion regarding this, is there anything else you suggest in this case?

I will indeed look into creating both arrowheads into one single patch object over the next week. @ftadel<https://github.com/ftadel> Do you know if it is possible to have different Visibility properties for each face/vertex? So far, we had them as 2 separate patch objects to allow us to toggle their visibility separately (i.e., toggle between having the second arrow visible or not, depending on whether the bidirectional option is selected).

ftadel:
> Hence, if we make the 2 arrows point in opposite directions for both links, we would not be able to distinguish these 2 links, which may have different connectivity scores in general.

This makes sense.
If the two ways are represented anyway, why do you need the two arrows then?

> I will indeed look into creating both arrowheads into one single patch object over the next week. @ftadel Do you know if it is possible to have different Visibility properties for each face/vertex? 

You need to use the `FaceAlpha` or `FaceVertexAlphaData` properties and set the visibility per face or per vertex, not per object. More painful to code, but much more efficient!
https://www.mathworks.com/help/matlab/ref/matlab.graphics.primitive.patch-properties.html

Yaqi97:
@ftadel 

We need the 2 arrows to differentiate "bidirectional" links from In/Out links (the user can filter In/Out/Bidirectional/All in the display panel). 

I have redesigned the same arrowheads into one single patch object, as you suggested. I notice an improvement in the runtime, but it is not as instantaneous as it was with the previous tool. Hence, as you suggested previously (1.), we added a progress bar (calling `bst_progress`) for instances where there is a noticeable lag:
- when the user clicks on a node on the graph
- when the user switches between measure links, region max or region mean 
- when the user selects the next/previous region or "select all"
- when the user changes between In/Out/Bidirectional/All for directional graphs

I also tried adding the progress bar for changes to the intensity threshold, but this turned out to be inconvenient and distracting for sliders. Since the user can drag the slider, this generates multiple function calls resulting in the progress bar "flashing" on the screen for a fraction of a second. The changes for these filters isn't that slow, so we are thinking of keeping it as is, with no progress bar. The progress bar will be maintained for the slower cases discussed above. 


> Then it's probably useless. I don't know why we would keep an option that only makes the image darker...
> Could you remove the option from the menu?
> (maybe you can keep in the code, in case this gets fixed in the future versions of Matlab...)

Note: To build the 2 arrows within a single patch, I made changes to the alpha and color properties of our graphics. By doing so, the transparency options implemented earlier no longer work. Since we don't see a need for the transparency option since color blending is not supported on MATLAB graphics objects, should we remove that code entirely or do you prefer to keep it commented out in case someone wants to do something with it later (it won't actually run, but just for inspiration)? Or would you like us to re-implement the transparency with the current patch objects?

Yaqi97:
Hi @ftadel , a few additional questions for you:

1.
> 7. **Handle buggy ConnectGraphOptions**: After some crashes (which I don't know how to reproduce but which happened already twice when I was doing these tests), all the fields of structure `ConnectGraphOptions` get set to `[]` and then saved to the user preferences. This create a permanent bug (NodeSize is [], which crashes at line 3401), that can be solved by the user only by resetting the Brainstorm environment.

Which version of MATLAB were you using? I've encountered this issue when running it on R2009b. However, the use of object handles (instead of numeric handles) for graphics objects only became available starting from R2014b (see [release notes](http://math.loyola.edu/~loberbro/matlab/graphics_changes.pdf)). I've tested it out myself and we can only run our tool for releases >=2014b. So far I have not encountered this issue on R2014b, R2018b, R2020a and R2020b. If you were indeed testing our tool on R2009b, how were you able to generate and open the graphs themselves, given how our code treats the graphics as objects? 

2. I tried to re-import the anatomy in order to get the updated atlas names. However, I run into this error: 
image

This happened when I simply clicked on "Import anatomy folder" for my subject, both with my current database files, and with the sample tutorial data from the Brainstorm webpage: https://neuroimage.usc.edu/bst/download.php. Is there also a way to re-import the anatomy while still keeping all the data/graphs that we have processed until now (and not having to restart from the raw tutorial data)?

Thanks!

ftadel:
> should we remove that code entirely or do you prefer to keep it commented out in case someone wants to do something with it later (it won't actually run, but just for inspiration)?

Please remove completely anything you don't use.

> Which version of MATLAB were you using? I've encountered this issue after running it on R2009b. However, the use of object handles (instead of numeric handles) for graphics objects only became available starting from R2014b (see release notes). I've tested it out myself and we can only run our tool for releases >=2014b. So far I have not encountered this issue on R2014b, R2018b, R2020a and R2020b. I am guessing that the bug could be due to older versions using numeric handles versus more recent versions using object handles?

I tried many versions between 2011a and 2021a.
I'm OK to drop anything older than 2011a, but compatibility with versions between 2011a and 2014a is easy to maintain and should be maintained. 
Please use only `get` and `set` accessors to the graphic objects.

> I tried to re-import the anatomy in order to get the updated atlas names. However, I run into this error:
> `Some fiducial points were not defined properly in the MRI.`

Well, you need to set the fiducials, indeed... or use the menu `Import anatomy folder (auto)`.

> Is there also a way to re-import the anatomy while still keeping all the data/graphs that we have processed until now (and not having to restart from the raw tutorial data)?

No, it is not designed this way. Your not supposed to edit the anatomy after importing recordings.
If the goal is only to rename Mindboggle in DKT40, you could simply edit directly edit the atlas name in the Matlab structure (https://neuroimage.usc.edu/brainstorm/Tutorials/Scripting#Custom_processing).

But I thought the new goal was to use random parcellations instead of DKT anyway. Isn't it the case?

Thanks

Yaqi97:
@ftadel 

> > Which version of MATLAB were you using? I've encountered this issue after running it on R2009b. However, the use of object handles (instead of numeric handles) for graphics objects only became available starting from R2014b (see release notes). I've tested it out myself and we can only run our tool for releases >=2014b. So far I have not encountered this issue on R2014b, R2018b, R2020a and R2020b. I am guessing that the bug could be due to older versions using numeric handles versus more recent versions using object handles?
> 
> I tried many versions between 2011a and 2021a.
> I'm OK to drop anything older than 2011a, but compatibility with versions between 2011a and 2014a is easy to maintain and should be maintained.

Is there a possibility that you may have installed some toolboxes that have been upgraded by any chance? Or do you have any external software/libraries installed on your system when you tested out our tool? I am asking because I am not even able to open a connectivity graph with Matlab versions < 2014b since a lot of the graphics properties and Matlab built-in functions we used (such as 'PickableParts', node markers used to draw the nodes in our graphs, etc.) were only made available starting from version R2014b. I consulted the release notes (see [here](http://math.loyola.edu/~loberbro/matlab/graphics_changes.pdf) and [here](https://itptutor.tugraz.at/static//MatlabBooks//graphg.pdf)) as well as various discussions on the Mathworks website (example [here](https://www.mathworks.com/matlabcentral/answers/157740-how-do-i-click-an-object-below-another-object-in-matlab-r2014b?fbclid=IwAR3Ib6tSZN4HBBkeQAI6SItf_bAMXxOdJvQ0MGlndvOY7ClaiFFShhL5NB8)), and they all confirm that the graphics properties we have used were only made available in R2014b. It is not just an issue with accessing the properties; they are not even defined for versions <2014b... I am therefore wondering how you were able to test out our tool? Do I need to install something else or am I missing something? Apologies for the confusion; we were not aware of the requirements with older Matlab versions.


> But I thought the new goal was to use random parcellations instead of DKT anyway. Isn't it the case?

This was your comment from the beginning:

>  Anatomical parcellations such as DK, DKT and Destrieux, are not good canditates for computing connectomes - but @sbaillet will probably decide to keep it as it is for not having to redo all the screen captures.

The animations and screenshots in our tutorial page are just one example of a possible graph. Thoughts? 

Could you also clarify what is meant by "random parcellations" and how those work? We've always selected atlases for our graphs throughout this project (apologies for my lack of knowledge in anatomy). 

Thank you!

ftadel:
I'm sorry, I agree my previous message was confusing and my tests not done correctly.

**Matlab version**

If you can make your code compatible with older versions of Matlab with limited rewriting, it would be preferable. But if it requires some major editing, let's consider it only works with Matlab >= 2014b. 
In that case, the function `view_connect` should crash "gracefully" when called from an older version of Matlab, e.g. returning an error when `bst_get('MatlabVersion') < 804`.

** Atlas used in your example **

Please ignore my previous comment. 
We need first to make decisions at a more general level about which atlas we want to feature as a good default for connectivity analysis. It will take some time before we make this final decision and finish the main connectivity tutorial. This future evolution is not part of your current project. We will edit your tutorial and redo the screen captures later when we have clearer objectives.

** Speed considerations **

I'm more concerned with the global speed of the graph display.
Have you considered plotting all the arrows (not only the arrowheads) as a single patch yet?

Thanks

Yaqi97:
@ftadel 

> If you can make your code compatible with older versions of Matlab with limited rewriting, it would be preferable. But if it requires some major editing, let's consider it only works with Matlab >= 2014b.
> In that case, the function `view_connect` should crash "gracefully" when called from an older version of Matlab, e.g. returning an error when `bst_get('MatlabVersion') < 804`.

Compatibility with versions < R2014b would require a significant amount of rewriting since all our graphics objects and their interactions are currently drawn using properties that are only available >= R2014b. We will instead add error handling code to all the issues you've mentioned so far. 


> I'm more concerned with the global speed of the graph display.
> Have you considered plotting all the arrows (not only the arrowheads) as a single patch yet?

Thanks for the suggestion. I'm currently working on this, and will report back shortly.

Yaqi97:
@ftadel 
> I'm more concerned with the global speed of the graph display.
> Have you considered plotting all the arrows (not only the arrowheads) as a single patch yet?

I tried many ways to make this happen, but unfortunately it does not seem possible with the mathematical models we used to compute the coordinates of our links and arrowheads. MATLAB automatically attempts to "close" patch objects, which is problematic since our connectivity links are simply computed as an open arc between 2 nodes. In our code, each connectivity link is computed as a line with 100 data points, not a 2D object. As such, it was not possible to define it as a face in a patch object. On another note, it also does not make sense to define 100 vertices for a patch object (which takes a very long time to load on a dummy test script I tried). At this time, we cannot really change the algorithm we used to compute our graphics objects. 

As for speed considerations, the load time was already reduced by ~25% after I combined both arrowheads into a single patch object. Right now, directional graphs with double-arrowheads currently take 5-7 seconds to load on my computer, as indicated by the Profiler tool. This does not feel too long from our perspective, and non-directional graphs only take much less than that. 
We've also attempted to optimize our code as much as we could using the Profiler tool. 

@martcous 
While testing our visualization tool for compatibility with previous Matlab versions, I noticed that your fibers display also only work with Matlab versions >= R2014b. This is also the case when I tried it out with the code that's currently in the main distribution (I imported the file provided in your tutorial page). You're probably already aware of this and we won't be touching your code, but I simply wanted to point it out in case you wanted to do something about that.

helen-m-lin:
Hi all, 

Though it is added, we were wondering if the progress bar for when the user selects nodes is necessary. During our tests we found that the changes are not that slow and the progress bar may be distracting. Otherwise, all the other issues have been addressed and we tested for different MATLAB versions. Please let us know if we can move forward with the pull request! Unfortunately we cannot allocate much more time to this project due to our other commitments since the end of semester.

Thank you,
Helen & Yaqi

ftadel:
I went through another round of testing and posted a few minor modifications. It looks great!
Please feel free to contradict any change I made that you think is inadequate.

> At this time, we cannot really change the algorithm we used to compute our graphics objects.

Thank you for investigating various options.
As a general conclusion: the new graphs are much better-looking, more stable, more readable, and portable to future versions of Matlab. However, the rendering and reaction times are much higher.
We'll keep the two running side-by-side for a while and hope to get some additional feedback from users.

> Though it is added, we were wondering if the progress bar for when the user selects nodes is necessary. During our tests we found that the changes are not that slow and the progress bar may be distracting. 

Just leave it like this for now, this will be very easy to change later. 

> Unfortunately we cannot allocate much more time to this project due to our other commitments since the end of semester.

I understand, you've already been beyond our expectations. Thank you very much for all the quality support you've provided to help us integrate your developments directly into the main Brainstorm distribution.

I just have a few last questions, in case these are are easy for you to address.
1) Using the introduction tutorials, in Process1, I selected the file `Subject01/S01_AEF_20131218_01_600Hz_notch/Avg: deviant (39 files)/MN: MEG(Constr) 2018`
2) I ran this process:
image
3) I opened the graph for the output file, then gradually reduce the intensity threshold and increase the selected frequency. I obtain the graph below:
image
4) Issue 1: Some arcs are not complete and look like they are floating around. Not sure whether this is a bug or an expected feature.
5) Issue 2: There are some simple arrows and some double arrows: what is the meaning of this? (sorry, I probably missed something obvious or already discussed). Below, if I lower the threshold again, the arrow from the top left to the bottom-right changes from a simple to double arrowhead.
image
6) Issue 3: Same file. I click on "Show labels for selection only", then click on a node (all the labels disappear as expected), then click on "Show labels for selection only" again. I would expect to get all the labels back, but instead they all disappear and never come back. 
image 
image

Yaqi97:
> 4. Issue 1: Some arcs are not complete and look like they are floating around. Not sure whether this is a bug or an expected feature.

This is indeed a bug on our end - I pushed a fix for it.

> 5. Issue 2: There are some simple arrows and some double arrows: what is the meaning of this? (sorry, I probably missed something obvious or already discussed). Below, if I lower the threshold again, the arrow from the top left to the bottom-right changes from a simple to double arrowhead.

For directional graphs, links are classified as "In", "Out" or "Bidirectional" (as per the options in the display panel). In our design, "In" and "Out" links have a single arrowhead, while "Bidirectional" links have double arrowheads. A link is classified as "Bidirectional" when both the link from A ->B and the link from B->A satisfy the thresholds and are visible in the graph. Therefore, you will see both links connecting a pair of nodes when double arrowheads are present. If you see an arrow change from single arrowhead to double arrowhead, this means that you have just crossed the threshold at which the second link connecting this pair of nodes becomes visible (you should see that second link appear as well). 

In the old tool, all bidirectional links were green. We thought that this was not desirable as the colour would neither reflect the actual connectivity score nor the directionality of each link. This was why we designed our tool to have these arrowheads. 

On another note, we would be happy to share a more detailed report of our design choices and mathematical derivations with you to help with the understanding and the evolution of this tool in the future, if you would like to have it. This report was completed in April as part of our coursework and some minor changes have been made through this pull request, but the majority of it stays accurate. @ftadel Could we send it to your McGill email? (We won't be putting this on a public GitHub as it is considered academic coursework)

> 6. Issue 3: Same file. I click on "Show labels for selection only", then click on a node (all the labels disappear as expected), then click on "Show labels for selection only" again. I would expect to get all the labels back, but instead they all disappear and never come back.

This is because the "Show labels" options is not automatically selected when "Show labels for selection only" gets unselected. If you re-select "Show labels", all the labels will be back. Note that the nodes that are not selected are grey, as per our design. If you unselect the node that you had clicked on, all the nodes and labels will be back with their bright colors. 
image

This is identical to the behaviour we see in the previous tool. I'm assuming there may be cases where the user may not want any labels displayed on the graph? In fact, the "Show labels" option you renamed in our popup menu corresponds to the "Labels Display" -> "Measure Nodes" option in the old tool's popup menu (that is what you wanted, right?). Everything still behaves the same as before (i.e., "Show labels for selection only" is mutually exclusive to the other options, etc.). 

Lastly, we want to sincerely thank you for taking the time to provide us detailed comments and feedback. We've definitely learned a lot throughout the process.

ftadel:
> In our design, "In" and "Out" links have a single arrowhead, while "Bidirectional" links have double arrowheads. A link is classified as "Bidirectional" when both the link from A ->B and the link from B->A satisfy the thresholds and are visible in the graph

This is what I was expecting, but I did not realize that an arc could be doubled with a straight line. In the screen captures I posted, the top-left/bottom-right bidirectional connection has one yellow curved link and one purple straight link - plus we don't see the purple arrows, which are under the central node. When decreasing the threshold until the double-arrow appears, I did not realize that this purple link appeared too. Now it all makes sense, thanks.

> On another note, we would be happy to share a more detailed report of our design choices and mathematical derivations with you to help with the understanding and the evolution of this tool in the future, if you would like to have it. This report was completed in April as part of our coursework and some minor changes have been made through this pull request, but the majority of it stays accurate. @ftadel Could we send it to your McGill email?

Yes, please send it to me by email.

> This is because the "Show labels" options is not automatically selected when "Show labels for selection only" gets unselected.

Intuitively, I would expect to get all the labels back when unchecking "Show labels for selection only".
Would it make sense to always select again the option "Show labels" after unselecting "Show labels for selection only"?
(It's good to have the option to hide all the labels for producing specific figures, but I'm not sure who would like to get no labels at all in a routine exploratory use)

Other than that, I think this is ready for merging.
Can you please confirm there is nothing else pending?

Yaqi97:
> Yes, please send it to me by email.

Sent.

> Intuitively, I would expect to get all the labels back when unchecking "Show labels for selection only".
> Would it make sense to always select again the option "Show labels" after unselecting "Show labels for selection only"?
> (It's good to have the option to hide all the labels for producing specific figures, but I'm not sure who would like to get no labels at all in a routine exploratory use)

Ok. I changed the logic to make it select "Show labels" when "Show labels for selection only" gets unselected. Now, all the labels are back when the user unselects "Show labels for selection only". There is always the possibility to hide all the labels by unselecting "Show labels", if the user wants to. 
>
> Other than that, I think this is ready for merging.
> Can you please confirm there is nothing else pending?

We've tested everything on our side, and can confirm that we are ready for merging. Thanks!

PR #410, opened 2021-05-14
Bugfix: Progress bar does not always hide

rcassani:
Replaces the writing and erasing on the command window with a 50 ms delay.
Linux systems without this bug problem are not affected.
Bug reported for OpenJDK in https://bugs.openjdk.java.net/browse/JDK-6483868

### DIFF
--- a/toolbox/core/bst_progress.m
+++ b/toolbox/core/bst_progress.m
@@ -152,8 +152,13 @@ if isempty(pBar)
     GlobalData.Program.ProgressBar = pBar;
 end
 
-% Linux: the dialog needs to be displayed for a short period before being hidden
+% Linux: need to print something on the command window (don't know why...)
 if strcmpi(commandName, 'stop') && ismember(computer('arch'), {'glnx86', 'glnxa64'})
+    drawnow();
+    fprintf(' ');
+    fprintf('\b');
+    drawnow();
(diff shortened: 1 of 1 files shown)


toolbox/core/bst_progress.m
```diff
@@ -152,12 +152,9 @@
...
 end
 
-% Linux: need to print something on the command window (don't know why...)
+% Linux: the dialog needs to be displayed for a short period before being hidden
 if strcmpi(commandName, 'stop') && ismember(computer('arch'), {'glnx86', 'glnxa64'})
-    drawnow();
-    fprintf(' ');
-    fprintf('\b');
-    drawnow();
+    pause(0.05);
```
ftadel, line 157: What about keeping both?
and only adding `pause(0.05);` before the second `drawnow`?

If I implemented it this way it is that it was helping for something at some point.
It's impossible to test all the possible configurations of OS/Matlab/Java versions for Linux, so we won't know if removing this printing trick alters the behavior somewhere else.

Unless it causes problems somewhere?
If it doesn't cause any specific problem, keeping both is maybe a slightly safer option, with a negligible impact on the interface reaction time.

rcassani: I agree, it's safer to keep both.

PR #411, opened 2021-05-18
[Cat12] Call spm_jobman('initcfg') before creating the batch file

Edouard2laire:
Hello Francois, 

it seems that in some case, matlabbatch (responsible for the function cfg_dep) is not in the matlab path before executing cfg_dep.

This folder is included in the path when calling spm_jobman('initcfg') (line 169 of spm_jobman) I therefore think it should be called before creating the cat12 batch

Regards, 
Edouard

### DIFF
--- a/toolbox/process/functions/process_segment_cat12.m
+++ b/toolbox/process/functions/process_segment_cat12.m
@@ -274,9 +274,22 @@ function [isOk, errMsg] = Compute(iSubject, iAnatomy, nVertices, isInteractive, 
         % Save modification on hard drive
         bst_save(file_fullpath(T1FileBst), sMri, 'v7');
     end
-
+    
+    
     % ===== CALL CAT12 SEGMENTATION =====
     bst_progress('text', '<HTML>Starting SPM batch... &nbsp;&nbsp;&nbsp;<FONT COLOR="#707070"><I>(see command window)</I></FONT>');
+    
+   % Switch to CAT12 expert mode
+    cat12('expert');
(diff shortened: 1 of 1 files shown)


ftadel:
Ooops, reading the other comment you made (which I got by email but can't see on Github??) I understood what you meant.
The problem is not the `matlabbatch` variable, but the `matlabbatch` folder which contains the definition of the `cfg_dep` object...

I pushed your change in this commit:
https://github.com/brainstorm-tools/brainstorm3/commit/75747df1c56ba34ca9fa81ed7e915dd3552b485e

Sorry for misunderstanding your initial proposition.

ftadel:
Note there might be some weird things occurring with the current implementation: 
I pushed the changes to support CAT 12.8 before it was actually made the default distribution (at the moment, Brainstorm still downloads automatically the latest CAT 12.7 release). The change will be made on the CAT side in the next few days. Keep an eye on this, as it will result in significant improvements.

Edouard2laire:
> Sorry for misunderstanding your initial proposition.

I have to admit that my message was not very clear :) For the original message.i put on one of your commit; I actually deleted it first because I thought the issue was with my SPM setup. 


> Note there might be some weird things occurring with the current implementation:
> I pushed the changes to support CAT 12.8 before it was actually made the default distribution (at the moment, Brainstorm still downloads automatically the latest CAT 12.7 release). The change will be made on the CAT side in the next few days. Keep an eye on this, as it will result in significant improvements.

Nice! looking forward to it :)

PR #414, opened 2021-06-07
DICS function

vyoussofzadeh:
This is a DICS beamformer implementation for the Brainstorm (BS) software package that was previously shared from,
https://github.com/vyoussofzadeh/DICS-beamformer-for-Brainstorm

For convenience and compatibility, variable inputs were integrated into a BS option GUI interface.


ftadel:
Thank you for this contribution!

I edited your code heavily but I'm not 100% sure of what I did, so I'm sending you this file here for you to review:
[process_ft_sourceanalysis_dics.zip](https://github.com/brainstorm-tools/brainstorm3/files/6671858/process_ft_sourceanalysis_dics.zip)

The changes I made:
- Reformatted the code to match the standards in Brainstorm
- Switched to the new plugin system for FieldTrip integration: https://neuroimage.usc.edu/brainstorm/Tutorials/Plugins
- Removed all the other options than DICS from your function, and removed DICS from the other ft_sourceanalysis process
- Fixed typos
- Simplified by removing lots of unnecessary lines (leftovers from previous process)
- Preventing from saving non-referenced files in the Brainstorm database (this won't be possible with the future versions of the database)
- There are a few minor things I'd like to add, but we could do this in a second round of review

A few things you could do to finalize this work:
- Validate the default values for all the parameters (there were all set to a non-existing value "strict", I replaced that with the first entry in the list)
- Make sure it gives the same results as before for some examples you have
- Push the modified code to your repository
- Remove all the extra MAC files (.DS_STORE etc)
- Update your README/tutorial on your github repo: https://github.com/vyoussofzadeh/DICS-beamformer-for-Brainstorm
(or add it to Brainstorm)
- It would be even better if the data used for the example was the same as the introduction tutorials (auditory oddball)

Thanks!

ftadel:
PS: Make sure you are starting again from the last commits on the Brainstorm repo.
I made a few modifications that are necessary to run the process I sent you.

vyoussofzadeh:
Hi Francois, apologies for the delay in getting you back, and thanks for the great edits, the script (DICS-BF pipeline) looking much better now. I accepted all the suggested changes. I also made some edits in some parts, e.g, reading of trial data. Finally, I compared findings with the previous version and found both findings identical. I pushed the modified script for further review. A tutorial using example data sounds like a great idea, yes, I am interested, please create a new page so I can contribute.
Looking forward to your comments/review!

ftadel:
Looks good!
I fixed a few things and some formatting before merging.
It's now in the main Brainstorm distribution, people will start asking about it soon :)

I created an empty page on the wiki for you to write a tutorial for this function:
https://neuroimage.usc.edu/brainstorm/Tutorials/DICS
I'll send the information for you to create an account on the wiki at your MCW email address (let me know if you don't get it).

It would be great if you could illustrate it starting from the introduction tutorial dataset.
- If you don't have it properly prepared on your computer anymore, you can try running `tutorial_introduction.m` to re-run all the processing steps automatically. You need to have the `sample_introduction` tutorial on your computer.
- If the epochs are two short (baseline is only 100ms long in this example), you can give instructions to import a new set of longer deviant epochs for example: 
image
- Follow the logic of the other tutorials: describe each step, include screen captures for every relevant step, showing the input files, the process options, the output files in the database explorer. 
- Example: https://neuroimage.usc.edu/brainstorm/Tutorials/DipoleFitting
- Document all the process options clearly, if possible giving the correspondence to the FieldTrip variable names.

Thank you very much for this contribution!

PR #415, opened 2021-06-15
Remove message that is not correct

ajoshiusc:
the fprintf statement that is being removed prints out a message unconditionally. This is incorrect behavior. The in_label_bs function called before it has its own error handling, so no additional error handling is needed.


ftadel:
Thanks for fixing this.

PR #420, opened 2021-06-29
small change PLV in time

danielemarinazzo:
Computing "angle" only once results in less wrapping artefact. Execution time is analogous, or just slightly faster. 
Added (commented) an alternative implementation too.


ftadel:
@sbaillet 
You initially wrote this code. 
Are you OK with the suggested modifications?

sbaillet:
Approved.

ftadel:
@danielemarinazzo 
Thanks for these contributions

PR #423, opened 2021-07-06
First commit, coherence-2021

rcassani:
This is the first draft towards the coherence-2021 update. 
It shows that the Input data X and Y are cell arrays, where each cell is a 2D array `[nSignals, nSamples]`.
For the moment the case N×N MSC is shown in `bst_cohn_2021.m`. Additional cases will be added in next commits.
An additional function `bst_epoching.m` is added for convenience. Its current path may not be the ideal.


toolbox/connectivity/bst_cohn_2021.m
```diff
@@ -0,0 +1,250 @@
...
+end
+% Minimum number of windows for signals in X and signals in Y
+minWin = nFiles * floor( (min(nSamplesXs) - nOverlap) / (nWinLen - nOverlap));
+
+% Error and Warning
+minWinError = 2;
+minWinWarning = 5;
+% ERROR: Not enough time points
+if minWin < minWinError
+    nMinMessage = nWinLen * (1 - nOverlap) * (minWinError - 1);
```
ftadel, line 135: This produces error messages that do not make sense. 
nMinMessage computation must be reviewed.

toolbox/connectivity/bst_cohn_2021.m
```diff
@@ -0,0 +1,250 @@
...
+nSignalsY = size(Ys{1}, 1);
+Syy = zeros(nSignalsY, length(freq));
+Sxy = complex(zeros(nSignalsX, nSignalsY, length(freq)));
+nWin = 0;
+
+for iFile = 1 : nFiles
+    % Compute Syy
+    y = Ys{iFile};    
+    % Epoching 
+    epy = bst_epoching(y, nWinLen, nOverlap);
```
ftadel, line 167: I'm not convinced by the necessity to have a separate subfunction.
bst_epoching does more than what we need here (2 of the 3 outputs are not used).
If written in a compact way, it would be only a few lines of code added here.

Since it duplicates the input data, it might also pack the memory unnecessarily.
Have you tested how much it decreases the execution speed, to have the ftt called in this vectorized format instead of in a for loop on the epochs?

If for clarity you want to have separate function, or if this helps you factoring code in when you add the other cases, maybe this could be moved to a local subfunction. I guess it won't be used much outside of this function.

ftadel (changes requested):
I added the wrappers you needed to call your new function bst_cohn_2021.m from the interface.  I created a new set of process functions and deprecated the previous ones.
I hope it's not too buggy...

Could you write a simple script, based on simulated data, that allows computing exactly the same thing with the older and the new processes? 
(it could be the Test functions of the process_cohere1* functions for instance).
This would help for comparing the execution speed of the two versions, in order to optimize our implementation choices as we go.

I review the bst_cohn_2021 function in more details when you are more advanced with the integration of the other options (or whenever you need - just ask)

toolbox/connectivity/bst_cohn_2021.m
rcassani: It does a bit more. I thought it could come handy whenever we need to epoch data and keep a reference to when the epochs are places `time_vector(ixs_ctr)`.

Yes, the vectorization of FFT helps, it's between 2 and 5 times faster. It also adds clarity. See this gist:  https://gist.github.com/rcassani/ca12d325c82997988e6313e1454da385

toolbox/connectivity/bst_cohn_2021.m
rcassani: Fixed in 668c1e7c1e84975a45d58572458d1bd2b05e970c

rcassani:
Find the comparison/test for `bst_cohn.m` and `bst_cohn_2021.m` here: 
https://gist.github.com/rcassani/ca12d325c82997988e6313e1454da385#file-fft_vectorized_test-m

ftadel:
> Find the comparison/test for bst_cohn.m and bst_cohn_2021.m here

Great, thanks.
Could you please include these test functions in one of the coherence process?
Or anywhere else in the actual brainstorm code repository, for future reference?

rcassani:
> Could you please include these test functions in one of the coherence process?  

Comparison included in `process_cohere1n_2021('Test')` in 34730dff9e4930594894acd0784a0d5bfb339f1b

Regarding the epoching function, I think we can use it in the future to vectorize some other processes that are window-wise are are written as loops.

rcassani:
@ftadel, 
- The epoching function was moved inside `bst_cohn_2021.m`
- Test functions added where pertinent

ftadel:
Do you think this is ready to be merged?

rcassani:
@ftadel 
It's ready to merge.
P.S. There is a commented section to compute `Sxy` in a vectorized way, it's much faster but it takes a good amount of RAM.

ftadel:
I moved back the old process functions from the deprecated folder to the regular process folder.
Since the results we obtain can look pretty different with the two processes, this will make it easier to compare the old and new values.
I hope this OK with you?

Could you document this update somewhere in the tutorials, so we could point at some online documentation (eg. the `Description` field of the process functions)?
Thanks!

rcassani:
Sounds good. The changes will be documented in the connectivity tutorials.

PR #422, opened 2021-07-03
improved/robust version for the iso2mesh option within process_fem_mesh

tmedani:
Hi all,

This PR is about the issue observed when the iso2mesh option is used, and surf2mesh produces more element labels than the input surfaces.
link to the reported issue: https://neuroimage.usc.edu/forums/t/duneuro-error-calculation/28965/6?u=tmedani

The change on this code will identify any output label based on the distance comparison between the inputs surfaces and the outputs labels.
I believe that is much robust than the previous version.

This change can be also useful for custom fem mesh generation for sEEG/iEEG models.

Thanks :)

PR #427, opened 2021-07-24
improved version of iso2mesh option

tmedani:
Hi all,

This is the recovered code version that I lost from from this PR : https://github.com/brainstorm-tools/brainstorm3/pull/422

This PR is about the issue observed when the iso2mesh option is used, and surf2mesh produces more element labels than the input surfaces.
link to the reported issue: https://neuroimage.usc.edu/forums/t/duneuro-error-calculation/28965/6?u=tmedani

The change on this code will identify any output label based on the distance comparison between the inputs surfaces and the outputs labels.

It fixes also the tissue labels from the MNE bem surfaces.

I believe that is much robust than the previous version.

This change can be also useful for custom mesh generation for sEEG/iEEG models.

Thanks :)

### DIFF
--- a/toolbox/process/functions/process_fem_mesh.m
+++ b/toolbox/process/functions/process_fem_mesh.m
@@ -55,7 +55,8 @@ function sProcess = GetDescription() %#ok<DEFNU>
                                        '<B>Brain2mesh</B>:<BR>Segment the <B>T1</B> (and <B>T2</B>) <B>MRI</B> with SPM12, mesh with Brain2Mesh<BR>', ...
                                        '<B>SimNIBS</B>:<BR>Call SimNIBS to segment and mesh the <B>T1</B> (and <B>T2</B>) <B>MRI</B>.', ...
                                        '<B>FieldTrip</B>:<BR> Call FieldTrip to create hexahedral mesh of the <B>T1 MRI</B>.'; ...
-                                       'iso2mesh', 'brain2mesh', 'simnibs', 'fieldtrip'};
+                                       '<B>Iso2mesh-2021</B>:<BR>Call iso2mesh (2021 version) to create a tetrahedral mesh from the <B>BEM surfaces</B><BR>',
+                                       'iso2mesh', 'brain2mesh', 'simnibs', 'fieldtrip','iso2mesh-2021'};
     sProcess.options.method.Type    = 'radio_label';
     sProcess.options.method.Value   = 'iso2mesh';
     % Iso2mesh options: 
@@ -135,7 +136,7 @@ function OutputFiles = Run(sProcess, sInputs) %#ok<DEFNU>
     end
     % Method
     OPTIONS.Method = sProcess.options.method.Value;
-    if isempty(OPTIONS.Method) || ~ischar(OPTIONS.Method) || ~ismember(OPTIONS.Method, {'iso2mesh','brain2mesh','simnibs','fieldtrip'})
+    if isempty(OPTIONS.Method) || ~ischar(OPTIONS.Method) || ~ismember(OPTIONS.Method, {'iso2mesh','brain2mesh','simnibs','fieldtrip','iso2mesh-2021'})
         bst_report('Error', sProcess, [], 'Invalid method.');
         return
     end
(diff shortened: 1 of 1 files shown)


toolbox/process/functions/process_fem_mesh.m
```diff
@@ -332,6 +333,167 @@
...
+                        sSubject.Surface(sSubject.iScalp).FileName};
+                    TissueLabels = {'brain', 'skull', 'scalp'};
+                else
+                    errMsg = ['Method "' OPTIONS.Method '" requires three surfaces: head, inner skull and outer skull.' 10 ...
+                        'Create them with process "Generate BEM surfaces" first.'];
+                    return;
+                end
+            % If surfaces are given: get their labels and sort from inner to outer
+            else
+                % Get tissue label
```
ftadel, line 359: Why not leaving the original code?
```
                % Get tissue label
                for iBem = 1:length(OPTIONS.BemFiles)
                    [sSubject, iSubject, iSurface] = bst_get('SurfaceFile', OPTIONS.BemFiles{iBem});
                    if ~strcmpi(sSubject.Surface(iSurface).SurfaceType, 'Other')
                        TissueLabels{iBem} = GetFemLabel(sSubject.Surface(iSurface).SurfaceType);
                    else
                        TissueLabels{iBem} = GetFemLabel(sSubject.Surface(iSurface).Comment);
                    end
                end
```

toolbox/process/functions/process_fem_mesh.m
```diff
@@ -332,6 +333,167 @@
...
+                    TissueLabels{iBem} = label;
+                end
+            end      
+            % Load surfaces
+            bst_progress('text', 'Loading surfaces...');
+            bemMerge = {};
+            disp(' ');
+            nBem = length(OPTIONS.BemFiles);
+            distance_in = zeros(1,nBem);
+            for iBem = 1:nBem
```
ftadel, line 385: Something does not work correctly with this code.
Take the ICBM152 default anatomy folder.
1) Right-click on the subject folder > MRI segmentation > Generate FEM mesh, in `BemFiles` the surfaces are ordered innerskull-outerskull-head => It works
2) Select the the three surfaces > right-click > Generate FEM mesh, in `BemFiles` the surfaces are ordered head-outerskull-innerskull => It crashes with the message `Error: Problem with tissue labels: Brainstorm cannot understand the output labels (2  3).`

Therefore, depending on the order of the input files, it produces different outputs.
The previous implementation didn't have this problem.

ftadel (changes requested):
I'm glad you could recover it.
It will now be safer in a separate branch :)

There are still some issues. See the comments in the code

toolbox/process/functions/process_fem_mesh.m
tmedani: I did it before, but it was the main reason for the display error when the surfaces are from the MNE files.

toolbox/process/functions/process_fem_mesh.m
tmedani: Indeed, the distances are computed using the center of the first bemfile, I changed the code and I compute a center based on all nodes and not only the first bemfile, but all the bemFiles

tmedani:
> I'm glad you could recover it.
> It will now be safer in a separate branch :)
> 
> There are still some issues. See the comments in the code

Thanks, I had a big part of the code in an old branch, then it was not difficult to recover it.

ftadel:
I fixed a few things:
- restored the previous `GetFemLabel` call, because it definitely help labeling correctly the layers, but I fixed it to work with MNE's `outerskin`
- reordered the `bemMerge` variable, to obtain consistent output, no matter which is the order of the input surfaces

Does this work for you?

tmedani:
> I fixed a few things:
> 
> * restored the previous `GetFemLabel` call, because it definitely help labeling correctly the layers, but I fixed it to work with MNE's `outerskin`
> * reordered the `bemMerge` variable, to obtain consistent output, no matter which is the order of the input surfaces
> 
> Does this work for you?

of course, since it's working correctly is most important.

Issue #419, opened 2021-06-29
(spectral) Granger causality implementation


danielemarinazzo:
Hello

the traditional way of implementing GC (i.e. comparing a full and a reduced model), leads to some issues, pointed out to in this paper https://www.pnas.org/content/114/34/E7063.short. The main issue is that the reduced model is VAR, the full model is VARMA; the full model has a given model order, the reduced model has order infinite by definition.
Luckily for the field, these issues had been already addressed prior to the Stokes&Purdon paper. One efficient way to solve them is by means of state space models.
More info (and pointers to code) here

https://f1000research.com/articles/6-1710
https://pubmed.ncbi.nlm.nih.gov/29883736/

I would be happy to help, I am not very familiar with the Brainstorm way of coding and dealing with data.

ftadel:
Is this issue the same as what is discussed in this PR: https://github.com/brainstorm-tools/brainstorm3/pull/433 ?

danielemarinazzo:
yes

rcassani:
Closed by #433

PR #428, opened 2021-07-25
Improvements to Granger Causality evaluation

DavideNuzzi:
Modified bst_granger.m and bst_granger_spectral.m (their results were not correct)
Implemented the conditional GC both in time and frequency domain.
The functions "YuleWalker_Inverse" and "YuleWalker_Mask" are only used for the solution (maybe they can be moved in another folder). 

Note that the approach I choose to follow in order to evaluate GCs can be readily generalized. For example it will be possible to evaluate multivariate GC (multiple source variables) both in time and frequency domain.


ftadel:
Thank you for these propositions. This represents a lot of work, and will require a lot of time to review on our end.
There are two distinct objectives to this PR, which need to be reviewed/approved separately: 
- the modification of the existing Granger causality code
- the inclusion of new methods, which is far from being finished, as it would require the corresponding process functions and more in-depth discussions about their necessity with the PIs of the project.

@DavideNuzzi 
- Please split this PR in two different PRs, to address each item separately.
- Remove capital letters from any new file names
- Include the small sub-functions in existing functions or "private" subfolders (we don't want to add too many .m files that are called in only one or two places to be added to the Matlab path and the general workspace)
- Please be patient, as you may not get any additional feedback before mid-August

@HosseinShahabi @richardmleahy @rcassani @sbaillet 
Could you please share your first impressions about these suggested changes?

Thanks

danielemarinazzo:
dear team @HosseinShahabi @sbaillet @richardmleahy @rcassani @ftadel 
this PR is related to #419 . I have gone through the code with @DavideNuzzi

DavideNuzzi:
Dear team, thank you for your response. I will split the PR into two different ones and modify the file names. 
Regarding the sub-functions YuleWalker_Mask.m and YuleWalker_Inverse.m, they are called by each of the GC functions that I've written (pairwise and conditional GC in both time and frequency domain). That's the reason why I preferred to define them in separate .m files instead of duplicating them in the main functions. Do you suggest I do it anyway?

PR #430, opened 2021-08-12
Update process_fooof.m

lucwilson:
Updated process description and citation.

Added functionality for ignoring power line frequencies and their first two harmonics.

Added option to return spectrum for each channel in FOOOF struct (already present in TF field, but SPM/FT cannot access from running function outside bst)

Removed bst_round dependencies (but maintained functionality; sorry Francois, it was a good script)


ftadel:
Thanks for maintaining this great tool!

When you get time, please add a description of the new option in the tutorial:
https://neuroimage.usc.edu/brainstorm/Tutorials/Fooof

Issue #431, opened 2021-08-13
FEM tutorial: Issues to fix


ftadel:
1) Remesh the SimNIBS model with iso2mesh doesn't work. Here is the error message:

````
    FEM> 1. scalp: C:\Work\Protocols\TutorialFem\anat\Subject01\tess_scalp.mat
    FEM> 2. skull: C:\Work\Protocols\TutorialFem\anat\Subject01\tess_skull.mat
    FEM> 3. white: C:\Work\Protocols\TutorialFem\anat\Subject01\tess_white.mat
    FEM> 4.  gray: C:\Work\Protocols\TutorialFem\anat\Subject01\tess_gray.mat
    FEM> 5.   csf: C:\Work\Protocols\TutorialFem\anat\Subject01\tess_csf.mat
     
    generating tetrahedral mesh from closed surfaces ...
    creating volumetric mesh from a surface mesh ...

    ***************************************************************************
    ** Error: Line 117:  surf2mesh (line 117)
    ** Tetgen command failed
    **
    ** Call stack:
    ** >surf2mesh.m at 117
    ** >process_fem_mesh.m>Compute at 442
```

2) Pre-processing: The PSD screen capture does not seem to be from this file (not the same time definition)

3) Averaging: The averaging screen capture shows a weighted average: this is confusing => Uncheck the "weighted" box 

4) Noise covariance matrix:
 - The name of the epochs in the screen capture is wrong ("2" and not "STI101")
 - The is a mention to a DC offset removal that was not applied when importing the epochs
 - The time window doesn't match between the text and the screen capture.

ftadel:
@juangpc  @tmedani 

I also posted a draft for the processing script:
https://neuroimage.usc.edu/brainstorm/Tutorials/FemMedianNerve#Scripting
https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/script/tutorial_fem.m

Please use try to use it to reproduce your computations: 
it will allow to automate the rebuild of everything when we change a parameter in the script or when we update Brainstorm or Duneuro.

For addressing the issues: please post your comments as you work on them, and I will edit the first comment of this thread to keep track of what is left in the todo list.

Thanks!

juangpc:
Will do. 

As a start. I can't unzip iso2mesh's zip folder when installing the plugin. I get "An internal error has ocurred."
I'm trying to see what's the matter. Not related to permissions, nor incorrect paths. Maybe zip versions?

[...]

PR #432, opened 2021-08-17
Wrap unzip function with bst_unzip to globally manage features 

juangpc:
Hi, 

After finding some problems unzipping a zip file downloaded from github [issue](https://github.com/brainstorm-tools/brainstorm3/issues/431#issuecomment-898508442), I've found this solution.  Not everyone is experiencing this problem (actually only me 😄 ) but, just in case. 

Apparently Matlab is changing its unzip function while adapting to a java-less life. And this new version might not be as universally compatible as the old unzip version. While this new version becomes more stable, we can: 

- use this bst_unzip function which basically wraps the unzip call.  I like this abstraction because I see no reason why any function using unzip should know or care about different versions of unzip, java dependencies, etc... 
- if a feature (i.e. ZIPV2) fails once, it is probably going to fail more than once. So it can globally be managed either through some settings parameter or some static variable inside the function ("persistent"). That's not implemented right now.
- I see that there are 30 calls to unzip right now in the repo. Some with 1 input parameter (zipFilename) or 2 (zipFilename, outputFolder). bst_unzip is compatible with both formats.
- Imho the complete switch [here ](https://github.com/brainstorm-tools/brainstorm3/blob/08c766c27db3f54f2b1378a8a250de84f1c08d53/toolbox/core/bst_plugin.m#L1074)could be abstracted with a bst_unpackage file. I see no reason why plugin should know or care about zip/tar/ispc/system calls, etc... (this hasn't been done). 

If you want to drop this. That's ok. Just trying to help.

### DIFF
--- a/toolbox/core/bst_plugin.m
+++ b/toolbox/core/bst_plugin.m
@@ -1073,7 +1073,7 @@ function [isOk, errMsg, PlugDesc] = Install(PlugName, isInteractive, minVersion)
     % Unzip file
     switch (pkgFormat)
         case 'zip'
-            unzip(pkgFile, PlugPath);
+            bst_unzip(pkgFile, PlugPath);
         case 'tgz'
             if ispc
                 untar(pkgFile, PlugPath);
--- a/toolbox/misc/bst_unzip.m
+++ b/toolbox/misc/bst_unzip.m
@@ -24,19 +24,25 @@ function varargout = bst_unzip(zipFilename, varargin)
 % For more information type "brainstorm license" at command prompt.
 % =============================================================================@
 %
-% Authors: Juan GPC, 2021
+% Authors: Juan GPC, Francois Tadel, 2021
 
 narginchk(1,2);
 nargoutchk(0,1);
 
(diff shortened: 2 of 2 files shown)


juangpc:
@tmedani you and some others can push modifications to this branch on my fork. Whatever suits you.

tmedani:
For me, it's fine Juan, since it handles more cases, including the current case.

tmedani:
We may need to ask @ftadel

ftadel:
Thanks.
But let's not invest more time in fixing all the calls to unzip(), since this looks like a very marginal bug that will ever exist only on one Matlab release on a handful of computers and for one specific zip file. If you see this happening in other places in the code, replace unzip() with bst_unzip() one occurrence at a time.

> Imho the complete switch here could be abstracted with a bst_unpackage file. I see no reason why plugin should know or care about zip/tar/ispc/system calls, etc... (this hasn't been done).

Yes, we could. But so far bst_plugin is the only function that needs to unpackage files without knowing in advance what the compression format is, so it would only add an extra function without factorizing anything or reducing the amount of code.

PR #433, opened 2021-08-18
Improvements to Granger Causality evaluation

DavideNuzzi:
Modified bst_granger.m and bst_granger_spectral.m
Their results were not correct (when compared to other popular frameworks like MVGC). In particular, bst_granger.m seemed to rely on a double VAR model identification, which is known to yield incorrect results. My correction solves the Yule-Walker equations under arbitrary constraints (of the VAR model couplings) and can be readily generalized to evaluate conditional Granger causality and multivariate Granger causality (both in the time and frequency domains).


ftadel:
@HosseinShahabi @rcassani @richardmleahy 
Could you please review and discuss these requested modifications?
Thanks

HosseinShahabi:
Dear @DavideNuzzi 

Thank you very much for submitting this pull request. We appreciate your interest in Brainstorm and your desire to improve it. 

We are glad that you raised this important issue and we understand that the recent research has emphasized the pros of the new methodology for the computation of Granger causality in comparison to the traditional approach. 

As you mentioned, the MVGC toolbox is a well-known software for GC computation. When I checked your code, I realized the approach you employed for the time-domain GC is similar to one in the MVGC framework. However, your code for the spectral  GC is based on another article. 

While we are trying to improve our software based on your suggestions, I have the following questions from you:

Is there any specific reason that you selected not to use the MVGC approach for spectral GC? Also, does your code uses a better algorithm for the computation of GC? 

I understand that you personalized the new algorithm for our toolbox but do you think there will be any problem if we fully integrate the MVGC toolbox in Brainstorm? 

Looking forward to collaborating with you, 
Hossein

ftadel:
@HosseinShahabi 
Thank you for reviewing this PR and for your report.

@DavideNuzzi 
Our interest for using an existing toolbox instead of customized code is related mostly to the maintenance and the support of the software. If the code is maintained directly by the authors of the methods, it is a guarantee that all the bugs will be fixed and that it will be updated to work with the future versions of Matlab, and we can redirect all the users' questions and bug reports to the library authors. It means much less work us on the long term. When possible, reusing existing packages without modifications is a good open-source practice.

Brainstorm includes now an easy-to-use package manager to download third-party libraries, so downloading the MVGC toolbox would require only to add a few lines of code to bst_plugins:  https://neuroimage.usc.edu/brainstorm/Tutorials/Plugins

Let us know what you think - Thanks

ftadel:
@HosseinShahabi 
Could you please provide an update on your work?
Thanks

ftadel:
@DavideNuzzi Please be patient: @HosseinShahabi is still planning on working on this.

ftadel:
@HosseinShahabi Any update?

AdelleBernal:
What is the status of this modification and how does it affect the current GC measures in the current BrainStorm version?

rcassani:
The computation of Granger causality in Brainstorm has been updated:

* Time- and frequency-domain GC process have the option `Granger causality method` to choose between `Unconditional` and `Conditional` GC. `Unconditional` is based on `bst_granger` and `bst_granger_spectral`. `Conditional` is based on the [MVGC Toolbox](https://github.com/lcbarnett/MVGC1)

* GC computed with `bst_granger` and `bst_granger_spectral` is now reported in connectivity files as  `log(var_e_restricted/var_e_full)` Commit: 6651f44

* Connectivity tutorial [page](https://neuroimage.usc.edu/brainstorm/Tutorials/Connectivity) and [script](https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/script/tutorial_connectivity.m), as well as Granger causality [page](https://neuroimage.usc.edu/brainstorm/GrangerCausality) have been updated. Conditional GC is now the recommended approach.

Also closes Issue #419

PR #435, opened 2021-08-31
bugfix

Edouard2laire:
Fix the following error: 

```
Operands to the || and && operators must be convertible to logical scalar values.
Error in panel_record>JumpToEvent (line 1537)
    isFullPage = ~isempty(iShortcut) && strcmpi(RawViewerOptions.Shortcuts(iShortcut,3), 'page') &&
    (size(events(iEvent).times,1) == 2);

Error in panel_record>CreatePanel/ListOccur_ClickCallback (line 372)
                JumpToEvent();
```

It can happen when iShortcut contains mutiple element(eg K>> strcmpi(RawViewerOptions.Shortcuts(iShortcut,3), 'page')  =  2×1 logical array (0  1) )

### DIFF
--- a/toolbox/gui/panel_record.m
+++ b/toolbox/gui/panel_record.m
@@ -1534,7 +1534,7 @@ function JumpToEvent(iEvent, iOccur)
     % Check if event is a "full page" shortcut
     RawViewerOptions = bst_get('RawViewerOptions');
     iShortcut = find(strcmpi(RawViewerOptions.Shortcuts(:,2), events(iEvent).label));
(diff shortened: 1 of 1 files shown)


ftadel:
I completed this fix with this other one: https://github.com/brainstorm-tools/brainstorm3/commit/764d4eaa30b52fd31c8fa80e063b04bc931e40f8

If shortcuts "1" and "2" create events with the same name (1=simple, 2=page), pressing "1" and then "2" led to an error.
Now the output does not make much sense, but at least it doesn't crash...

ftadel:
Actually, having multiple shortcuts creating the same event name with different types can only lead to issues.
I excluded this completely: https://github.com/brainstorm-tools/brainstorm3/commit/49bac1f3245f38c5ce83c12d5a7a3063dee4f0ad
(it is still possible to have multiple shortcuts for creating the same extended event with different durations)

PR #436, opened 2021-09-06
Update export_channel.m

iPsych:
Resolve error due to calling FileFormat before it is declared when using script.


ftadel:
Implemented here, as part of more changes I was currently working on: 
https://github.com/brainstorm-tools/brainstorm3/commit/5a55c6e0618df18b36dfe4f94b9b373551169312

PR #437, opened 2021-09-08
Optimize coherence AxB for full source maps

ftadel:


ftadel:
@rcassani  Are there still think we need to merge from this PR?

PR #438, opened 2021-09-15
Bugfix: don't flip source sign for NIRS

Edouard2laire:
Hello, 

For nirs it is better to not do anything on the source map when plotting the average of a scout.
I am not sure this is the best way to make sure the file is a nirs map, but it seems to work. 

Here is an example of a signal before and after the change: 
before:
before
after:

after

also, do you think it could be possible to make sure when plotting this kind of figure that HbO is red and HbR is blue? 
Edouard


ftadel:
Implemented in this commit: https://github.com/brainstorm-tools/brainstorm3/commit/bb1ffb8a5dda1b56ccf19bb0dc3b8bbd8670c367

Edouard2laire:
Thank you :)

PR #440, opened 2021-09-22
Add MCXlab as plugin

Edouard2laire:
Hello @ftadel, 
This PR is adding Mcxlab to the plugin system. MCXlab is a toolbox used to compute the forward model in nirs. It would be very helpful to include it in brainstorm as it will facilitate the installation for the current users using Matlab and allow users without Matlab to use it from the compile version ( not possible currently)

Let me know if you have questions
Edouard

### DIFF
--- a/toolbox/core/bst_plugin.m
+++ b/toolbox/core/bst_plugin.m
@@ -157,7 +157,7 @@ function PlugDesc = GetSupported(SelPlug)
     PlugDesc(end+1)              = GetStruct('mcxlab');
     PlugDesc(end).Version        = '2020';
     PlugDesc(end).Category       = 'Forward';
-    PlugDesc(end).AutoUpdate     = 1;
+    PlugDesc(end).AutoUpdate     = 2;
     switch(OsType)
         case 'linux64'
             PlugDesc(end).URLzip   = 'http://mcx.space/nightly/release/v2020/mcxlab-linux-x86_64-v2020.zip';
--- a/toolbox/core/bst_plugin.m
+++ b/toolbox/core/bst_plugin.m
@@ -153,6 +153,25 @@ function PlugDesc = GetSupported(SelPlug)
     PlugDesc(end).CompiledStatus = 1;
     PlugDesc(end).LoadFolders    = {'bin'};
     
+    
+    PlugDesc(end+1)              = GetStruct('mcxlab');
+    PlugDesc(end).Version        = '2020';
+    PlugDesc(end).Category       = 'Forward';
+    PlugDesc(end).AutoUpdate     = 1;
+    switch(OsType)
+        case 'linux64'
+            PlugDesc(end).URLzip   = 'http://mcx.space/nightly/release/v2020/mcxlab-linux-x86_64-v2020.zip';
+        case 'mac64'
+            PlugDesc(end).URLzip   = 'http://mcx.space/nightly/release/v2020/mcxlab-osx-x86_64-v2020.zip';
+        case 'win64'
+            PlugDesc(end).URLzip   = 'http://mcx.space/nightly/release/v2020/mcxlab-win-x86_64-v2020.zip';
+            
+    end
+    PlugDesc(end).TestFile       = 'mcxlab.m';
+    PlugDesc(end).URLinfo        = 'http://mcx.space/wiki/';
+    PlugDesc(end).CompiledStatus = 1;
+    PlugDesc(end).LoadFolders    = {'*'};
+    
     % === INVERSE: BRAINENTROPY ===
     PlugDesc(end+1)              = GetStruct('brainentropy');
     PlugDesc(end).Version        = 'github-master';
--- a/toolbox/core/bst_plugin.m
+++ b/toolbox/core/bst_plugin.m
@@ -171,7 +171,7 @@ function PlugDesc = GetSupported(SelPlug)
     PlugDesc(end).URLinfo        = 'http://mcx.space/wiki/';
     PlugDesc(end).CompiledStatus = 0;
     PlugDesc(end).LoadFolders    = {'*'};
-    PlugDesc(end).UnloadPlugs    = 'mcxlab-cl';
+    PlugDesc(end).UnloadPlugs    = {'mcxlab-cl'};
     
         
     PlugDesc(end+1)              = GetStruct('mcxlab-cl');
@@ -191,7 +191,7 @@ function PlugDesc = GetSupported(SelPlug)
     PlugDesc(end).URLinfo        = 'http://mcx.space/wiki/';
     PlugDesc(end).CompiledStatus = 1;
     PlugDesc(end).LoadFolders    = {'*'};
-    PlugDesc(end).UnloadPlugs    = 'mcxlab-cuda';
+    PlugDesc(end).UnloadPlugs    = {'mcxlab-cuda'};
     
     
     % === INVERSE: BRAINENTROPY ===
--- a/toolbox/core/bst_plugin.m
+++ b/toolbox/core/bst_plugin.m
@@ -154,7 +154,7 @@ function PlugDesc = GetSupported(SelPlug)
     PlugDesc(end).LoadFolders    = {'bin'};
     
     
-    PlugDesc(end+1)              = GetStruct('mcxlab');
+    PlugDesc(end+1)              = GetStruct('mcxlab-cuda');
     PlugDesc(end).Version        = '2020';
     PlugDesc(end).Category       = 'Forward';
     PlugDesc(end).AutoUpdate     = 2;
@@ -169,8 +169,30 @@ function PlugDesc = GetSupported(SelPlug)
     end
     PlugDesc(end).TestFile       = 'mcxlab.m';
     PlugDesc(end).URLinfo        = 'http://mcx.space/wiki/';
+    PlugDesc(end).CompiledStatus = 0;
+    PlugDesc(end).LoadFolders    = {'*'};
+    PlugDesc(end).UnloadPlugs    = 'mcxlab-cl';
+    
+        
+    PlugDesc(end+1)              = GetStruct('mcxlab-cl');
+    PlugDesc(end).Version        = '2020';
+    PlugDesc(end).Category       = 'Forward';
+    PlugDesc(end).AutoUpdate     = 2;
+    switch(OsType)
+        case 'linux64'
+            PlugDesc(end).URLzip   = 'http://mcx.space/nightly/release/v2020/mcxlabcl-linux-x86_64-v2020.zip';
+        case 'mac64'
+            PlugDesc(end).URLzip   = 'http://mcx.space/nightly/release/v2020/mcxlabcl-osx-x86_64-v2020.zip';
+        case 'win64'
+            PlugDesc(end).URLzip   = 'http://mcx.space/nightly/release/v2020/mcxlabcl-win-x86_64-v2020.zip';
+            
+    end
+    PlugDesc(end).TestFile       = 'mcxlabcl.m';
+    PlugDesc(end).URLinfo        = 'http://mcx.space/wiki/';
     PlugDesc(end).CompiledStatus = 1;
     PlugDesc(end).LoadFolders    = {'*'};
+    PlugDesc(end).UnloadPlugs    = 'mcxlab-cuda';
+    
     
     % === INVERSE: BRAINENTROPY ===
     PlugDesc(end+1)              = GetStruct('brainentropy');
--- a/toolbox/core/bst_plugin.m
+++ b/toolbox/core/bst_plugin.m
@@ -153,47 +153,6 @@ function PlugDesc = GetSupported(SelPlug)
     PlugDesc(end).CompiledStatus = 1;
     PlugDesc(end).LoadFolders    = {'bin'};
     
-    
-    PlugDesc(end+1)              = GetStruct('mcxlab-cuda');
-    PlugDesc(end).Version        = '2020';
-    PlugDesc(end).Category       = 'Forward';
-    PlugDesc(end).AutoUpdate     = 2;
-    switch(OsType)
-        case 'linux64'
-            PlugDesc(end).URLzip   = 'http://mcx.space/nightly/release/v2020/mcxlab-linux-x86_64-v2020.zip';
-        case 'mac64'
-            PlugDesc(end).URLzip   = 'http://mcx.space/nightly/release/v2020/mcxlab-osx-x86_64-v2020.zip';
-        case 'win64'
-            PlugDesc(end).URLzip   = 'http://mcx.space/nightly/release/v2020/mcxlab-win-x86_64-v2020.zip';
-            
-    end
-    PlugDesc(end).TestFile       = 'mcxlab.m';
-    PlugDesc(end).URLinfo        = 'http://mcx.space/wiki/';
-    PlugDesc(end).CompiledStatus = 0;
-    PlugDesc(end).LoadFolders    = {'*'};
-    PlugDesc(end).UnloadPlugs    = {'mcxlab-cl'};
-    
-        
-    PlugDesc(end+1)              = GetStruct('mcxlab-cl');
-    PlugDesc(end).Version        = '2020';
-    PlugDesc(end).Category       = 'Forward';
-    PlugDesc(end).AutoUpdate     = 2;
-    switch(OsType)
-        case 'linux64'
-            PlugDesc(end).URLzip   = 'http://mcx.space/nightly/release/v2020/mcxlabcl-linux-x86_64-v2020.zip';
-        case 'mac64'
-            PlugDesc(end).URLzip   = 'http://mcx.space/nightly/release/v2020/mcxlabcl-osx-x86_64-v2020.zip';
-        case 'win64'
-            PlugDesc(end).URLzip   = 'http://mcx.space/nightly/release/v2020/mcxlabcl-win-x86_64-v2020.zip';
-            
-    end
-    PlugDesc(end).TestFile       = 'mcxlabcl.m';
-    PlugDesc(end).URLinfo        = 'http://mcx.space/wiki/';
-    PlugDesc(end).CompiledStatus = 0;
-    PlugDesc(end).LoadFolders    = {'*'};
-    PlugDesc(end).UnloadPlugs    = {'mcxlab-cuda'};
-    
-    
     % === INVERSE: BRAINENTROPY ===
     PlugDesc(end+1)              = GetStruct('brainentropy');
     PlugDesc(end).Version        = 'github-master';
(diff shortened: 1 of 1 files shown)


ftadel:
> It would be very helpful to include it in brainstorm as it will facilitate the installation for the current users using Matlab

This could work, but then you need to make nirstorm dependent on it. 
In the process that needs MCXlab, at the beginning of the Run() function, add:
```
    % Initialize MCXlab plugin
    [isInstalled, errMsg] = bst_plugin('Install', 'mcxlab');
    if ~isInstalled
        bst_report('Error', sProcess, [], errMsg);
        return;
    end
```

> allow users without Matlab to use it from the compile version ( not possible currently)

This is a bit more complicated than what you propose, mostly because it contains both Matlab and OS-dependent compiled executables. If you want to compile Brainstorm with MCXLab, it requires including the versions for all the .jar, with `CompiledStatus=2`. I'm sure the developers would prepare an all-included package for Brainstorm if we ask for it. But it would add an extra 20Mb to the compiled .jar, while the full application is only 58Mb and I'm trying to keep this small...

OpenMEEG can be included with `CompiledStatus=1` because it contains only compiled executables that are run through system calls. There is no Matlab code that needs to be compiled with the rest of Brainstorm.

I suggest we keep MCXlab as a plugin with  `CompiledStatus=0` for the moment (not included and not available in the compiled version), and we revise this later (whenever someone complains on the forum, for instance :)

Would that work for you?

fangq:
@ftadel, being a file-size sensitive developer like you, I suggest the following to cut down the add-on file sizes:

1. mcxlab packages contain up to 3 mex files - one for matlab, one for octave, one for mcxfilter (a denoisier, matlab only). because brainstorm only supports matlab, so the mex file for octave can be removed; the mcxfilter.mex* file can also be removed because it is not a critical part.
2. mcx.mex* file (~4MB) is already internally compressed on Linux/Windows (using [upx](https://upx.github.io/)) but not Mac. this was caused by a bug in an older version of upx. if you want, I can re-enable upx and make the mex file for Mac half of the size.
3. mcx mex file contains a statically linked cuda library, so that users do not have to install cuda toolkit on their computers. this increases the file size. if you asks users to install cuda libraries separately, then, the file size can be further reduced.

ftadel:
@fangq Thanks for the details!

> mcxlab packages contain up to 3 mex files - one for matlab, one for octave, one for mcxfilter (a denoisier, matlab only). because brainstorm only supports matlab, so the mex file for octave can be removed; the mcxfilter.mex* file can also be removed because it is not a critical part.

So it would be a minimum of 5.5Mb?
[mcxlab.zip](https://github.com/brainstorm-tools/brainstorm3/files/7243089/mcxlab.zip)
This still represents an increase of more than 10% of the size of the executable for a function that would be used only by a few users.

> if you want, I can re-enable upx and make the mex file for Mac half of the size.

This would be all compressed in a .jar file anyway, so it doesn't matter.

> mcx mex file contains a statically linked cuda library, so that users do not have to install cuda toolkit on their computers. this increases the file size. if you asks users to install cuda libraries separately, then, the file size can be further reduced.

Probably not a good idea... adding complexity for the user to save a few Mb is not a good solution.

@Edouard2laire 
Another efficient solution could be to have mcxlab split in two parts:
- The Matlab code as plugin with `CompiledStatus=2`, or directly integrated in the download of the nirstorm plugin
- The mex* binaries downloaded on the fly (`CompiledStatus=1`)

How important would you consider this feature to be included in the compiled version of Brainstorm?
Are you aware of many users using NIRSTORM without a Matlab license?

fangq:
> So it would be a minimum of 5.5Mb? mcxlab.zip

for mcxlab and for 3 platforms, yes, this is about the min size needed, unless you ask users to install cuda as a dependency. 

> This still represents an increase of more than 10% of the size of the executable for a function that would be used only by a few users.

if this size increase is still considered unacceptable, then, you will have to use mcxlabcl - which is the OpenCL version of mcx. the mex file size is much smaller, and can run on both CPU/GPU across vendors, but the downside is 1) about 3x-5x slower on nvidia gpus compared to the CUDA version, and 2) mostly work out-of-box with not-too-old hardware, but may need opencl run-time on older hardware (like Intel's CPU/iGPU).

> This would be all compressed in a .jar file anyway, so it doesn't matter.

upx does more than just compressing the file - a upx+zip-compressed binary can usually be at least 40% smaller than a zipped executable/dll.

Edouard2laire:
> How important would you consider this feature to be included in the compiled version of Brainstorm? Are you aware of many users using NIRSTORM without a Matlab license?

I don't think that represents a lot of users. In that case, we might be able to ask the user to execute MXClab outside of brainstorm and have a separate process to import the results. I need to check if that could be a possibility for the compiled version.  Another solution would be to have a separate compiled version per OS which might also solve the issue of OpenMEEG. But i understand that you would not have to recompile brainstorm 3 times after each updates :)

Also, as @fangq mentioned, it could be great to be able to use either the Cuda or the OpenCL version of MXClab.  @ftadel, Do you think we could let the user choose between the two versions at the time of the installation?

ftadel:
> Another solution would be to have a separate compiled version per OS which might also solve the issue of OpenMEEG. But i understand that you would not have to recompile brainstorm 3 times after each updates :)

Indeed, let's keep this logic of having only one compiled package.
Note that here is no problem with OpenMEEG, as it a stand-alone executable that can be downloaded on-the-fly by the compiled version. It doesn't have to be compiled together with Brainstorm.

I apologize to be so annoying for only 6Mb. If I had not be paying close attention to size over the years, the packaged size of Brainstom would be over 1Gb. It aggregates parts of so many different libraries... 

> @fangq mentioned, it could be great to be able to use either the Cuda or the OpenCL version of MXClab.  @ftadel, Do you think we could let the user choose between the two versions at the time of the installation?

@Edouard2laire 
What we could do is defining them as two different plugins: mcxlab-cuda with `CompiledStatus=0` and mcxlab-cl with `CompiledStatus=1`. In the nirstorm code, if the function you need is not avaialable (`exist('function_name', 'file'`)), it should mean you don't have any of the two install, which also means you are not in a compiled setup, and you can install mcxlab-cuda.

```
% Install/load mcxlab plugin
if ~exist('mcxlab', 'file')
    [isInstalled, errMsg, PlugDesc] = bst_plugin('Install', 'mcxlab-cuda', 0);
    if ~isInstalled
        return;
    end
end
```

Edouard2laire:
> @Edouard2laire What we could do is defining them as two different plugins: mcxlab-cuda with `CompiledStatus=0` and mcxlab-cl with `CompiledStatus=1`. In the nirstorm code, if the function you need is not avaialable (`exist('function_name', 'file'`)), it should mean you don't have any of the two install, which also means you are not in a compiled setup, and you can install mcxlab-cuda.
> 
> ```
> % Install/load mcxlab plugin
> if ~exist('mcxlab', 'file')
>     [isInstalled, errMsg, PlugDesc] = bst_plugin('Install', 'mcxlab-cuda', 0);
>     if ~isInstalled
>         return;
>     end
> end
> ```

Does something like that work for you ?

Edouard2laire:
Hello @fangq 

I tried MCXlab-CL on my Mac; however, I get the following error message (even on exemple) : 

```
- variant name: [Detective MCXCL] compiled with OpenCL version [1]
- compiled with: [RNG] xoroshiro128+ [Seed Length] 4
initializing streams ...	init complete : 0 ms
building kernel with option: -cl-mad-enable -DMCX_USE_NATIVE -DMCX_SRC_PENCIL -DMED_TYPE=1  -DUSE_ATOMIC -DMCX_SAVE_DETECTORS  -DMCX_DO_REFLECTION -DINTERNAL_SOURCE
Kernel build log:
Compile Server Error.
```

here are my GPU's information returned by mcxlabcl('gpuinfo')

```
Platform [0] Name Apple
============ GPU device ID 0 [1 of 1]: HD Graphics 4000  ============
 Device 1 of 1:		HD Graphics 4000
 Compute units   :	16 core(s)
 Global memory   :	1610612736 B
 Local memory    :	65536 B
 Constant memory :	65536 B
 Clock speed     :	1100 MHz
 Vendor name    :	Unknown
 Auto-thread    :	1024
 Auto-block     :	64
============ CPU device ID 1 [1 of 1]: Intel(R) Core(TM) i5-3210M CPU @ 2.50GHz  ============
 Device 2 of 1:		Intel(R) Core(TM) i5-3210M CPU @ 2.50GHz
 Compute units   :	4 core(s)
 Global memory   :	8589934592 B
 Local memory    :	32768 B
 Constant memory :	65536 B
 Clock speed     :	2500 MHz
 Vendor name    :	AppleCPU
 Auto-thread    :	2048
 Auto-block     :	1
```

@ftadel, I started the changes in NIRSTORM here: https://github.com/Nirstorm/nirstorm/pull/180

fangq:
HD Graphics 4000 is too old for OpenCL unfortunately. set `cfg.gpuid=2` to use the cpu may work with the latest intel opencl CPU runtime.

see the bottom of this section for the HD4xxx issue, and the link for intel's cpu runtimes
http://mcx.space/wiki/index.cgi?Workshop/MCX18Preparation/MethodA#Step_5_Run_a_trial_simulation

for newer intel iGPUs, you should use this one
https://github.com/intel/compute-runtime

fangq:
>> upx does more than just compressing the file - a upx+zip-compressed binary can usually be at least 40% smaller than a zipped executable/dll.
> @fangq Is that a lot of work to recompile?

already did for linux/win binaries. for mac, I can create a separate zip file, but let's first decide if you want to include mcxlab or mcxlabcl first.

Edouard2laire:
> @Edouard2laire I'm sorry I haven't followed actively this topic. Have you reached a final decision for the MCXlab integration with Nirstorm?
> 
> Are the two plugins currently described in your mcxlab branch working and useful?

I mainly tried with mcxlab-cuda and its working



> * **mcxlab-cuda**: Not available in the compiled distribution of Brainstorm
>   URL: http://mcx.space/nightly/release/v2020/mcxlab-*-x86_64-v2020.zip)
> * **mcxlab-cl**: Available for download from the compiled distribution of Brainstorm?? The mex files or compiled executables are directly called without any .m wrapper?
>   URL: http://mcx.space/nightly/release/v2020/mcxlabcl-*-x86_64-v2020.zip

There is a specific wrapper for ncxlab-cuda and mcxlab-cl: respectively mcxlab.m and mcxlabcl.m
For me it's ok to push it like that if that's also ok with you and @fangq :)

ftadel:
From what I understand, both plugins `mcxlab-cl` and `mcxlab-cuda` require the execution of Matlab code, therefore they can't be downloaded in the compiled version (all the .m scripts must be compiled together). At the moment, `CompiledStatus=1` for `mcxlab-cl`, which is not going to work.

The possible options are:
- `CompiledStatus=0` => plugin not available in the are simply not available
- `CompiledStatus=2` => plugin compiled together with Brainstorm, but in that case the binaries for all the OS should be included in the same package (not 3 different ones as it is now). This would require some repackaging + optimization of the distribution.

The ideal would be to have at least one with `CompiledStatus=2` (`mcxlab-cl`?) so that Nirstorm can work in the compiled version. This requires some modifications, and if possible some help from @fangq to prepare a super tight all-OS-in-one package.

Edouard2laire:
> From what I understand, both plugins `mcxlab-cl` and `mcxlab-cuda` require the execution of Matlab code, therefore they can't be downloaded in the compiled version (all the .m scripts must be compiled together). At the moment, `CompiledStatus=1` for `mcxlab-cl`, which is not going to work.
> 
> The possible options are:
> 
> * `CompiledStatus=0` => plugin not available in the are simply not available
> * `CompiledStatus=2` => plugin compiled together with Brainstorm, but in that case the binaries for all the OS should be included in the same package (not 3 different ones as it is now). This would require some repackaging + optimization of the distribution.
> 
> The ideal would be to have at least one with `CompiledStatus=2` (`mcxlab-cl`?) so that Nirstorm can work in the compiled version. This requires some modifications, and if possible some help from @fangq to prepare a super tight all-OS-in-one package.


yes, I agree that having at least one version in the compiled version could be great. Do you think we could already merge this PR like this and then discuss how to have Mcxlab in the compiled version in another PR so I can push the changes relative to mcxlab in the NIRSTORM repository?

fangq:
let me know anything i can help - the zip file in this post looks pretty compact, let me know if want me to attach it to a release or a permanent URL on my sever (mcx.space) - and which one you prefer.

Edouard2laire:
> let me know anything i can help - the zip file in this post looks pretty compact, let me know if want me to attach it to a release or a permanent URL on my sever (mcx.space) - and which one you prefer.
> 
> can do the same for mcxlab-cl

Thank you. The main issue for Francois is that there is one zip file per OS whereas, for the compiled version of Brainstorm, everything would need to be in one zip.  So when grouped together, the three zip might be too heavy for Brainstorm.

fangq:
are you saying you will need a lite version of mcxlab zip package for each OS? can do that too - do you also need separate mcxlabcl zip packages too?

Edouard2laire:
> are you saying you will need a lite version of mcxlab zip package for each OS? can do that too - do you also need separate mcxlabcl zip packages too?

I am saying that we need one zip for mcxlab working all OS. (and/or one zip for mcxlab-cl working for all OS)

ftadel:
> let me know anything i can help - the zip file in this post looks pretty compact, let me know if want me to attach it to a release or a permanent URL on my sever (mcx.space) - and which one you prefer.

The ideal would be to have:
- One minimal package for mcxlab-cuda for ALL OS (all in one), with only what could be needed by Brainstorm as a plugin part of the compiled distribution (no example, no doc, as compact as possible), available as a .zip from a permanent URL on your website
- Exactly the same for mcxlab-cl

Thanks! :-)

ftadel:
@Edouard2laire 

> Do you think we could already merge this PR like this and then discuss how to have Mcxlab in the compiled version in another PR so I can push the changes relative to mcxlab in the NIRSTORM repository?

How urgent is this?
Do we wait until we have these minimal packages?

Edouard2laire:
> @Edouard2laire
> 
> > Do you think we could already merge this PR like this and then discuss how to have Mcxlab in the compiled version in another PR so I can push the changes relative to mcxlab in the NIRSTORM repository?
> 
> How urgent is this? Do we wait until we have these minimal packages?

it's not really urgent.

fangq:
I made two sets of "lite" all-in-one packages, one for v2020
http://mcx.space/nightly/release/v2020/lite/
```
mcxlab-allinone-x86_64-v2020.zip
2021-12-04 11:31	5.6M
mcxlabcl-allinone-x86_64-v2020.zip
2021-12-04 11:34	1.9M
```
another one for the latest git version (which is also quite stable) git20211204
http://mcx.space/nightly/release/git20211204/
```
mcxlab-allinone-x86_64-git20211204.zip
2021-12-04 11:04	10M
mcxlabcl-allinone-x86_64-git20211204.zip
2021-12-04 11:05	2.0M
```
the only difference is that the mcxlab package in v2020 is smaller. In both cases, the largest file is the mex file for Mac because I was not able to compress the binary using upx due to this bug https://github.com/upx/upx/issues/404

I tested all packages on Linux (Ubuntu 20.04), but did not test it on Windows/MacOS, please feel free to pick one and test on all OSes to make sure no dependencies are missing.

Another thing - I grouped all helper functions under [mcxlab/utils/](http://mcx.space/nightly/release/git20211204/mcxlab/utils/) (all .m files, small), if you prefer users to be able to independently install mcxlab and mcxlabcl, I can place the utils folder under mcxlabcl (only adds 40KB in zip file size).

Edouard2laire:
Hello, 
Thanks a lot  !!

> Another thing - I grouped all helper functions under mcxlab/utils/ (all .m files, small), if you prefer users to be able to independently install mcxlab and mcxlabcl, I can duplicate the utils folder under mcxlabcl (only adds 40KB in zip file size).

Yes, if possible, that would be better. 
Regarding the two versions, which one do you recommend using?

fangq:
> Yes, if possible, that would be better.

done, the zip files in the above URLs were updated

> Regarding the two versions, which one do you recommend using?

I am ok either way. just a matter of preferences from the upstream (I am still a bit confused whether it will be included in the large package or download on-demand)

Edouard2laire:
> I am still a bit confused whether it will be included in the large package or download on-demand

Both. In the compiled version, it will be included in the large package and updated when Francois recompile Brainstorm and downloaded on-demand in the Matlab version

fangq:
sounds good. I will let you and @ftadel decide - the git version contains some recent bug fixes but requires 3.5MB more space. I will keep both URLs permanent on our server.

I also just tested all packages on a windows machine, everything is working ok. The `mcx.mexw64` file has a dependency of vcomp120.dll, which is often available on most windows - if you don't want this uncertainty, I can add this dll in the package, taking another 130KB space. For mcxcl.mexw64, the OpenCL.dll is needed which comes with the graphics driver.
mcxlab_dependencies

ftadel:
Thank you for these updates!

In order to cover as many use case as possible, I propose using:
- Included in the compiled distribution of Brainstorm (that we want to keep as small and as safe as possible): mcxlabcl-allinone-x86_64-v2020.zip - 1.9Mb
- Downloaded on demand when running Brainstorm from a full licensed Matlab (in that case we want to offer the latest tools, and size does not matter): mcxlab-allinone-x86_64-git20211204.zip

Does it make sense?

> I also just tested all packages on a windows machine, everything is working ok. The mcx.mexw64 file has a dependency of vcomp120.dll, which is often available on most windows - if you don't want this uncertainty, I can add this dll in the package, taking another 130KB space.

I think this is OK using the one from the OS.
@Edouard2laire You will need to include somewhere in the NIRSTORM tutorial that if this dll is missing, the user should install the "Microsoft Visual C++ Redistributable Packages for Visual Studio 2013".

ftadel:
@Edouard2laire 
I recompiled Brainstorm including mcxlab-cl and nirstorm: https://github.com/brainstorm-tools/brainstorm3/commit/b6e6789a798111867db7c36aca2ad67f5cdb3414
Can you please test this version?

Edouard2laire:
I need to update NIRSTORM first. I'll let you know when the required changed has been merged into nirstorm

Edouard2laire:
Hello @fangq, 

i just tried the update on linux and get the following error: '** Error: Line 397: Invalid MEX-file '/NAS/home/edelaire/.brainstorm/plugins/mcxlab-cuda/mcxlab/mcx.mexa64': /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.27' not found (required by /NAS/home/edelaire/.brainstorm/plugins/mcxlab-cuda/mcxlab/mcx.mexa64).' 

is there a lib i should update ?

fangq:
I assume your OS is Ubuntu 16.04?

did you try the v2020 all-in-one package?

the mex files in the git-version were compiled on Ubuntu 18.04 with gcc 7.5, and I tested it on 20.04. The glibc version on your Ubuntu 16.04 is older than the newer LTS releases.

however, if you want to run it on older Ubuntu, v2020 package should work because it was compiled on 16.04.

the only way to make this mex file portable across wide range of gcc/glibc versions is to statically link libgcc and libstdc++ in the binary, which we actually do in the executable version of mcx (about 30% size increase), but doing static linking on the mex file is tricky.

Edouard2laire:
> I assume your OS is Ubuntu 16.04?

yes

> did you try the v2020 all-in-one package?

I just tried on another computer with the same version and it seems to be working with the v2020 all-in-one package.  i'll confirm you once my computation are finished on the computer with the GPu's :)

PR #442, opened 2021-09-24
Plugin MIA 

annesodub:
Hi,  
I am working on the integration of MIA as a BST plugin (great development you have done for integrating plugins!!!). 

I think I am missing one step (mia won't get started on the execution of the function process_mia_export_db). This idea is that the function process_mia_export_db would 1) export data from BST, save them into a format handled by MIA (database folder containing 1 folder per patient + signal) and 2) start MIA main GUI.

Would you please help me on finishing a proper plugin? :-)

Thanks! AnneSo


rcassani:
Hi @annesodub, thank you for your contribution. 

There were two fields missing in the plugin description, added in e9dd5f9e451f09cbfb5137c51528889ab567bd4c.  
Now the plugin can be properly loaded and unloaded.

Regarding [process_mia_export_db.m](https://github.com/MIA-iEEG/mia/blob/fee1782b5b04d5f7a1b669203a7866d6551efdf1/bst_plugin/process_mia_export_db.m#L40), it indicates that the minimum number of files to make the process available is 2.
Is this correct?

annesodub:
Hi @rcassani,  
Thank you for your prompt response. No, the minimum files to process by process_mia_export_db.m, should be 1 per subject (although 1 file per subject would not make much sense...). Thank you for catching that up.

annesodub:
In [process_mia_export_db.m](https://github.com/MIA-iEEG/mia/blob/master/bst_plugin/process_mia_export_db.m) the minimum number of files to make the process available is now **1**.

toolbox/process/bst_avg_files.m
```diff
@@ -518,7 +518,7 @@
     OutChannelFlag  = ones(NbChannels, 1);
     OutChannelFlag(MeanBadChannels) = -1;
 else
-    OutChannelFlag = [];
+    OutChannelFlag = ChannelFlag;
```
ftadel, line 521: I don't get the point of this modification.
In all cases, it is incorrect to use a random ChannelFlag from the last file. 
If you need to change anything in there, it might be the test just above.

It is not directly related to the MIA plugin anyway, so let's discuss it somewhere else.

ftadel:
@annesodub Do you want this to be merged now, or do you need more time to work on your process first?

annesodub:
Hi Francois,
Thank you for asking. The first process is up and running! I will keep on working on bridging more tightly MIA and Brainstorm (e.g. adapt the function which loads the labeling table in MIA such that it works with Brainstorm format). If it is OK with you that I progressively implements these new features (both in MIA and/or as news BST processes) , yes, please merge. 
Thank you for you guidance!!

toolbox/process/bst_avg_files.m
annesodub: OK, I think it was left over from previous work on my branch sorry about that, will remove.

ftadel:
I wrote a few recommendations for the distribution of your code here: https://github.com/MIA-iEEG/mia/issues/23

For the process sub-menu: if you have only one process, maybe you could add it to the `File` submenu (next to Export BIDS dataset), instead of creating an "SEEG" menu just for one process.

Issue #370, opened 2020-12-11
Nirs unit in brainstorm


Edouard2laire:
hello, 

I am writing this issue to report a bug on how units are handled with nirs data. Currently, in nirstorm we are dealing with three types of signal: raw signal I (intensity of light detected by the detector, we usually don't specify the physical unit), dOD which is -log(I/Io) where I0 is a normalization factor, typically the mean over the baseline: this is unitless, we then have the concentration calculated after the beer-lambert law in mol/l and all the subunit (note that the typical hemodynamic signal is in mmol/l (noted mmol.l-1)  or micromole/l (noted mumol.l-1or \mumol.l-1 in nirstorn) and that the most common unit used to display nirs is mmol/l but most processed in nirstorm will process and save then in mumol/l) 

Or currently, raw signal is marked as dOD, dOD is rightfully marked as unitless when having the entire signal, but if you import it into the database to perform block-averaging then it is labelled as mmol/l. Concentration is labelled as mol/l but the data displayed are in mmol/l (see the following capture of brainstorm and what we have when exporting data in Matlab, also you can see that the data cursor in brainstorm rightfully indicate that data are in mmol/l but not the ylabel of the figure)

![Screenshot from 2020-12-11 14-25-27](https://user-images.githubusercontent.com/24530402/101946508-0beef700-3bbd-11eb-82db-de6e55da5d70.png)

You can do tests using the following subject: https://www.dropbox.com/s/mgpabdqaebq2jv9/viz_nirs_Subject01.zip?dl=0 

Best regards, 
Edouard

ftadel:
Hmmm... could you directly tell me what needs to be fixed?

ftadel:
Ping

ftadel:
Is this still an active issue?

ftadel:
I also meant: for consistency, use the same notation everywhere for the same units.
Do not use `mol.l-1` somewhere and `mol/l` somewhere else, it adds to the confusion.

> So i don't understand why this information is not used by the function bst_getunits.

`bst_getunits` should be used only when DisplayUnits is not defined.
DisplayUnits was not used when reporting min/max values in the selected time window, as you reported.
I tried to fix that here: https://github.com/brainstorm-tools/brainstorm3/commit/70e2234d4499b2456bae9bdff259891a4b98fe5b
Can you please check that it works for handling the NIRST units?

ftadel:
[...]

PR #448, opened 2021-10-18
[bugfix] Improve nirs unit management

Edouard2laire:
This is the first fix for #370. It seems to be fixing the issue of attributing mmol unit to doD or raw data.


Edouard2laire:
I think this will soon be ready to be pushed. Just need to do few more test. I also added a function to be able to use SI prefix instead of any power of 10. I think you could use it also to simplify the code for EEG to automatically use pA. :)

ftadel:
Indeed, this would require to carry this along many additional functions...

Is it really not possible to guess the units of NIRS scouts time series?

Edouard2laire:
Ok; This is ready to be pushed for me. Let me know if you have any comments. 

> Is it really not possible to guess the units of NIRS scouts time series?

It might be doable, but I have to admit that I don't like guessing values that are already saved :)

Edouard2laire:
Actually, don't push it.  There is a very annoying bug... 
Most of the signal are displayed correctly, unless if when viewing a signal, you select a channel and use 'View selected channel'. In that case, the get_unit function is called again but the displayed unit doesn't match the unit in the file anymore. 

Let say the signal is in micromol/l and stored as mmol/l. When opening it, it will be changed to micromol/l for display and display unit is changed to  micromol/l; but if viewing a specific channel, Brainstorm seems to re-read the file as mmol/l while thinking it is micromol/l and therefore think the appropriate unit is 'picomol/l'.  (note the scaling doesn't change, only the displayed unit)




Screenshot from 2021-11-01 22-45-18

ftadel:
I did the display units changes blindly: can you please make sure it works for you?

The duration display modification you proposed was not working when displaying short epochs (with timeunits=ms).
I proposed something simpler, this way the time units are always consistent.

Let me know when this is ready for merging.

Edouard2laire:
I tried on a few files; and it seems it's working. Thanks a lot :)

PR #449, opened 2021-10-19
Keep unit when epoching data

Edouard2laire:
This PR is here to be able to keep track of the units when epoching signal

### DIFF
--- a/toolbox/io/in_fread.m
+++ b/toolbox/io/in_fread.m
@@ -77,7 +77,7 @@ else
 end
 
 %% ===== READ RECORDINGS BLOCK =====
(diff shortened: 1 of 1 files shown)


ftadel:
Adding a `load` of `sFile.filename` here is not a clean solution. 
This is not an unusual use case for `sFile.filename` to contain a .mat file.
It would be better to put this in a lower-level function, possibly in_fread.m: maybe add a return parameter DisplayUnits, set it to [] at the top of the function, and set it to something else only when reading a .mat file.

Edouard2laire:
Does something like that works for you ?

PR #453, opened 2021-11-05
[process_stdchan] also standardize the std matrix

Edouard2laire:


ftadel:
Does it work this way?

Edouard2laire:
yes, it's working on the data I have. I was just wondering, however: is the std matrix always number of channels * number of time points?  i think it's more complexe for time-frequency file and it will not work for that case.

ftadel:
This process is designed to handle only "data" files (`InputTypes = {'data'}`).
Therefore the `Std` matrix can't have more than 2 dimensions.

PR #455, opened 2021-11-07
small fixes

Edouard2laire:
Two minor fixes :)


ftadel:
Thanks for catching these!

PR #456, opened 2021-11-10
Update in_channel_nk.m

schoffelen:
There's a typo in the switch box, as discussed on: https://github.com/fieldtrip/fieldtrip/issues/1916
Please check whether my change makes sense.


ftadel:
Thanks for fixing this!

PR #462, opened 2021-11-18
Export report by mail

Edouard2laire:
Hello, 

This PR is here to allow the direct exportation of reports by email. I found this very useful as I like to often lunch very long computations before leaving the office or going in weekend and like to get the results by email :) 

Exemple of email i just recived : 
Capture d’écran 2021-11-17 à 20 48 46

note: bst_sendemail is here to allow sending html content in email. 

Edouard


rcassani:
This is a nice idea, here some comments:
- The send-email functionality would be better a process, thus it could be added at different points in the pipeline. 
- In the process, the user could decide to send only the subject and text, or the full report. 
- The options for `SMTP_Server` `SMTP_Username` (and `SMTP_Password` if required) could also be in Preferences
- What's the difference between `bst_sendmail.m` and `sendmail.m`?

ftadel:
Useful feature, indeed!!

> The send-email functionality would be better a process, thus it could be added at different points in the pipeline.

I think it's correctly designed this way.
It results in a single line added at the end of the script, instead of `bst_report('Open', ReportFile);`.

> The options for SMTP_Server SMTP_Username (and SMTP_Password if required) could also be in Preferences

Not the general Brainstorm preferences, if it's used only by a few users.
But it could be an extra parameters to the function `ExportMail`.
And/or the function `ExportMail` should check whether `SMTP_Server` and `SMTP_Username` are set, and if not, display an error message explaining how to set them.

> What's the difference between bst_sendmail.m and sendmail.m?

https://undocumentedmatlab.com/articles/sending-html-emails-from-matlab

I would rather move `toolbox/misc/bst_sendmail.m` to `external/other/um_sendmail.m` (um for Undocumented Matlab).
=> we are not the authors of this function, and there is at Mathworks copyright in the header...

rcassani:
> I think it's correctly designed this way. It results in a single line added at the end of the script, instead of `bst_report('Open', ReportFile);`.

How about calling it from the Pipeline editor? To notify that the pipeline is over (in case it's running in another computer.

ftadel:
> How about calling it from the Pipeline editor? To notify that the pipeline is over (in case it's running in another computer.

It is not something you would run from the pipeline editor, but a command to control what to do with the reports when executing a full processing script. If the script is sent to another computer for processing, it's the other computer that would send the email.

rcassani:
> It is not something you would run from the pipeline editor, but a command to control what to do with the reports when executing a full processing script. If the script is sent to another computer for processing, it's the other computer that would send the email.

I get that part. I was thinking in the scenario where:
1. The User logs in with graphics (X2Go, NoMachine, RDP, VNC, etc) to a remote computer.
2. The User makes a pipeline with the Brainstorm GUI, and run it.
  (While this can be done with a script, but not all user write scripts.)
3. Then, the User logs off from that remote computer, which is now busy running the pipeline.
4. Then, the User gets notified by email once the pipeline is finished in the remote computer.
  (Thus avoiding logging in to see if the pipeline is over.)

Edouard2laire:
Hello, 

Thanks for the very interesting comments.

> The send-email functionality would be better a process, thus it could be added at different points in the pipeline.

I think both you and Francois have good points. For now, I would keep it this way: only from Matlab code. And then, we can discuss creating a process to wrap up the entire bst_report functions.  This can serve two purposes: - as Raymundo pointed out, most people won't use script and still might need the report functions; - This might be also useful in the compiled version of brainstorm who cannot use Matlab script. 

> The options for SMTP_Server SMTP_Username (and SMTP_Password if required) could also be in Preferences
> 
> Not the general Brainstorm preferences, if it's used only by a few users.
> But it could be an extra parameters to the function ExportMail.
> And/or the function ExportMail should check whether SMTP_Server and SMTP_Username are set, and if not, display an error message explaining how to set them.

I think the Sendmail function already reports clear error messages. If we create a process to send mail from GUI, it might be useful to add something in the preference windows to configure it. Otherwise, we can assume it is already set. 


> I would rather move toolbox/misc/bst_sendmail.m to external/other/um_sendmail.m (um for Undocumented Matlab).
=> we are not the authors of this function, and there is at Mathworks copyright in the header...

ok, will change that :)

juangpc:
Hello, 

I also think this is very useful, I've hacked this functionality into bst and would like to chip in. I only want to help so if this is too off-topic, or adds to much noise, my apologies. 
 
1. I'm understanding that the sender email would be one provided by brainstorm (that getServerAndFrom call in line 66).  If not, you not only have to set username, password, protocol, port, etc... but also have to configure the email server so that it accepts bst itself as a client. Not too difficult, but not automatic in most cases. 

2. If it is Brainstorm sending that report email from an address specifically devoted to that purpose. Given that you have to fill in your username (or email) and password when you start the toolbox, this option could be a check-box. This would be much easier for the user. For the scripting user I would think something like sendEmailReport('yes'); would be enough. The problem is that the server would have information from many users, filenames, results, etc... So the user should have to be warned about this.  

The second option seems to me more appealing since it (i) adds a reason to require username/pass when you startup bst (ii) allows for a closer way to track usage, and most importantly, it would kind of promote bst to the category of Community in the sense that the software itself would connect functionally more with the users revolving around the neuroimage server. This would be great. It would add another reason why to create a user in neuroimage, etc....

Again, sorry for the noise, but this is great.

tmedani:
Hi Guys, 

This will be a great feature, 

I remember at some time I tried to use a function that sends an SMS when the computation is done,
something like this one,

https://www.mathworks.com/matlabcentral/fileexchange/16649-send-text-message-to-cell-phone

 but it has some limitations.

If you think it is worth it, I can check it again for Bst.

ftadel:
I think it would be good to implement both:
1) Send the email from a script: 
`bst_report('SendEmail', ReportFile, isFullReport, to, subject, from=[pref], SMTP_Server=[pref], SMTP_Username=[pref], SMTP_Password=[pref])`
2) Create a process to call this function, that can be added after any process (all the fields can be configured as process options)

But while testing this, I could not manage to send any email using any of my email addresses...
(same errors with both sendmail.m and um_sendmail.m)

With Gmail:
```
Error using sendmail (line 179)
530 5.7.0 Must issue a STARTTLS command first. l7sm11177279wry.86 - gsmtp
```

With McGill/Exchange:
```
Error using um_sendmail (line 189)
451 5.7.3 STARTTLS is required to send mail [PR0P264CA0121.FRAP264.PROD.OUTLOOK.COM]
```

Which SMTP server did you use and version of Matlab did you use in your tests?

Edouard2laire:
Hello. 

For the test i used: 
- smtp: smtp.concordia.ca , email: 'edelaire@perf-imglab07.concordia.ca' 
Note that this email doesn't exist so I didn't fill SMTP_Username, and SMTP_passworld.

rcassani:
Hi,

I tried to send the email with my **McGill** account without success.
With **Gmail**, this is working:
```
myaddress = 'username@gmail.com';  % Gmail address user@gmail.com
mypassword = 'App Pasword'; % This password need to be generated from Gmail  
port = '465';

setpref('Internet','E_mail',myaddress);
setpref('Internet','SMTP_Server','smtp.gmail.com');
setpref('Internet','SMTP_Username',myaddress);
setpref('Internet','SMTP_Password',mypassword);

props = java.lang.System.getProperties;
props.setProperty('mail.smtp.auth','true');
props.setProperty('mail.smtp.socketFactory.class', 'javax.net.ssl.SSLSocketFactory');
props.setProperty('mail.smtp.socketFactory.port', port);

sendmail(myaddress, 'Gmail test from Matlab', 'This is a test message.');
```
As I'm using  2-Step-Verification in Gmail, there it's needed to create an [App Password](https://support.google.com/mail/answer/185833?hl=en).

It seems that there will be way too many different way to configure the output server. Is it an option to use a generic Brainstorm address/server? As it was pointed [above](https://github.com/brainstorm-tools/brainstorm3/pull/462#issuecomment-973188052) by @juangpc.

Edouard2laire:
> It seems that there will be way too many different way to configure the output server. Is it an option to use a generic Brainstorm address/server? As it was pointed [above](https://github.com/brainstorm-tools/brainstorm3/pull/462#issuecomment-973188052) by @juangpc.

I think there should be an easy solution. For example, when using freesurfer, we can use the -mail option to send an email when the segmentation is done.  The email used doesn't exist ( when i used it from my computer, the email I receive is from 'edelaire@perf-imglab07.concordia.ca' even if it doesn't exist).

ftadel:
This sound way too complicated to try and configure a SMTP client on the user side.
I'll see if I can easily write a PHP script on the Brainstorm web server that can send these emails.

ftadel:
Script call:
```
bst_report('Email',   ReportFile, to, subject, isFullReport=1)
```

Process:
image

Email sent, full reports = 1 (HTML):
image

Email sent, full reports = 0 (plain text):
image

Anything else to add?

ftadel:
@Edouard2laire  @rcassani  @juangpc 
Can you please review if this looks good to you?

rcassani:
@ftadel, the PHP script does not send the email if the body (`html`) is empty; is this expected? 
 For example if `GlobalData.ProcessReports` is empty.

ftadel:
> the PHP script does not send the email if the body (html) is empty; is this expected?

This is now fixed. If the email body is empty, it will copy the subject of the email.

ftadel:
@rcassani raised a concern about this PHP script being able to send emails to anybody without restriction.

I changed the logic of the function, which can now only send messages to a registered Brainstorm user + an optional CC email address. The email address of the user is fetched by the PHP script. If the CC address is not the same, the email is sent to the user + the CC address. 

This way, a Brainstorm user will always be notified when an email is sent, but it will be possible to also receive the message to an additional personal email.

I also added an extra message at the bottom of the sent emails: 
`If you did not request this email, please contact: brainstorm@sipi.usc.edu`

image

image

@Moo-Marc FYI

ftadel:
@Edouard2laire @rcassani @juangpc @Moo-Marc @tmedani 
Any last comment before I merge it?

Edouard2laire:
Sounds good to me. 

> @rcassani raised a concern about this PHP script being able to send emails to anybody without restriction.

if you really want to be safe, you might want to add a timer so a user cannot send more than 1 report/minute.

Edouard2laire:
Actually, i was using Matlab2020 when i tested the PR and it seems it not working using matlab2017a: 

```
K>> resp = webwrite('http://neuroimage.usc.edu/bst/send_email.php', 'g', '7gA9b3EW54', 'u', username, 't', to, 's', subject, 'b', html);
Error using webwrite (line 123)
The first argument should be a cell array of strings. 
```
Do you have any idea how we could fix it ? :)

PR #465, opened 2021-11-26
Updates to York Instruments hdf5 data reader

RA-Neuro:
The file format specification for York Instruments hdf5 MEG has been updated, breaking the Brainstorm reader. These edits fix these problems and clear up some other issues like attribute definitions that were unclear when the reader was originally written.


ftadel:
Thank you for posting these updates here.

You did not keep the compatibility with the older file format.
Can you please confirm that you have no interest in maintaining the older files (.hdf5) that were read by the previous version of these functions?

Could you please share an example file for the new file format, so that we can test these functions?
(upload it somewhere and post the download link here)

Thanks

RA-Neuro:
Hi François,

Thanks for looking into this.

No real research data was ever generated using the old specification with .hdf5 suffix, and no other sites produced any data in that format, so it will not be a problem to loose backwards compatibility.

I have emailed you a link to an example data set, please let me know if there is anything else I can do.

Cheers,

Rich

ftadel:
The code looks good.
Thank you for the example file, I could validate that it works as expected.

However, I can't validate whether the compensation weights or sensors orientations are correct.
If you are expecting to do some source localization with Brainstorm, it could be interesting to run some tests using a phantom. It would help checking both the import code and the MEG system itself. You can find two examples we developed for Elekta and CTF MEG systems on the website:
- https://neuroimage.usc.edu/brainstorm/Tutorials/PhantomCtf
- https://neuroimage.usc.edu/brainstorm/Tutorials/PhantomElekta

Note that we don't recommend using the digitizers in continuous mode, as it seems to be the case in this example file:
image

For doing some automated registration between the digitized head shape and the head surface reconstructed from the MRI, we think it is much more convenient to have a homogeneous sampling of the head surface:
https://neuroimage.usc.edu/brainstorm/Tutorials/ChannelFile#Automatic_registration

PR #470, opened 2021-12-06
Bids nirs

### LINKED ISSUE #469: BIDS for NIRS
Hello, 

Recently, BIDS started to be implemented for NIRS (mainly using the snirf format, already supported by Brainstorm). I was wondering if you could guide me on what part of the code modify to import  BIDS dataset containing nirs data such as :  https://github.com/rob-luke/BIDS-NIRS-Tapping/tree/388d2cdc3ae831fc767e06d9b77298e9c5cd307b ?

Regards, 
Edouard

Edouard2laire:
First implementation of BIDS for NIRS. (fix #469)
- Specification can be found here: https://github.com/bids-standard/bids-specification/blob/cbf34ac830d3eb2b917b410747169b41bd05f8e1/src/04-modality-specific-files/10-functional-near-infrared-spectroscopy.md
- Exemple of dataset:  https://github.com/rob-luke/BIDS-NIRS-Tapping/tree/388d2cdc3ae831fc767e06d9b77298e9c5cd307b

This is almost ready. I am only strugling to read the coordinate of the optodes. We need to read '*_coordsystem
.json' and '*_optodes.tsv'

corrdsystem looks like: 
```
{
    "NIRSCoordinateSystem": "CapTrak",
    "NIRSCoordinateSystemDescription": "The X-axis goes from the left preauricular point (LPA) through the right preauricular point (RPA). The Y-axis goes orthogonally to the X-axis through the nasion (NAS). The Z-axis goes orthogonally to the XY-plane through the vertex of the head. This corresponds to a \"RAS\" orientation with the origin of the coordinate system approximately between the ears. See Appendix VIII in the BIDS specification.",
    "NIRSCoordinateUnits": "m"
}
```
and optodes looks like 
```
﻿name	type	x	y	z
S1	source	-0.041613204679326624	0.026799775287857947	0.1299043936308115
D1	detector	-0.037619588707178915	0.06322851630256272	0.11572802770110814
```

What i think need to be modified: 
-  https://github.com/Edouard2laire/brainstorm3/blob/BIDS-nirs/toolbox/process/functions/process_import_bids.m#L597 : currently electrodesSpace is UNKNOWN; guess should be either ORIG or MNI
- Modify process_channel_addloc to work with nirs. optodes.tsv contains the coordinate of all sensors; strucutre in matlab contains the channels (combination of a source and detector). So when reading for exemple coordinate of S1 we need to modify the coordinate of all channels containing S1. 

edit: actually i don't understand the subfunction SelectCoordSystem

Regards, 
Edouard


ftadel:
> edit: actually i don't understand the subfunction SelectCoordSystem

It selects which _electrode.tsv (or _optode.tsv) should be read as the coordinates for the Brainstorm channel files.
Only "MNI" and "Other" (related to the associate MRI) are supported.
By order of preference: Other (subject space), and if not available MNI (requires a non-linear normalization procedure).

As in this example: https://github.com/bids-standard/bids-examples/tree/master/ieeg_epilepsy_ecog/sub-ecog01/ses-postimp/ieeg

How do you expect to be able to convert from the coordinates in these files into Brainstorm SCS coordinates?
Is it at all possible?

Edouard2laire:
I guess it's possible: the coordinate are also saved in the .snirf file along with the fiducials coordinate and when read fron there it's aligned with the anatomy : https://github.com/brainstorm-tools/brainstorm3/blob/d5a0885e92b54c89cce65e5306d5813b56d4969c/toolbox/io/in_data_snirf.m#L83-L111


We need to have similar code but reading from the otpodes file i guess.

toolbox/io/in_data_snirf.m
```diff
@@ -31,7 +31,7 @@
 
 %% ===== CHANNEL FILE ====
 % Get scaling units
-switch strtrim(str_remove_spec_chars(jnirs.nirs.metaDataTags.LengthUnit(:)'))
+switch strtrim(str_remove_spec_chars(jnirs.nirs.metaDataTags.LengthUnit{:}'))
```
ftadel, line 34: This can't work: there is only one parameter in input to `str_remove_spec_chars`, and this syntax `{:}` would call the function with multiple parameters if there are multiple values.
Plus, using `switch` with multiple values won't work.

If `LengthUnit` is a cell array with one string only, you can use `{1}` instead.

toolbox/io/in_data_snirf.m
```diff
@@ -200,7 +215,7 @@
 
 for iEvt = 1:length(jnirs.nirs.stim)
     
-    DataMat.Events(iEvt).label      = strtrim(str_remove_spec_chars(jnirs.nirs.stim(iEvt).name(:)'));
+    DataMat.Events(iEvt).label      = strtrim(str_remove_spec_chars(jnirs.nirs.stim(iEvt).name{:}));
```
ftadel, line 218: Same comment as before.
In `.label`, you want have only one string, not multiple values.
Therefore you might want to use `{1}` instead of `{:}`.

toolbox/io/in_tsv.m
```diff
@@ -41,6 +41,9 @@
     
 % Read header
 tsvHeader = str_split(fgetl(fid), Delimiter);
+if tsvHeader{1}(1) == char(65279) % WEIRD
```
ftadel, line 44: What is this?

(double problem: it conflicts with the recent edits I made to in_tsv.m)

toolbox/process/functions/process_import_bids.m
```diff
@@ -555,14 +555,21 @@
                         if isfield(sCoordsystem, 'iEEGCoordinateUnits') && ~isempty(sCoordsystem.iEEGCoordinateUnits) && ismember(sCoordsystem.iEEGCoordinateUnits, {'mm','cm','m'})
                             posUnits = sCoordsystem.iEEGCoordinateUnits;
                         end
+                        if isfield(sCoordsystem, 'NIRSCoordinateUnits') && ~isempty(sCoordsystem.NIRSCoordinateUnits) && ismember(sCoordsystem.NIRSCoordinateUnits, {'mm','cm','m'})
```
ftadel, line 558: Combine this with the previous test with and `elseif`.
It's either one OR the other.

ftadel (changes requested):

ftadel:
(sorry for the delay...)

Edouard2laire:
Thanks a lot for the comments. However I won’t have time to have a look before January. 

Happy Christmas :)

ftadel:
Any update?

Edouard2laire:
Sorry, i totally forgot about it. i will try to come back to it later this week. I am still not very sure about how to deal with the coordinate system.

Edouard2laire:
closing the PR for now. Will reopen when I have time to work on it

PR #471, opened 2021-12-13
fix sFileIn variable undefined

juangpc:
When importing a bids dataset there is a variable in ```io/import_raw.m``` named ```sFileIn``` which is undefined and therefore an error is thrown.

I think ```sFileIn``` should be ```sFile```.

See: https://github.com/brainstorm-tools/brainstorm3/blob/822dab0d66f8dbab0685f491b34f1b946095a4d7/toolbox/io/import_raw.m#L315


ftadel:
Thanks for the live debugging :)

PR #472, opened 2021-12-13
Support load FilesB with Kernel for coherence

rcassani:


rcassani:
@ftadel, the change of `bst_connectivity` to be a loop over files for coherence (and other metircs) will come in a future PR.

ftadel:
Do you need me to test all this (which might take me a little while), or are you quite confident this is working?
If you think this is OK, please go ahead a merge it.

PR #478, opened 2021-12-21
Fix PSD averaging with bad segments

Moo-Marc:
Fixes online average and variance computation when skipping bad segments.


ftadel:
Good catch, thanks!

PR #486, opened 2022-01-24
[bugfix, cat12 segmentation] Keep the correct MRI before segmenting

### LINKED ISSUE #485: Sgmenting MRI - Brainstorm is not keeping the right MRI
Hello, 

if you try to segment subject anatomy with multiple MRIs, brainstorm will delete all MRIs except the one named MRI (and not keep the one selected as default MRI). 

To reproduce; import an MRI; resample it; try to launch cat12 on the resampled MRI. It will delete the file and then crash

Edouard

Edouard2laire:
fix #485 

Why is cat12 the only segmentation tool for which we remove all MRI before launching the segmentation? 
 it would be great to only remove existing surfaces and not MRIs (for example if you run cat12 on a resampled MRI; it's always good to keep the original MRI)

### DIFF
--- a/toolbox/db/db_delete_anatomy.m
+++ b/toolbox/db/db_delete_anatomy.m
@@ -37,8 +37,8 @@ if ~isKeepMri && ~isempty(sSubject.Anatomy)
     sSubject.Anatomy(1:end) = [];
     sSubject.iAnatomy = [];
 elseif isKeepMri && (length(sSubject.Anatomy) >= 2)
-    ind =  ones(1,length(length(sSubject.Anatomy)));
-    ind(sSubject.iAnatomy) = 0;
+    ind =  true(1,length(sSubject.Anatomy));
+    ind(sSubject.iAnatomy) = false;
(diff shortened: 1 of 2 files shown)


ftadel:
You are completely right: 
- We should warn the user that surfaces will be deleted before the computation (so that we don't get stuck before the end with a message box OR delete the files without the user permission), but we do not need to actually do it before the computation is done.  => I added the question at the beginning of the process for all methods, do not delete them right away, and call the import function in non-interactive mode so the surfaces are deleted immediately.
- There is not reason for deleting all the other MRIs: now with isKeepMri=1, it will keep them all
- And all segmentation methods should work the same.

This is implemented here:
- https://github.com/brainstorm-tools/brainstorm3/commit/c9cb9f4c35ab39e1ef4df55e00469eda94f55892
- https://github.com/brainstorm-tools/brainstorm3/commit/2d13b755aaf24d6b8c8a7871475dc87e838115e9

Let me know if there is any problem with this
Thanks for reporting these issues!

Edouard2laire:
looks good. thanks :)

PR #488, opened 2022-01-25
Update coherence 2021 scripts

rcassani:
- Improve computation of performance in bst_cohn_2021.m, specially for cases using ImagingKernel
- Update on processes GUI


rcassani:
@ftadel, what do you think of hiding the grayed option **Imaginary Coherence (before 2019)** from the GUI in the coherence 2021 processes?

toolbox/connectivity/bst_connectivity.m
```diff
@@ -322,8 +322,8 @@
                 OPTIONS.Freqs(iZero) = [];
                 R(:,:,iZero) = [];
             end
-            % Keep only the frequency bins we are interested in
-            if ~isempty(OPTIONS.MaxFreq) && (OPTIONS.MaxFreq ~= 0)
+            % Keep only the frequency bins we are interested in. For deprecated coherence
+            if ~isempty(OPTIONS.MaxFreq) && (OPTIONS.MaxFreq ~= 0) && ~isempty(OPTIONS.MaxFreqRes)
```
ftadel, line 326: I don't understand this part.
The option MaxFreq exists in the new coherence processes too, and need to be handled here.

ftadel (changes requested):

ftadel:
> @ftadel, what do you think of hiding the grayed option Imaginary Coherence (before 2019) from the GUI in the coherence 2021 processes?

I think it's better to keep it for now.
Users who started a study with the old measure (which is not that old, considering how long a study can last...) may need to be able to reproduce results that they've obtained before, or when including new participants in an existing study they would need to compute the same measure as before.

It's clearly visible that we don't recommend using it, therefore I don't think it adds confusion or a major readability issue to the processs options.

toolbox/connectivity/bst_connectivity.m
rcassani: For the new coherence, the MaxFreq is handle inside bst_cohn_2021.m 020d57d74c2d142247ac9803e0c258758ee9ef2e
The this part may be needed only for the deprecated coherence.

ftadel: Ok, sorry for missing that.

Issue #489, opened 2022-01-27
.plx files show linear reading delay based on time-segment selection


mpompolas:
The loading time of the `.plx` file is affected by the relative position (in time) of the segment requested.
This is a `Plexon SDK` issue, not a `Brainstorm` issue.

So for loading 1 second segment from a later stage of the recording takes way more time than a 1 second segment at the start of the recording.

See example loading times for 1 second of all 33 channels at different recording stages:

![image](https://user-images.githubusercontent.com/23224563/151263243-b1ec85c1-f31a-4278-a92c-7846df48eca8.png)
The printed output that summarizes everything is:
```
Time segment [0.0001-1] took 0.11039 seconds
Time segment [80.0001-81] took 4.8363 seconds
Time segment [160.0001-161] took 9.7937 seconds
Time segment [240.0001-241] took 14.778 seconds
Time segment [320.0001-321] took 19.4196 seconds
Time segment [400.0001-401] took 24.7371 seconds
Time segment [480.0001-481] took 29.2971 seconds
```

This makes our browsing purposes almost unusable. 30 seconds wait time to display 1 second of recording will not work out.

I've encountered this in the past with an older of an importer (I think it was Tucker Davis, maybe I'm wrong). The entire recording up to the last sample requested was loaded, and then the matrix was chopped to match the requested segment.

The linear character of the wait times is probably an indicator that the same thing is happening here. Memory monitoring while calling the different segments can reveal if that's the case.

Here is the code for replication of the figure:

```
filename ='C:\Users\knasi\Consulting\Brainstorm\new tutorial files\ytu288c-01.plx';  % fs =10kHz
iSelectedChannels = 1:33;

all_measured_times = zeros(7,1);
all_time_boundaries = zeros(7,2);
all_labels = cell(7,1);

ii = 1;
for iSample = 1:800000:5000000
    SamplesBounds = [(iSample-1)+1,iSample + 9999];  % Read only 1 second since Fs = 10000;
    tic
    for iChannel = 1:length(iSelectedChannels)   
        [adfreq, n, data] = plx_ad_span_v(filename, iSelectedChannels(iChannel)-1, SamplesBounds(1), SamplesBounds(2));
    end
    
    % Gather values
    all_time_boundaries(ii,:) = [SamplesBounds(1)/10000, SamplesBounds(2)/10000];
    all_measured_times(ii) = toc;
[...]

PR #490, opened 2022-01-27
bug fix - .plx unsorted spikes are now ignored

mpompolas:
Bug fix: Readjusted boundaries in `in_fread_plexon` to avoid inconsistencies with spiking and acquisition events.
Added: Added progress bars for better transparency on what is happening under the hood to the users

### DIFF
--- a/toolbox/io/in_fopen_plexon.m
+++ b/toolbox/io/in_fopen_plexon.m
@@ -54,6 +54,8 @@ Comment = rawFile;
 
 
 %% ===== READ DATA HEADERS =====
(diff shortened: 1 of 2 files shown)


ftadel (changes requested):
Maybe you can include the changes I suggested in the issue https://github.com/brainstorm-tools/brainstorm3/issues/489 in this PR as well?

ftadel:
It looks like all the changes proposed here are reported in PR: https://github.com/brainstorm-tools/brainstorm3/pull/516

PR #493, opened 2022-01-29
Allow the difference of two scouts

Edouard2laire:
hello Francois, 

This PR allows us to also take the difference between two scouts. This can be useful when you create scouts manually and want to remove a large portion you mistakenly added. 

See example, I have a scout 1 and want to remove a portion of it. I created a new scout 2 covering the part I want to remove and then use the difference. 

Capture d’écran 2022-01-28 à 20 45 27

For future reference; to do it from curent Brainstorm version; export scout A as A; amd B as B then execute: 
```
sNewScout = A;
sNewScout.Vertices = setdiff(A.Vertices,B.Vertices );
sNewScout.Label = [A.Label ' - ' B.Label];
```

and then import sNewScout

Another function that has been asked to me by another member of my lab would be to be able to transfer a scout from one atlas to another. (from example from Yeo7network to user Scouts). Can we do it? 

Regards, 
Edouard

### DIFF
--- a/toolbox/gui/panel_scout.m
+++ b/toolbox/gui/panel_scout.m
@@ -32,6 +32,7 @@ function varargout = panel_scout(varargin)
 %                       panel_scout('RemoveScouts', iScouts) : remove a list of scouts
 %                       panel_scout('RemoveScouts', )        : remove the scouts selected in the JList 
 %                       panel_scout('JoinScouts')
+%                       panel_scout('DifferenceScouts')
 %                       panel_scout('UpdateScoutsVertices', SurfaceFile)
 %                       panel_scout('SaveScouts')
 %                       panel_scout('LoadScouts')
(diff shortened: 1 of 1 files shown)


ftadel:
I'm think this use case is sufficiently significant to justify adding one extra menu.
It doesn't seem to correspond to any routine manipulation, but only error handling, and there is already a clear way to handle this.

In the example figure you propose, I think it would be faster and less cumbersome to remove the vertices manually, rather than creating a new scout and compute a subtraction:
- Click on the "Add vertex" button
- Click on the scout you want to edit
- SHIFT+Click on all the vertices you want to remove (zoom in and display the edges for easier localization of the vertices)

Less buttons and menus = increased usability :)

Edouard2laire:
> In the example figure you propose, I think it would be faster and less cumbersome to remove the vertices manually, rather than creating a new scout and compute a subtraction:


less cumbersome, yes. Faster no. I am proposing this specifically after having seen a colleague spending literally more than 40 minutes removing manually vertex from a scout. When I saw it, she was only half done with the task. Removing the other half with the difference tool less than a minute.  Also, when removing the vertex manually, there is always a risk to leave a few vertices  behind (especially if hidden in the gyrus)

ftadel:
I don't think this is an intuitive feature: I don't think most users needing to remove vertices are going to naturally go towards this kind of solution. It would require specific training, or a section in the introduction tutorials.
But since this does not correspond to any standard use case, I don't want to add a section dedicated to this in the tutorials.

Anyway, now it's done, I'll review it and add it :)

> Another function that has been asked to me by another member of my lab would be to be able to transfer a scout from one atlas to another. (from example from Yeo7network to user Scouts). Can we do it?

This is easy with menus `Export to Matlab` / `Import from Matlab`.

Edouard2laire:
ok, closing the PR then :) 
I'll suggest doing export to Matlab / do the setdiff there and then reimport

ftadel:
Arghhh... I was committing the changes and you closed the PR half way :)

Edouard2laire:
> Arghhh... I was committing the changes and you closed the PR half way :)

Sorry, i thought you rejected the PR :) Thanks for merging it  !!

PR #494, opened 2022-01-29
Use FSL to compute head mask

Edouard2laire:
Hello, 

**Object of the PR**: This PR allows us to use FSL to compute the head mask using FSL bet. Fsl is also creating a binary mask that can be used to mask the MRI, improving the latter segmentation. 

**Example:** 
Screenshot from 2022-01-29 16-48-04

Overlay of the MRI and the mask (in red). 

and after applying the mask: 
Screenshot from 2022-01-29 16-49-55

**Issues:** 

N1: Applying the mask inside the process doesn't work. However, importing the mask inside brainstorm; then exporting both mask and MRI to Matlab; applying the mask from Matlab; before re-importing the new MRI works : 
```
>> mri.Cube = mri.Cube .* mask.Cube; 
>> mri.Comment = [ mri.Comment 'matlab'];
```
N2: The type of mri.Cube doesn't seems to be consistent : sometimes uint8, sometimes single, or double. So mri.Cube = mri.Cube .* mask.Cube;  doesn't work sometimes. Is there a rule for that ? 

N3:  Skin surface is created as Subject16_JH_brain_outskin_mesh.nii.gz  (1 for skin / 0 otherweise) : 

Screenshot from 2022-01-29 17-12-19 
How can we create a skin mesh from that file ? currently, I am importing the MRI and then use brainstorm to generate the head :) 

Thanks a lot for your help
Edouard


ftadel:
I don't have FSL installed on my WSL, and didn't want to start doing this now just for BET.
=> I haven't tested any of what I committed here, so maybe some of it doesn't work, apologies... But I'm sure you'll get the idea of the changes and will know how to fix all my errors :)

Changing the Index to 39 requires moving **process_ft_prepare_mesh_hexa to  40**.

> N2: The type of mri.Cube doesn't seems to be consistent : sometimes uint8, sometimes single, or double. So mri.Cube = mri.Cube .* mask.Cube; doesn't work sometimes. Is there a rule for that ?

No rules. Your code needs to be capable of handling different data types.

You could maybe add something like this, in order to force the mask to be in the same type:
```
switch class(sMri.Cube)
    case 'uint8', uint8(sMask.Cube);
    case 'int16',  uint16(sMask.Cube);
    case 'int32',  uint32(sMask.Cube);
    case 'single',  single(sMask.Cube);
    case 'double', double(sMask.Cube);
end
```

> How can we create a skin mesh from that file ? 

You can simply run tess_isohead on the masked MRI. You probably don't need this skin file.
Otherwise, you could add within your process only the central calls of tess_isohead: read the file, call isosurface and reducepatch, and add the resulting surface to the database.

Edouard2laire:
Thanks, it seems to be working. 
I used a small trick to avoid having the switch/case :) I'll try on few more example and let you know when it's ready for merging

Edouard2laire:
ok; it's ready for review :)

ftadel:
Can you check my changes and make sure they don't break anything on your end?
(you might need to adjust your FSL install to define $FSLDIR)

When you're OK, I merge.

ftadel:
What about this simple fix?

Edouard2laire:
ok, it's ready to be merged. I moved the menu one last time to group it with Brainstorm head surface generation. You can discard it if you want.

ftadel:
>  I moved the menu one last time to group it with Brainstorm head surface generation. You can discard it if you want.

Sorry to insist: I prefer not having it listed in the middle of the recommended pipelines.
The use cases for this process are much more limited than the head/BEM generation, and it requires an external program that few users would have access to.

Issue #495, opened 2022-02-01
Bug in robust_ap_fit when using knee parameter


neurofractal:
When using a knee, the robust_ap_fit doesn't mask any frequencies or values in the power spectrum. 

PR fix coming soon. 

Will also do a PR fix for Fieldtrip

PR #496, opened 2022-02-01
Fix for robust_ap_fit

neurofractal:
Bug specific for knee option. When doing second line fit, exclude certain frequencies and power values.

#495


ftadel:
@lucwilson Should I merge this before or after the other PR?

lucwilson:
@ftadel This is safe to merge immediately. I'll update my branch to not conflict when its eventually merged. 

@neurofractal great catch, thank you for your help!

PR #497, opened 2022-02-02
Connectivity: Added wPLI and ciPLV

ftadel:
From forum requests:
- https://neuroimage.usc.edu/forums/t/weighted-phase-lag-index/9434/8
- https://neuroimage.usc.edu/forums/t/new-method-corrected-imaginary-part-of-phase-locking-value-ciplv/9751/17


ftadel:
These measures have issues:
- wPLI: very suspicious results
- ciPLV: it can return some NaN values, which breaks the display functions
This must be handled in bst_connectivity

@danielemarinazzo Can you please double-check what I did?

danielemarinazzo:
wPLI can be left out for the moment, indeed I had left it out from my PR, I had commented on the forum (will edit there), but I hadn't followed up with it, after seeing that there's an exact relationship with PLV. Gonna work on it later.

ciPLV we tested it on an extensive dataset and got no NaNs, still we should find a way to deal with it, not sure how you do it in brainstorm, if you send me an example (better starting with DataAband and DataBband structures, so I can run it from the command line directly) I can have a look.

toolbox/connectivity/bst_connectivity.m
```diff
@@ -537,23 +553,21 @@
...
-                %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+                switch (OPTIONS.Method)
+                    case 'plvt'
+                        R(:,:,iBand) = exp(1i * angle(HA(iA,:)./HB(iB,:)));
+                        Comment = 'PLVt: ';
+                    case 'ciplvt'
+                        R(:,:,iBand) = (imag(exp(1i * angle(HA(iA,:)./HB(iB,:)))))./sqrt(1-(real(exp(1i * angle(HA(iA,:)./HB(iB,:))))/nTime).^2); % Proposed by Daniele Marinazzo
+                        Comment = 'ciPLVt: ';
+                    case 'wplit'
+                        R(:,:,iBand) = abs(mean(sin(angle(HA)-angle(HB))))./mean(abs(sin(angle(HA)-angle(HB))));  % Proposed by Daniele Marinazzo
```
danielemarinazzo, line 568: R(:,:,iBand) = abs(sin(angle(HA)-angle(HB)))./(sin(angle(HA)-angle(HB)));

toolbox/connectivity/bst_connectivity.m
```diff
@@ -501,17 +508,26 @@
...
+                % Compute PLV (Divided by number of time samples)
+                switch (OPTIONS.Method)
+                    case 'plv'
+                        R(:,:,iBand) = (phaseA*phaseB') / size(HA,2);
+                        Comment = 'PLV: ';
+                    case 'ciplv'
+                        R(:,:,iBand) = (imag((phaseA*phaseB') / size(HA,2)))./sqrt(1-(real((phaseA*phaseB') / size(HA,2))).^2);  % Proposed by Daniele Marinazzo
+                        Comment = 'ciPLV: ';
+                    case 'wpli'
+                        R(:,:,iBand) = abs(mean(sin(angle(HA(iA))-angle(HB(iB)))))./mean(abs(sin(angle(HA(iA))-angle(HB(iB))))); % Proposed by Daniele Marinazzo
```
danielemarinazzo, line 520: R(:,:,iBand) = abs(mean(sin(angle(HA)-angle(HB))))./mean(abs(sin(angle(HA)-angle(HB))));

danielemarinazzo, line 520: R(:,:,iBand) = abs(mean(sin(angle(HA(iA,:)')-angle(HB(iB,:)'))))./mean(abs(sin(angle(HA(iA,:)')-angle(HB(iB,:)'))));

toolbox/connectivity/bst_connectivity.m
```diff
@@ -537,23 +553,21 @@
...
-                %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
+                switch (OPTIONS.Method)
+                    case 'plvt'
+                        R(:,:,iBand) = exp(1i * angle(HA(iA,:)./HB(iB,:)));
+                        Comment = 'PLVt: ';
+                    case 'ciplvt'
+                        R(:,:,iBand) = (imag(exp(1i * angle(HA(iA,:)./HB(iB,:)))))./sqrt(1-(real(exp(1i * angle(HA(iA,:)./HB(iB,:))))/nTime).^2); % Proposed by Daniele Marinazzo
+                        Comment = 'ciPLVt: ';
+                    case 'wplit'
+                        R(:,:,iBand) = abs(sin(angle(HA(iA,:))-angle(HB(iB,:))))./(sin(angle(HA(iA,:))-angle(HB(iB,:)))); % Proposed by Daniele Marinazzo
```
danielemarinazzo, line 568: R(:,:,iBand) = abs(sin(angle(HA(iA,:)')-angle(HB(iB,:)')))./(sin(angle(HA(iA,:)')-angle(HB(iB,:)')));

ftadel:
Thanks for the prompt fixes.
However, there are still issues to address:

### Error 1

**1xN ciPLV**: returns `Inf` values. See example for the file in the example database.

image

### Error 2

**NxN wPLI**: Crashes with the error below

```
** Error: [process_plv1n]  Connectivity > Phase locking value NxN
** Line 520: Unable to perform assignment because the size of the left side is 274-by-274 and the size of the right side is 1-by-75076.
** 
** Call stack:
** >bst_connectivity.m at 520
** >process_plv1n.m>Run at 110
** >process_plv1n.m at 24
** >bst_process.m>Run at 230
** >bst_process.m at 37
** >panel_process_select.m>ShowPanel at 3042
** >panel_process_select.m at 32
** >bst_report.m>Recall at 1245
** >bst_report.m at 62
** >bst_call.m at 28
** >gui_brainstorm.m>@(h,ev)bst_call(@bst_report,'Recall','current') at 299
** 
** 
** File: TestConnect/Connect/data_standard_average_211107_1441.mat
** 
```

image

### Deployment

Can you please give me the exact references these measures are coming from?
I need to add them both in the code in the tutorial.

Note that if we add these measures to the Brainstorm repository, hundreds of users are going to start using them and to publish with the results they obtain.  It is our responsibility to provide only solutions we are 100% confident it.
Please take the necessary time to triple-check that what is obtained with the code in this PR is really what you want to distribute publicly. 
Thanks!

danielemarinazzo (changes requested):
added transpose in wplit

ftadel:
### Test dataset

Download this file: 
https://www.dropbox.com/s/pfhrvrd1hft1m0i/TestConnect.zip?dl=0

Load it with `File > Load protocol > Load from zip file`.

ftadel:
> added transpose in wplit

Where? 
I'm not using your initial PR. You must commit to: https://github.com/ftadel/brainstorm3-1

danielemarinazzo:
The dimensionality issues in wPLI should be fixed, as well as the inf values in ciPLV. The implementation using phaseA and phaseB (the current one for PLV), gives Inf for the self connections in ciPLV, so for the latter I used another implementation, which I added commented for consistency in the PLV case

danielemarinazzo:
For the references:

ciPLV
Bruña R, Maestú F, Pereda E. Phase locking value revisited: teaching new tricks to an old dog. J Neural Eng. 2018 Oct;15(5):056011. doi: 10.1088/1741-2552/aacfe4. Epub 2018 Jun 28. PMID: 29952757.

wPLI
Vinck M, Oostenveld R, van Wingerden M, Battaglia F, Pennartz CM. An improved index of phase-synchronization for electrophysiological data in the presence of volume-conduction, noise and sample-size bias. Neuroimage. 2011 Apr 15;55(4):1548-65. doi: 10.1016/j.neuroimage.2011.01.055. Epub 2011 Jan 27. PMID: 21276857.

ftadel:
Thanks for the interactive debugging!
I've added the references to the code and tutorial.

At the current stage, there are still some errors, that you can reproduce with the protocol I shared previously:
- **1xN ciPLV**: Still contains Inf values
- **NxN ciPLV**: Computation error:
`** Line 524: Unable to perform assignment because the size of the left side is 274-by-274 and the size of the right side is 75076-by-1.`
- **NxN wPLI**: Computation error:
`** Line 532: Unable to perform assignment because the size of the left side is 274-by-274 and the size of the right side is 75076-by-1.`

danielemarinazzo:
Can you please check now?
thanks!

ftadel:
I still have some errors, from the TestConnect protocol:

### Error 1: 1xN ciPLV

Script:
```
% Input files
sFiles = {...
    'TestConnect/Connect/data_standard_average_211107_1441.mat'};

% Start a new report
bst_report('Start', sFiles);

% Process: ciPLV:  Corrected imaginary phase locking value
sFiles = bst_process('CallProcess', 'process_plv1', sFiles, [], ...
    'timewindow',   [-0.1, 0.5], ...
    'src_channel',  'MLC55', ...
    'dest_sensors', 'MEG, EEG', ...
    'includebad',   1, ...
    'freqbands',    {'delta', '2, 4', 'mean'; 'theta', '5, 7', 'mean'; 'alpha', '8, 12', 'mean'; 'beta', '15, 29', 'mean'; 'gamma1', '30, 59', 'mean'; 'gamma2', '60, 90', 'mean'}, ...
    'plvmethod',    'ciplv', ...  % ciPLV:  Corrected imaginary phase locking value
    'keeptime',     0, ...
    'plvmeasure',   2, ...  % Magnitude
    'outputmode',   1);  % Save individual results (one file per input file)

% Save and display report
ReportFile = bst_report('Save', sFiles);
bst_report('Open', ReportFile);
```

Error message:
```
** Error: [process_plv1]  Connectivity > Phase locking value 1xN
** Line 52:   ^ 
** Incorrect dimensions for raising a matrix to a power. Check that the matrix is square and the power is a scalar. To operate on each element of the matrix individually, use POWER (.^) for elementwise power.
** 
** Call stack:
** >mpower.m at 52
** >bst_connectivity.m at 523
** >process_plv1.m>Run at 110
** >process_plv1.m at 24
** >bst_process.m>Run at 230
** >bst_process.m at 37
** >panel_process_select.m>ShowPanel at 3042
** >panel_process_select.m at 32
** >bst_report.m>Recall at 1245
** >bst_report.m at 62
** >bst_call.m at 28
** >gui_brainstorm.m>@(h,ev)bst_call(@bst_report,'Recall','current') at 299
```

### Error 2: NxN ciPLV

Not really an error:
It looks like this measure is unstable, it has some values that are diverging a lot from the others:
image

While all the other graphs look more like:
image

Is this expected?

danielemarinazzo:
I fixed the first issue.

The ciPLV NxN looks ok when I look at it (it does not look like the one in your figure). Maybe try again (not sure if it was a previous commit)?

ftadel:
### ciPLV 1xN

There are still Inf values in the output files.
Example script:
```
% Input files
sFiles = {...
    'TestConnect/Connect/data_standard_average_211107_1441.mat'};

% Process: ciPLV:  Corrected imaginary phase locking value
sFiles = bst_process('CallProcess', 'process_plv1', sFiles, [], ...
    'timewindow',   [-0.1, 0.5], ...
    'src_channel',  'MLC55', ...
    'dest_sensors', 'MEG, EEG', ...
    'includebad',   1, ...
    'freqbands',    {'delta', '2, 4', 'mean'; 'theta', '5, 7', 'mean'; 'alpha', '8, 12', 'mean'; 'beta', '15, 29', 'mean'; 'gamma1', '30, 59', 'mean'; 'gamma2', '60, 90', 'mean'}, ...
    'plvmethod',    'ciplv', ...  % ciPLV:  Corrected imaginary phase locking value
    'keeptime',     0, ...
    'plvmeasure',   2, ...  % Magnitude
    'outputmode',   1);  % Save individual results (one file per input file)

% Count number of Inf values in the output file
TfMat = in_bst_timefreq(sFiles.FileName);
nnz(isinf(TfMat.TF(:)))
```

It returns 5: values that we can see in this map, with any freq band >= theta.
The Inf values correspond to the ciPLV computed from one signal with itself (MLC55).
If this is expected, this case must be explicitly handled, and replaced with a real value (0 if you can't find anything better).

image

### ciPLV NxN

> The ciPLV NxN looks ok when I look at it (it does not look like the one in your figure). Maybe try again (not sure if it was a previous commit)?

I can't reproduce this, indeed. 
I was maybe running the code from the wrong branch...

danielemarinazzo:
Which viz are you using? I don't see this with the 2D/3D sensors and the disk.. Here the file 
https://www.dropbox.com/s/48ua2t4k1f7gnzj/connect1_ciplv_220204_1423.mat?dl=0

anyway it's indeed a value that should be zero, or very close to it, by construction.

ftadel:
Your file doesn't have the Inf values, indeed.
I've been spending the past 2hrs trying to figure out why mine does have Inf values...
I guess this is due to some numerical approximations, which are different depending on the version of Matlab.

First of all, are you sure you have committed all the changes you made to the master branch of repository?
https://github.com/ftadel/brainstorm3-1

To investigate further, you can download this file and unzip it:
[data_standard_average_211107_1441.zip](https://github.com/brainstorm-tools/brainstorm3/files/8004196/data_standard_average_211107_1441.zip)

Then execute this code:

```
TfMat = load('data_standard_average_211107_1441.mat');

hilbert_fcn = @hilbert;
%hilbert_fcn = @oc_hilbert;

iMLC55 = 21;
DataAband = process_bandpass('Compute', TfMat.F(iMLC55,:), 600, 5, 7, 'bst-hfilter-2019', 0);
DataBband = process_bandpass('Compute', TfMat.F, 600, 5, 7, 'bst-hfilter-2019', 0);
HA = hilbert_fcn(DataAband')';
HB = hilbert_fcn(DataBband')';
phaseA = HA ./ abs(HA);
phaseB = HB ./ abs(HB);

R = (imag((phaseA*phaseB'))/size(HA,2))./sqrt(1-(real((phaseA*phaseB'))/size(HA,2)).^2);
R(iMLC55)
```

On my computer, this returns **-Inf**, with both Hilbert functions.
Can you reproduce this?

The problem is that the filtered version of MLC55 is not strictly identical in DataAband and DataBband:
```
max(abs(DataAband-DataBband(iMLC55,:)))
ans =
   3.9443e-31
```

This case should be handled properly in `bst_connectivity.m`.
If you think the formula is what it is supposed to be, then we need to add an extra line to replace the Inf values with 0 (similarly to what is done at line 635 for NaN.

danielemarinazzo:
weird indeed. with your file I also have inf, in the file I sent you I was
using the NxN so that DataAband and DataBband were the same.

The ciPLV values for self connections will anyway be practically zero.
Either we explicitly zero them, as for NaNs, or we add an eps to the
definition

                        R(:,:,iBand) =
(imag((phaseA*phaseB'))/size(HA,2))./sqrt(1+eps-(real((phaseA*phaseB'))/size(HA,2)).^2);

whatever you feel more comfortable with.

ftadel:
Thanks!

If that works for you, let's do both: 
- the +eps to avoid a strict 0 at the denominator
- a generic search for Inf values, in case any other measure returns an Inf value for any reason.

I think we're getting there! Thank you for helping me fix all these details.
Whenever you're OK with it, I'll merge the PR.

danielemarinazzo:
Thanks to you!
It could be useful to give it a try with another dataset, or with some vertex-wise measures (more high-dimensional data), and maybe simulated data. I can try to retrieve one from some collaborators who regularly use brainstorm.

Regarding simulated data, what's the easiest way to include a 3D dataset [channels trials time] in a protocol? or to embed in a DataAband-like structure? If I try to load some 3D data into the GUI it makes it 2D. For example the *wideband* data here 
https://www.dropbox.com/s/veuccxgw9dtqfrx/test_PLI.mat?dl=0

ftadel:
> Regarding simulated data, what's the easiest way to include a 3D dataset [channels trials time] in a protocol?
> For example the wideband data here

If you want to import a 3D matrix as a bunch of trials in a Brainstorm database, your data should be stored as **[channels x time x trials]** or **[time x channels x trials]**. THen it would be easier if you were saving in the .mat file only the field you want to import.

Then right-click on the empty folder > **Import MEG/EEG**, file format "**EEG: Matlab matrix**".

image

image

image

ftadel:
I'll wait until you run some more tests before merging then.

danielemarinazzo:
works very well, both the import and the test!

One question though, in the sample dataset you sent me, there weren't many
trials, it was just one trial right? and if yes, then the time-resolved
calculation was not an average across trials, but a single estimate?

ftadel:
> works very well, both the import and the test!

I'll merge it then!

> One question though, in the sample dataset you sent me, there weren't many trials, it was just one trial right? and if yes, then the time-resolved calculation was not an average across trials, but a single estimate?

It was the average of the trials from the introduction tutorials: 
https://neuroimage.usc.edu/brainstorm/Tutorials/Averaging
The results are meaningless, it was only a random way to illustrate some computation bugs.

danielemarinazzo:
Hi, I implemented (and tested) a faster implementation. Wanna push this too, or shall I PR to the main repo?

Moo-Marc:
Hi @danielemarinazzo & @ftadel,

Sorry, I'm a bit late to the party, but I'm interested in this code, in particular because I've looked closely at these various phase measures and implemented them once before.  I have a few comments specifically on wPLI.  I haven't looked at the ciPLV paper yet.

1. I don't think the code actually corresponds to wPLI as it was published.  It uses phase differences, phaseA*phaseB', whereas wPLI in Vinck et al. 2011 is defined with HA*HB'.  Where did this formula come from?
2. Furthermore, it's not currently the "debiased" version. So the comment on line 599 should be corrected.
3. There's also an important question of how trial-averaging should be performed (which we can discuss at a Brainstorm meeting).  Many of the connectivity measures have multiple expectation value terms, or other operations after the expectation value.  This means the computation should proceed in two steps: trial averaging each expectation value term separately, then the final assembly of the terms after looping through files.  Computing the measure for each trial and then averaging is not equivalent and will probably result in increased variance among other things.  This is similar to what was done for coherence.  This would also allow implementing the debiased wPLI, "unbiased square PLV", etc.

ftadel:
@Moo-Marc Could you please open this an issue, with link to all the PRs and issues related with this topic?
Otherwise, there is no easy way to keep track that there are unsolved issues in a closed/merged PR.
Thanks

PR #498, opened 2022-02-02
Bugfix: Offset of bad segments in evt_detect when using TimeWindow

rcassani:
I believe these are the only two places the fix is needed


ftadel:
Good catch!!

ftadel:
Shouldn't this fix be added also to `process_evt_detect_badsegment.m`, that also has bad segments and a selected time window?

PR #503, opened 2022-02-07
faster implementation ciPLV, PLVt, ciPLVt, wPLIt

danielemarinazzo:


ftadel:
Thanks!

I removed a few extra parentheses 
and added back previous implementation as comments for future reference and/or other readers.

PR #504, opened 2022-02-09
fix NIRS export

Edouard2laire:
This PR fixes an issue when, when exporting an MRI and a NIRS montage, the two would end up not aligned in Brainsight software.  This is linked to the change made here: https://github.com/brainstorm-tools/brainstorm3/commit/891c3640b7e8baed3e6287903c4c523dd5479956

A small summary of my current understanding:  MRI is stored in ?  coordinate.  MRI also contains two transformations; qform; and sform: qform transform the voxel coordinate to scanner space. sform transform voxel coordinate to standard space (MNI).   When both transformation exist, brainstorm is only saving qform. (so transformation to scanner space). 

When exporting nirs optodes coordinate, we first convert the SCS coordinate to MRI coordinate. then we convert the MRI coordinate to the same system of coordinate used when exporting the MRI.  Currently, Brainstorm is always exporting the NIRS optodes to standard space (using sform). We need to export to scanner space when Brainstorm is exporting the QFORM.  

The two questions I still haven't figure out: 
- Why do some subjects have qform, some don't?  (We had the problem of alignment only on a few subjects). At first, I would have thought the opposite: all subjects have qform (we know scanner coordinate), and some also have sform (to know MNI coordinate). But it seems to be the opposite? 
- Why Brainstorm is only saving the qform, when it exists, and not qform and sform ? 
Based on https://nifti.nimh.nih.gov/nifti-1/documentation/faq#Q19 it seems more logical

> Q19. Why does NIfTI-1 allow for two coordinate systems (the qform and sform)? (Mark Jenkinson)
> The basic idea behind having two coordinate systems is to allow the image to store information about (1) the scanner coordinate system used in the acquisition of the volume (in the qform) and (2) the relationship to a standard coordinate system - e.g. MNI coordinates (in the sform).
> 
> The qform allows orientation information to be kept for alignment purposes without losing volumetric information since the qform only stores a rigid-body transformation which preserves volume. On the other hand, the sform stores a general affine transformation which can map the image coordinates into a standard coordinate system, like Talairach or MNI, without the need to resample the image.
> 
> By having both coordinate systems, it is possible to keep the original data (without resampling), along with information on how it was acquired (qform) and how it relates to other images via a standard space (sform). This ability is advantageous for many analysis pipelines, and has previously required storing additional files along with the image files. By using NIfTI-1 this extra information can be kept in the image files themselves.
> Note: the qform and sform also store information on whether the coordinate system is left-handed or right-handed (see [Q15](https://nifti.nimh.nih.gov/nifti-1/documentation/faq#Q15)) and so when both are set they must be consistent, otherwise the handedness of the coordinate system (often used to distinguish left-right order) is unknown and the results of applying operations to such an image are unspecified.

Few usefull links: 
https://community.mrtrix.org/t/nifti-transform-ambiguity/2254

https://community.mrtrix.org/t/clarification-on-qform-and-sform/3291/2

https://nifti.nimh.nih.gov/nifti-1/documentation/nifti1fields/nifti1fields_pages/qsform_brief_usage

### DIFF
--- a/toolbox/io/out_channel_nirs_brainsight.m
+++ b/toolbox/io/out_channel_nirs_brainsight.m
@@ -64,40 +64,94 @@ if nargin >= 3
     % Same code as in out_mri_nii.m to make sure exported coordinates have the
     % same origin as exported MRI
     % Use existing matrices (from the header)
-    if isfield(sMri, 'Header') && isfield(sMri.Header, 'nifti') && ...
-            all(isfield(sMri.Header.nifti, {'qform_code', 'sform_code', 'quatern_b', 'quatern_c', 'quatern_d', 'qoffset_x', 'qoffset_y', 'qoffset_z', 'srow_x', 'srow_y', 'srow_z'})) && ...
-            (any([sMri.Header.nifti.srow_x sMri.Header.nifti.srow_y sMri.Header.nifti.srow_z]~=0) || ...
-             any(sMri.Header.nifti.qform(:)~=0))
-         if any([sMri.Header.nifti.srow_x sMri.Header.nifti.srow_y sMri.Header.nifti.srow_z]~=0)
-             sform = [sMri.Header.nifti.srow_x ;  sMri.Header.nifti.srow_y ; sMri.Header.nifti.srow_z];
-         elseif any(sMri.Header.nifti.qform(:) ~= 0)
-             if ~all(sMri.Header.nifti.qform(4, :) == [0 0 0 1])
-                 bst_error('Scaling in qform is not handled.');
-                 return;
-             end
-             sform = sMri.Header.nifti.qform(1:3, :);
-         end 
-    else % Otherwise: Try to define from existing information in the database
-        if isfield(sMri, 'NCS') && isfield(sMri.NCS, 'Origin') && ~isempty(sMri.NCS.Origin)
-            Origin = sMri.NCS.Origin - [1 2 2];
-        elseif isfield(sMri, 'NCS') && ((isfield(sMri.NCS, 'R') && ~isempty(sMri.NCS.R)) || (isfield(sMri.NCS, 'y') && ~isempty(sMri.NCS.y)))
-            Origin = cs_convert(sMri, 'mni', 'mri', [0 0 0]) .* 1000;
-        elseif isfield(sMri, 'NCS') && isfield(sMri.NCS, 'AC') && ~isempty(sMri.NCS.AC)
-            Origin = sMri.NCS.AC + [0, -3, 4];
-        else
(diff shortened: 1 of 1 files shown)


ftadel:
Understanding the qform/sform issues in NIfTI is complicated, indeed.
One very annoying part is that files that contain both QFORM and SFORM are handled differently by different programs.
For this reason, when exporting .nii files Brainstorm saves only one of the two, with a preference for the QFORM.
(See out_mri_nii.m)

The QFORM matrix corresponds to the World coordinates, not the MRI coordinates.
https://neuroimage.usc.edu/brainstorm/CoordinateSystems#World_coordinates

If you import a .nii file, then export it again, the NIfTI headers should remain untouched.
If you compute some normalization, resampling or registration, then the NIfTI headers are deleted, and then when exporting the MRI as a .nii file, Brainstorm may generate a different combination of sform/qform.
If you import a FreeSurfer .mgz and export it to .nii (which don't have any nifti headers), you might get a different output .nii than if you had imported the .nii equivalent.

I don't like this idea of including again all this conversion code you have submitted here. Instead:
- Use `cs_convert.m` to convert the positions from the channel file from `'scs'` to `'world'` coordinates.
- Fix your processing pipelines to have the MRI saved in a reproducible way (e.g. with the vox2ras / InitTransf matrix set, and with the original nifti headers deleted)

If you need more help with understanding any of this: please share examples.

Edouard2laire:
Thx. What would be the best way for me to send an example to you? I can create a minimal brainstorm protocol and send it to you. I am not sure however how you could see if the exportation worked. 

(for background brainsight is the software we use for data acquisition so we export both MRI and the optodes locations. We then import them in Brainsight to know where to put them on the subjects; when its not aligned the optodes appears to be in subject head)

Edouard2laire:
ok. So this seems to be working. But I am wondering if there are cases where we would like to export to MNI coordinate instead of world?

ftadel:
This last change looks a lot neater than the previous suggestion, indeed! :)

> But I am wondering if there are cases where we would like to export to MNI coordinate instead of world?

Do you really mean MNI here? or was it MRI?
- Export 3D points as MNI coordinates: this is useful for exporting brain locations (SEEG contact or scouts) to another program, for reporting standard coordinates in a publication, or for group analysis. This should never be used for anything outside the head.
- Export 3D points as MRI coordinates: These MRI coordinates are only used by the CTF software and internally in Brainstorm. I don't see where that could be relevant outside of Brainstorm.

> I will test on few more exemple and let you know> You might be able to merge this at the end of the week :)

Should I merge it now?

Edouard2laire:
Yes, you can push it.

PR #507, opened 2022-02-14
small fixes for importing CTF BIDS dataset

Moo-Marc:
Fix importing events.tsv from BIDS dataset: trial_type only labels certain events (that define trial categories), so also look at 'value' column.

Removed other unnecessary warnings.

### DIFF
--- a/toolbox/io/in_channel_bids.m
+++ b/toolbox/io/in_channel_bids.m
@@ -28,9 +28,11 @@ function ChannelMat = in_channel_bids(ChannelFile, factor)
 % Authors: Francois Tadel, 2018
 
 % Read the TSV file
-tsvValues = in_tsv(ChannelFile, {'name', 'x', 'y', 'z', 'group', 'type'});
+% For _electrodes.tsv, the first 4 columns below are required.
+% For _channels.tsv, 'name', 'type' and 'units' are required.
+tsvValues = in_tsv(ChannelFile, {'name', 'x', 'y', 'z', 'group', 'type'}, 0); % don't warn since columns depend on file type.
 if isempty(tsvValues) || isempty(tsvValues{1,1})
-    disp('BIDS> Error: Invalid _electrodes.tsv file.');
+    disp('BIDS> Error: Invalid tsv file.');
     ChannelMat = [];
     return;
 end
--- a/toolbox/io/in_channel_bids.m
+++ b/toolbox/io/in_channel_bids.m
@@ -28,11 +28,9 @@ function ChannelMat = in_channel_bids(ChannelFile, factor)
 % Authors: Francois Tadel, 2018
 
 % Read the TSV file
-% For _electrodes.tsv, the first 4 columns below are required.
-% For _channels.tsv, 'name', 'type' and 'units' are required.
-tsvValues = in_tsv(ChannelFile, {'name', 'x', 'y', 'z', 'group', 'type'}, 0); % don't warn since columns depend on file type.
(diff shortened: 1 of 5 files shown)


toolbox/io/in_events_bids.m
```diff
@@ -36,19 +40,36 @@
...
-    if length(uniqueEvtVal) > 1
+    if length(uniqueEvt) == 1 && length(uniqueEvtVal) > 1
         iColumn = 5;
-        uniqueEvt = uniqueEvtVal;
+    elseif length(uniqueEvt) > 1 && length(uniqueEvtVal) == 1
+        iColumn = 3;
+    else
+        % Merge both columns
+        iColumn = 3;
+        Markers(:,iColumn) = regexprep(Markers(:,iColumn), {'n/a', 'N/A'}, '');
```
ftadel, line 59: This should be done before testing which columns are empty.
Before the test, just after reading Markers, something like:
```
Markers(cellfun(@(c)or(isequal(c,'n/a'),isequal(c,'N/A'), Markers(:)) = {[]};
```

toolbox/io/in_events_bids.m
```diff
@@ -36,19 +40,36 @@
...
+        for iM = 1:size(Markers, 1)
+            Markers{iM,iColumn} = [Markers{iM,iColumn} '-' Markers{iM,5}];
```
ftadel, line 61: This is complicated to read.
I'd suggest your remove the use of iColumn, keep only the column 3.
For all the cases (outside of any test): loop on all the Markers rows, and do something like:
- if col3 only: nothing to change
- if col5 only: copy col5 to col3
- if col3 AND col5 non-empty: col3=col3 '-' col5

The loop would be executed for all the files, including the files with only one column defined, but the increase of computation time would be negligible.

ftadel (commented):
There are few things I'd suggest to reformat, otherwise it looks good!

toolbox/io/in_events_bids.m
Moo-Marc: Simplified.

toolbox/io/in_channel_bids.m
```diff
@@ -28,9 +28,11 @@
 % Authors: Francois Tadel, 2018
 
 % Read the TSV file
-tsvValues = in_tsv(ChannelFile, {'name', 'x', 'y', 'z', 'group', 'type'});
+% For _electrodes.tsv, the first 4 columns below are required.
+% For _channels.tsv, 'name', 'type' and 'units' are required.
```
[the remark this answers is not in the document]
Moo-Marc: Well, it's what happens when importing a BIDS dataset: it uses this function.  I'll check the trace.

Moo-Marc:
Ready: corrected as suggested and tested.

toolbox/io/in_events_bids.m
ftadel: If there are empty cells, regexprep crashes.
Please replace regexprep with the cellfun/isempty suggestion above.

toolbox/io/in_events_bids.m
ftadel: Looks good!

toolbox/io/in_events_bids.m
Moo-Marc: It did not crash for me...  But the reason I edited this file was that missing columns had empty double {[]}, which crashed when using unique.  So I'm also replacing those with {''} now.

ftadel:
Thanks!

PR #510, opened 2022-02-18
Tutorial corticomuscular

rcassani:
Script for corticomuscular coherence tutorial
https://neuroimage.usc.edu/brainstorm/Tutorials/CorticomuscularCoherence


ftadel:
I made directly a few changes. 
Here are a few more change requests:
- Remove all the unnecessary plots that you don't include in the reports: you used `tutorial_introduction.m` as a starting point, but this includes way too many details, to reproduce the figures of the introduction tutorials.  Here: try to be as synthetic as possible, straight to the point, and generate only the figures that you will add in the report. Better starting points are tutorial_epilepsy.m or tutorial_fem.m for instance.
- The snapshot process showing the MEG+MRI registration would be sufficient for checking both the anatomical segmentation and the import of the sensors.
- CAT12: do not change the default cortex
- Do not change the exploration mode: this is not something that should be done in a processing pipeline.
- No need to change anything to events manually
- No need to apply the CTF  compensation
- Add the splitting to the import directly

For the scouts before/after, I'm still confused... 
It looks that we still don't have any valid option available for computing full connectomes...

rcassani:
@ftadel, this PR is ready

ftadel:
I've been editing the code heavily, along the lines of my previous comment.
I'm stopping here for today, will resume tomorrow...

For the tutorials, I still need to proof-read everything.
With a quick overview: I still have the impression that computing at the same time the surface and volumes  source models is difficult to read, and still think that the details about the computation of source-level connectivity belongs to the connectivity tutorial. But I'll come back later this week with more constructive comments :)

toolbox/script/tutorial_coherence.m
```diff
@@ -142,28 +106,11 @@ function tutorial_coherence(tutorial_dir, reports_dir)
     'orient',   1, ...  % left
     'Comment',  'MEG/MRI Registration');
 
-% View MEG/MRI registration
```
rcassani: As far as I remember, the MEG/MRI registration was going to be shown

toolbox/script/tutorial_coherence.m
```diff
@@ -298,15 +238,7 @@ function tutorial_coherence(tutorial_dir, reports_dir)
     'freq',          [], ...
     'baseline',      'all', ...
     'blsensortypes', 'MEG');
-% View recordings, trial 1
```
rcassani: This plot is for comparison with the FieldTrip tutorial

toolbox/script/tutorial_coherence.m
ftadel: And it is, the line just above :)

ftadel:
Done for today: I reviewed and edited everything until "Coherence 1xN (source level)" in the tutorial and the script.
Will review the source-level connectivity tomorrow.

ftadel:
Some disorganized comments:
- I apologize for removing a lot of the text you wrote from the corticomuscular coherence tutorial: Most of it was well explained and well written, but unrelated with the purpose of the tutorial. The more synthetic, the easier to read through. 
- For cross-readability between the FieldTrip and Brainstorm tutorials, I suggest we try using "trials" to refer to the 10s-trials of the experiment, and "epochs" to designate the 1s-blocks imported in Brainstorm. And not using the two interchangeably like in the introduction tutorials.
- SSP: The second component for the "blink" SSP and the first of the "blink2" and "blink3" SSP are all related with eye movements. Let's not dive into more details here, as this is not the purpose of this tutorial. I removed completely these sections to avoid having to discuss it here.
- I moved the section Scout NxN to the Connectivity tutorial. I'll work later on re-arranging better a source-level connectivity section in the Connectivity tutorial. 
- Please make some efforts in reducing the size of the images you are using in the online tutorials. Individually, for each image, a few hundreds of Kb doesn't seem like a lot, but the total size of the images of the corticomuscular coherence is above 12Mb at the moment. This makes this page painful to load outside of the super-fast inter-university connections within North America, and maybe impossible to load completely in some less fortunate parts of the world. A bad side effect (even for me on a 4G connection in Europe) is that when loading a link to a specific anchor in the middle of the page, the browser window will keep flashing for a while and end-up pointing at an incorrect correct location of the anchor. Some guidelines for lighter pages:
    - For any image containing only interface elements (a handful of flat colors and no continuous color gradients): always use GIF instead of PNG, this simple change can divide the size of the image by 2 to 4. 
    - Use small images, both for readability of the page and for size matters: either a screen capture of a small image, or a large image resized before being attached to the page.
    - When needing a large image, like the diagrams for the source connectivity: create a smaller GIF to link to the large PNG, like I did for the moved Scout NxN diagrams: https://neuroimage.usc.edu/brainstorm/Tutorials/Connectivity#Source-level_connectivity_.5BTODO.5D
    - GIMP and Photoshop do a bad job at resizing large images with text and diagrams, but ImageMagick gives very similar results to Firefox, which are are much more readable: https://superuser.com/questions/1093805/how-to-achieve-high-quality-downscaling-of-diagrams-like-in-firefox

ftadel:
I could not finish this morning... 
On Monday, I will resume at "Coherence: EMG x Scouts" in the tutorial, and "Coherence: EMG x Sources" in the script.

ftadel:
Report generated with the current script:
[report_TutorialCMC_220315_2156.zip](https://github.com/brainstorm-tools/brainstorm3/files/8261953/report_TutorialCMC_220315_2156.zip)

I still need to work on the following items:
 - Bug: When a new file is created, it should be added to all the search tabs + the original Database tab (having to refresh the Database tab to see the new files doesn't seem right)
 - Bug: 1xN connectivity displayed as image: missing X-axis labels
 - Optimization: Computation of the scout-level connectivity, not preloading all the data

ftadel:
Sure! This deletion was an unintended mistake. 
Thanks for catching it.

https://github.com/brainstorm-tools/brainstorm3/commit/374f466203a06fa935d313bee3ebc291c29f34e9

PR #511, opened 2022-02-20
New option: copy digitized fids to MRI when aligning with head points

Moo-Marc:
[Initial discussion on the forum.](https://neuroimage.usc.edu/forums/t/subject-coordinate-system-after-refine-with-head-points/8431/9)

The main idea is to avoid having to adjust MRI fiducials manually (nasion and ears) when these points were digitized.  As Francois points out, this is not required, but I feel many users would not realise this, or would still want the points that define the subject coordinate system (SCS) to be where they expect and "look good".  This was also simple enough to implement, so it happens just once per subject.

I added this option to process_headpoints_refine, process_adjust_coordinates, and process_import_bids.  There are probably other places where this head point adjustment can be launched and this option could be added there too, if we decide to accept this change.


ftadel:
I disagree with the logic of modifying the fiducials from the MRI, as explained here:
https://neuroimage.usc.edu/forums/t/subject-coordinate-system-after-refine-with-head-points/8431/13

> I feel many users would not realise this, or would still want the points that define the subject coordinate system (SCS) to be where they expect and "look good". 

What you point at in this justification is a problem of education/documentation, not a problem of data workflow.
If you feel like there is anything that is not clearly explained in the tutorial, please feel free to suggest adding more information to it: https://neuroimage.usc.edu/brainstorm/Tutorials/ChannelFile#Automatic_registration

Please do not mention the default MNI fiducials here, as this is not considered to be the gold standard in terms of MEG/MRI registration. This is a recommended procedure only for EEG and iEEG, hence its use is limited to this context:
https://neuroimage.usc.edu/brainstorm/Tutorials/Epilepsy#Import_the_anatomy
https://neuroimage.usc.edu/brainstorm/Tutorials/ECoG#Import_the_anatomy
https://neuroimage.usc.edu/brainstorm/Tutorials/Epileptogenicity#Import_the_anatomy

Moo-Marc:
@ftadel, @rcassani, @sbaillet: 
Following our meeting discussion, here is the link to the BIDS conversation about saving coregistration:
https://groups.google.com/g/bids-discussion/c/BeyUeuNGl7I/m/gOrUGaRcCQAJ
There you can read that in MNE, after coregistration with head points, they keep the digitized fiducials, not the ones from the MRI.  The latter are no longer used in MNE, while we do the opposite.  I couldn't today find where I saw this in their code, but Mainak did confirm in that thread that my understanding was correct.

They also have an automated process for coregistration that's basically what I suggested, i.e. using fids from a template (equivalent to our default MNI fids) as the initialization for head point fitting, and keeping the digitized fids after the fit: https://mne.tools/stable/auto_tutorials/forward/25_automated_coreg.html
The difference is that they always keep the digitized fids, whereas we'd have to copy them to the MRI since that's what we keep.  So the argument that what I propose is new or not used in the community doesn't hold.

Stepping back a little, using head points for coregistration implies that we believe it yields a better alignment than only using 3 points manually picked on the MRI.  Yes it has to be verified, but if we believed that there was more error or risk with this fitting procedure, then we wouldn't do it.  

Finally, while in Brainstorm, digitized points are attached to functional data, they are conceptually anatomical data.  This process is independent from the head coils, it only uses anatomical info.

ftadel:
> Following our meeting discussion, here is the link to the BIDS conversation about saving coregistration:
https://groups.google.com/g/bids-discussion/c/BeyUeuNGl7I/m/gOrUGaRcCQAJ

This discussion is mostly about the way to use the fields in BIDS, most of it is not relevant in our discussion.

> There you can read that in MNE, after coregistration with head points, they keep the digitized fiducials, not the ones from the MRI. 

MNE-Python and Brainstorm have very different ways to organize their data. In MNE-Python, there is no notion of database, no tracking of links between files, the users are 100% responsible for understanding what they do in terms of coregistration, in which coordinate system they are working, etc. 
This is not our approach: the data storage in Brainstorm is a lot more strict, you are not allowed to have two files in different coordinate systems in the same folder. This is less flexible, but safer for non-expert users.
In my opinion, comparing workflows and internal conventions between the two environment doesn't necessarily make a lot of sense.

> Stepping back a little, using head points for coregistration implies that we believe it yields a better alignment than only using 3 points manually picked on the MRI. Yes it has to be verified, but if we believed that there was more error or risk with this fitting procedure, then we wouldn't do it.

Indeed, the recommend approach is:
- Start a registration with manually defined fiducial points, or default values from a template depending on the context)
- Refine with head points if we have enough
- If the results is worse, go back to the basic 3-point registration.
Reference: https://neuroimage.usc.edu/brainstorm/Tutorials/ChannelFile#Automatic_registration
Your proposed changes do not improve any of this.

> The difference is that they always keep the digitized fids, whereas we'd have to copy them to the MRI since that's what we keep. 

Wrong: we keep both.
The anatomical fiducials defined from the anatomy reference volumein the MRI.SCS.
The digitized fiducials in ChannelMat.HeadPoints, which are used to register the MEG on the MRI. The fiducials are moved together with the sensors during the automatic refinement.

> Finally, while in Brainstorm, digitized points are attached to functional data, they are conceptually anatomical data. This process is independent from the head coils, it only uses anatomical info.

The two sets of three points are indeed both spatial information.
One set comes from the MRI scanner and is saved in the MRI file. This is independent from the head coils.
One comes from the MEG acquisition system, and is saved together with the MEG data. This is DEPENDENT of the head coils.

--

I would like us to stop spending time discussing these aspects, which are mostly conventions, to focus on the reasons for which I don't want to integrate the proposed changes in the Brainstorm:
- The changes you are proposing to implement are only cosmetic, they do not add any functionnality. The only goals are:
   - when a user clicks on the popup menu "MRI registration > Check" (which should be done only once after linking the MEG recordings to the database), they would see only one set of fiducial points instead of the two.
   - when a user clicks on the popup menu "Edit MRI" (which they should not use in the standard workflow illustrated in the introduction tutorials), they would see the digitized fiducials imported from the MEG back instead of the anatomical points initially set. My guess is that there would not be more than a handful of users to notice the difference induced by these changes.
- Your propose many modifications to sensitive files that took about a decade to finally get stable. I won't accept any change to `db_set_channel.m` or `channel_align_auto.m` unless for fixing bugs or adding new major features. The risk of suffering from side effects from these changes is too high, they are called in many different contexts and testing all of them would take a lot of time.
- As a result of your procedure: all the surface files would need to be rewritten. There would therefore be multiple versions of the anatomy files, and the result would depending in the order in which you import the MEG sessions in the database. This would be of source of reproducibility issues and would be a nightmare to track. We should work on simplifying the workflows and limit as much as possible the number of file versions to make the provenance tracking easier (as this will be one of our main goals in the next round of the Brainstorm grant). This is a suggestion for making the anatomy workflow much more complex, with feedback loops depending on the import of additional functional data.
- We have already taken too much time discussing these changes, which represent too many technical issues and added complexity with respect to the expected benefit. We should focus on developing tools that are truly useful for our users, and there is already an infinity of them: https://neuroimage.usc.edu/brainstorm/Next

Overall, your problem is only a display issue, not a fundamental data processing or data reprensentation issue.
As a replacement, I propose you write a much simpler function that opens the MRI Viewer and plots dots representing the digitized fiducials from the channel file, as non-editable information. If you want to see the MRI + the digitized fiducials, you don't need to change the fiducials saved in the MRI.
The same can be done for the 3D view: do not plot the original anatomical landmarks (or delete them) if you don't want to see them.

Moo-Marc:
Francois's mind is obviously set that this is not useful, but I spent my personal time doing this because I needed it.  And I'm sure others would like it.

I will certainly NOT use approximate points from the template as anatomical points used to define the coregistration in a large public BIDS dataset.  Users would be justified to think there was something wrong with the data.  It's not really a "display issue" at this point (outside Brainstorm), but rather following an expected convention as to what we call anatomical fiducials.  

So my pipeline is to import a BIDS dataset containing both MRI and MEG, do coregistration automatically with head points and then use a meaningful set of anatomical points from a single digitized file.  To easily export these coordinates for multiple sessions, it makes perfect sense to apply these points to the MRI in Brainstorm, then for each functional file I do the conversion from MRI to "Native" coordinates.  I don't see how else to do this logically within Brainstorm.  

So it's really a small change.  Most of your concerns apply to the current state of affairs.
- The only step that's new in what I propose is adjusting the MRI fids from the digitized ones instead of manually.  Any potential problem from doing things in the wrong order is already there.
- When importing anatomy in various ways, some surfaces are created in Brainstorm before manual placement of MRI fiducials. So you have to rewrite surface files after placing the fids. This is already the case. 
- There is no "feedback loop".  Coregistration occurs the same as before and we simply select points differently to define the SCS.
- Picking points manually is **less** reproducible than copying them from a file.  To improve reproducibility, we can save the points we choose, for example in BIDS metadata files, which is precisely what I am working on.  
- The head coils are not needed for what I propose, it's purely anatomical. Even if coils are present in the digitized file, there's no dependence on them. 

Finally, it's not the first time that you argue that changes I propose are not useful for others, and that it's too risky to edit the code.  This last argument is rather frustrating.

ftadel:
> Francois's mind is obviously set that this is not useful

More precisely: not relevant enough to spend that much time on it.
Our development resources are limited, our developments needs are infinite, we have to prioritize.

> So my pipeline is to import a BIDS dataset containing both MRI and MEG, do coregistration automatically with head points and then use a meaningful set of anatomical points from a single digitized file. 

Please prepare and share:
- a minimal example dataset (no files not needed for illustrating the discussion)
- a Brainstorm pipeline **without** your changes
- a description of why the results you obtain are unacceptable, including screen captures and a list of all the data fields containing inadequate information 
- your replacement proposition: screen captures and data fields, and a description of why this is better.

Please submit this before next Monday, so we have time to review it before we discuss it on next Monday.

> When importing anatomy in various ways, some surfaces are created in Brainstorm before manual placement of MRI fiducials. So you have to rewrite surface files after placing the fids. This is already the case.

I'm not sure what you refer to. Please share the reference to a use case presented in the tutorials or the Brainstorm forum.
This solution to rewrite the fiducials was banned a few years ago because it was too complicated to handle.

ftadel:
> I will certainly NOT use approximate points from the template as anatomical points used to define the coregistration in a large public BIDS dataset. Users would be justified to think there was something wrong with the data. It's not really a "display issue" at this point (outside Brainstorm), but rather following an expected convention as to what we call anatomical fiducials.

Case 1: If your concern is "outside Brainstorm", for the "public BIDS dataset": you can add an option in the BIDS export process, to use the spatial landmarks used during the data import process (the ones you don't like), or the digitized points converted back into the system of coordinates used in the BIDS metadata.

Case 2: If your concern is "within Brainstorm": you can write a separate process that does all of this: update the fiducials in the MRI based on the digitized fiducials defined in data file selected, rewrite all the surfaces, rewrite all the channel files. 

In both cases, you have alternate options to get where you want without making changes to the core functions of Brainstorm, which will keep on serving only the main convention and workflow (i.e. not allowing the changes to the anatomy files after they are imported). 

There are already exception to this logic, for example the warping based on digitized head points. Warping the anatomy rewrites all the anatomical files after the import of "functional data" (anatomical information derived from the MEG or EEG acquisition). This is done in an independent process, which execution can be scripted and tracked easily, and will never be included as part of the standard registration routines. It is documented separately from the basic tutorials, with a dedicated tutorial, for advanced users and specific use cases.

What prevents you from going into this direction of a separate process?
You'd be free to do whatever modification you need to all the files of your database, with no risk of unwanted side effects.
To document precisely what your process does (and why), you could write a dedicated tutorial section, either in the [BIDS export](https://neuroimage.usc.edu/brainstorm/ExportBids) tutorial (case 1) or in the [Coordinate systems](https://neuroimage.usc.edu/brainstorm/CoordinateSystems) tutorial. The documentation link in the process would point at this function.

Moo-Marc:
Given the other difficulties I ran into fitting the head points, I'll close this and make a new PR later.

PR #512, opened 2022-02-23
Bugfix: Specific channels for baseline when combining import+removeDC

rcassani:
* Allow to removing DC offset for specific channel types (or names) when `process_baseline` is combined with `process_import_data_event`, `process_import_data_time` or `process_import_data_epoch`

* Additional field handled `OptimizePipeline` and `OptimizePipelineRevert` in `bst_process`


ftadel:
Hmmm... that doesn't work either, because then if you call any import process by itself, then `sProcess.options.baseline.Value =[]` (because it is the default value), and therefore it applies RemoveBaseline = 'all'...

When combining the two processes in bst_process, we could maybe set `sProcess.options.baseline.Value = 'all';' instead of simply passing along the `[]` from the RemoveDC process, and then process this special value in each import process.
It's not very clean, but I don't think it matters much if the various processes make a slightly different usage of the `baseline` option...

And it would not cause any backward compatibility issue, because at the moment, it is a BUG: combining import process + removeDC does not remove the DC at all, so the existing scripts would keep on not doing what they were designed to... :)

Feel free to implement anything you think is cleaner than this.

PR #513, opened 2022-02-23
Bugfix: Baseline option was ignored if [] when present at process_imp…

rcassani:
`baseline` option was ignored if `[]` when passed with `process_import_data_{epoch, event, time}` in script generated with Brainstorm. The behaviour should be to apply DC removal with the `All file` option checked.

### DIFF
--- a/toolbox/process/bst_process.m
+++ b/toolbox/process/bst_process.m
@@ -2456,6 +2456,9 @@ function sProcesses = OptimizePipeline(sProcesses)
         switch (func2str(sProcesses(iProcess).Function))
             case 'process_baseline'
                 sProcesses(iImport).options.baseline = sProcesses(iProcess).options.baseline;
+                if isempty(sProcesses(iImport).options.baseline.Value{1})
+                    sProcesses(iImport).options.baseline.Value{1} = 'all';
+                end
                 sProcesses(iImport).options.blsensortypes = sProcesses(iProcess).options.sensortypes;
             case 'process_resample'
                 sProcesses(iImport).options.freq = sProcesses(iProcess).options.freq;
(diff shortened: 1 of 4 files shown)


rcassani:
Implemented with the parameter `baseline = 'all'`

> From https://github.com/brainstorm-tools/brainstorm3/pull/512#issuecomment-1049144836:
And it would not cause any backward compatibility issue, because at the moment, it is a BUG: combining import process + removeDC does not remove the DC at all, so the existing scripts would keep on not doing what they were designed to... :)

I found that the epilepsy and fem tutorials are in that group (ignored baseline). All other tutorials have an explicit `[ini, fin]` baseline.

If we want to use the `baseline = []` to indicate "All file" as in other processes; we could change the default value inside the process_import_data_{epoch, event, ttime} to `{[0 0]}` and check against it to verify that the `baseline = []` was really set by the user or the process combination. As you saw that will have the potential back compatibility.

ftadel:
> I found that the epilepsy and fem tutorials are in that group (ignored baseline). All other tutorials have an explicit [ini, fin] baseline.

In these two tutorials, we don't want to apply any baseline correction, because we've already been applying a high-pass filter :)
https://neuroimage.usc.edu/brainstorm/Tutorials/Epilepsy#Import_the_recordings
https://neuroimage.usc.edu/brainstorm/Tutorials/FemMedianNerve#Epoching

We really don't want `baseline=[]` to be interpreted as "All file" in the the import processes.

ftadel:
I fixed two more things:
- strcmp crashes with [] in input, or anything non-char
- we want to keep the initial behavior `.Value=[]` to set `RemoveBaseline='no'` (as per being the default in `db_template('ImportOptions')`)

ftadel:
I let you do one last round of verification of my changes, then merge it if you're OK with it.
Minor imprecision in there could lead to bad bugs in lots of existing scripts... :/

PR #514, opened 2022-02-24
Feature: Allow to keep original events when merged (in script)

rcassani:
Add an option to keep original events to `process_evt_merge`.
Default is Delete originals for back compatibility.

Case of use:
The user has multiple triggers to group in different ways for their analysis.
The alternative would be to duplicate event groups before merging.


tmedani:
Here it is :)

PR #515, opened 2022-02-24
3DViz bugfix and MRI visualization for FOOOF and SPRiNT

lucwilson:
Fixes a bug where swapping FOOOF/SPRiNT data being displayed in 3DViz would not update. Also enables FOOOF visualization tools for MRI view modes (missed that one, originally).

Note: Changes to mri_draw_cuts are made in the event that users try to load data from 0Hz (from FOOOF), which is a matrix of NaNs. If there are any other situations in which a matrix of NaNs could be involved in the code (which should instead cause an error), then an alternative fix will be needed.


rcassani:
Thanks @lucwilson!
🚫🐞

toolbox/anatomy/mri_draw_cuts.m
```diff
@@ -241,6 +241,10 @@
         % Reduce array amplitude to the the colormap size
         A = floor( (A - intensityBounds(1)) ./ (intensityBounds(2)-intensityBounds(1)) .* (size(CMap,1)-1) ) + 1;
     end
+    % If slice is full of NaNs (FOOOF @ 0Hz)
+    if all(isnan(A), 'all')
```
ftadel, line 245: `all(..., 'all')` is a recent syntax, and crashes with 2015b for example.

I replaced this test with:
```
    if (nnz(isnan(A)) > 0)
        A(isnan(A)) = 1;
    end
```

ftadel:
@lucwilson @rcassani 

Instead of the last minute test that is implemented here, could you try instead to create a volume that is not full of NaN?

We are not supposed to have NaN in these variables, so:
1) it might create other annoying side effects
2) it adds an extra useless test in a function that is critical for the interactivity of the interface, and for a very minor use case.... we really want to optimize mri_draw_cuts, not add stuff to it.

lucwilson:
@ftadel @rcassani 

I've opened a new PR that addresses the concerns above; NaNs are instead dealt with while loading cross-sections (in freq space) of data in bst_memory (only occurs when using topo/3dvis/mriviewer). This only affects data coming in from FOOOF, so all previous behaviour observed for mri_draw_cuts should be preserved.

PR #517, opened 2022-02-25
Alternative for FOOOF visualization in MRIViewer

lucwilson:
Migrated data substitution to bst_memory.


ftadel:
It looks simpler now, thanks! :)

PR #518, opened 2022-03-03
Add start mia

annesodub:
MIA plugin : added an item "Start MIA" to the plugin menu (sEEG). There might be a problem if MIA has not been already installed. Is there any solution to fix this? (e.g. a way to grey out the menu item if the plugin is not installed). Sorry for the mess in the commits.  Thanks!


toolbox/process/bst_avg_files.m
```diff
@@ -518,7 +518,7 @@
     OutChannelFlag  = ones(NbChannels, 1);
     OutChannelFlag(MeanBadChannels) = -1;
 else
-    OutChannelFlag = [];
```
ftadel, line 521: Again?! :)

annesodub: sorry about that, I just cloned again the repo to fix this.

annesodub:
> I pushed it all here: [ef1624c](https://github.com/brainstorm-tools/brainstorm3/commit/ef1624c509068b88ab1fe328858d86bbb645e44b)

Thank you Francois!

> 
> The MIA webpage is not accessible anymore: http://www.neurotrack.fr/mia/ This is the URL reference for the plugin: either you change the link in the plugin, or you restore an introduction page for MIA.

Thank you Francois for pointing that out. I accidentally unlinked the web page yesterday while trying to add a demo video. It is now back online.

PR #521, opened 2022-03-09
Pipeline editor: Add scroll panel for windows taller than screen

ftadel:


ftadel:
@rcassani 
The resize bug of the pipeline editor occurred when the window was larger than the screen.
I tried to put everything in a scroll panel, but it may have other unwanted side effects...
Can you check it fixes the issue on your end, without adding other weird behaviors?
Thanks

rcassani:
@ftadel works fine :)

Added missing scaling for two labels https://github.com/brainstorm-tools/brainstorm3/pull/521/commits/c2ecf4e8ccc82f172749475b42734f447586ed6f

The lists of clusters and scouts still miss scaling

PR #522, opened 2022-03-09
Allow rename cluster without GUI interaction

rcassani:
Allow rename cluster without GUI interaction, e.g.:

```
[sCluster, iCluster] = panel_cluster('CreateNewCluster', {'F2','F4','FC2','FC4'} );
panel_cluster('EditClusterLabel', 'right_anterior_medial', iCluster);
```

### DIFF
--- a/toolbox/gui/panel_cluster.m
+++ b/toolbox/gui/panel_cluster.m
@@ -613,44 +613,45 @@ function RemoveAllClusters()
     end
 end
 
-
 %% ===== EDIT CLUSTER LABEL =====
-% Rename one and only one selected cluster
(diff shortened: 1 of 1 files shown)


ftadel:
There was a typo with twice testing `varargin{1}`, and if the label already existed it was displaying an interactive message, which is not what you want for a script. I'm not sure what your plans are for this function, but I thought that it was safer to have the script calls never calling `java_dialog`.

Does it for you this way?

rcassani:
@ftadel, yes it makes more sense to completely avoid `java_dialog` windows when called from scripts

PR #523, opened 2022-03-10
Envelope correlation: Added 1xN and AxB processes

### LINKED ISSUE #499: Envelope correlation: Missing 1xN and AxB processes
The HENV functions developed by @HosseinShahabi still miss their generalization to any pair of signals.

At the moment, the function `bst_henv.m` is only capable of computing NxN graphs, as it only takes one matrix `Data` in input:
https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/connectivity/bst_henv.m#L1
 
It was requested from the beginning that this must work like all the other connectivity functions, with two lists of signals `X` and `Y`. 
https://neuroimage.usc.edu/forums/t/problem-with-time-dynamic-envelope-correlation/21542

Examples: 
- bst_cohn.m: https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/connectivity/bst_cohn.m
- bst_granger.m: https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/connectivity/bst_granger.m#L1

Implementing this requires some significant work, as many coding tricks won't apply.

Without this generalization, it is impossible to:
- Compute the envelope correlation between two different files (eg. one EMG sensor and the cortex, one scout and the rest of the brain)
- Compute 1xN measures without computing the full NxN graphs: which in practice makes it impossible to compute a brain map of HENV measures (1x15000) because it would require the computation of the full 15000x15000 graph.

These are standard use cases, therefore these restrictions that are too important for what we're expecting to provide. At the moment, we can't replace the old AEC processes with the new HENV processes because of this technical limitation.

--
Attached: the processes for 1xN and AxB computations, once `bst_henv.m` is made more generic:
[processes_henv.zip](https://github.com/brainstorm-tools/brainstorm3/files/8003591/processes_henv.zip)

ftadel:
fixes #499


ftadel:
@HosseinShahabi 
I thought you would open a PR for discussing these changes, but I guess there was some misunderstanding at the last meeting. 
Anyway, let's try to do this here: I pushed in this branch the latest changes you submitted in issue #499, together with the necessary updates to other functions to adjust to these changes.

The results obtained are very different from the previous version. I illustrate this using the connectivity tutorial: https://neuroimage.usc.edu/brainstorm/Tutorials/Connectivity#Envelope_correlation

For the same signals and same input parameters, here are the differences observed for the interaction 1-3, in the Delta band, in the first 5s window. **Before~0.16** (left) and **after~0.01** (right) your recent changes. 
image image

HosseinShahabi:
Hi @ftadel 
Thanks for PR and testing. It looks weird cause all other elements are identical and only one has changed. I assume there might be some element displacement? Do you see the difference when using other connectivity measures like Coherence? What if we have something like 4-5 inputs. If you can give me one more simulation with other metrics and a different number of inputs in another band, It can be more helpful. 
I download the new files this weekend and try to work on them to see what is the issue.

ftadel:
> It looks weird cause all other elements are identical and only one has changed. I assume there might be some element displacement?

They are not identical... Here are the same files, but with the "absolute values" option disabled.
image 
image
I uploaded the simulated signals and the old/new henv results here: [Simulation.zip](https://github.com/brainstorm-tools/brainstorm3/files/8230070/Simulation.zip)
But you just have to run the script to reproduce the environment: https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/script/tutorial_connectivity.m

> Do you see the difference when using other connectivity measures like Coherence? 

Interestingly, this shows other bugs...
 - **Coherence**: it saves complex values in the TF field, and the results cannot be displayed.
image

- **oenv / morlet**: different TF values AND different time vectors (old and new results can't be displayed together).
image 
image

ftadel:
@HosseinShahabi Any updates?

PR #524, opened 2022-03-15
Reading CTF comp from fiff file

gabrielbmotta:
I work at mgh under @mshamalainen. I'm trying to read a fiff file with raw data that has ctf compensator information (Fiff tag kinds 3580, 3581, 3582), but was unable to. To get the data loaded in, I:

- Renamed unused input variable 'one' in fiff_rename_comp to 'comp', the output variable, due to:

comp

- Moved the call to fiff_rename_comp in fiff_read_ctf_comp to after the 'data' member of the input variable gets populated, due to:

data

After these changes I was able to load in the file. I noticed these same functions are used in mne-matlab, so this might also be present there.

### DIFF
--- a/external/mne/matlab/fiff_read_ctf_comp.m
+++ b/external/mne/matlab/fiff_read_ctf_comp.m
@@ -103,7 +103,6 @@ for k = 1:length(comps)
     one.save_calibrated = calibrated;
     one.rowcals = ones(1,size(mat.data,1));
     one.colcals = ones(1,size(mat.data,2));
-    one = fiff_rename_comp(one, ch_rename);
     if ~calibrated
         %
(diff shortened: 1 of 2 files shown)


ftadel:
Thanks for for the code updates.

These functions are primarily part of the mne-matlab toolbox, which is used in Brainstorm or other programs.
You should open the same PR on the mne-matlab repo: https://github.com/mne-tools/mne-matlab

If you need to submit other fixes in the future: open the PR on the mne-matlab repo, as it is supposed to be reviewed by the mne-matlab authors, not by us. Then please let us know when it's merged that you updated the library so that we can report the changes in Brainstorm.

PR #525, opened 2022-03-15
Allow to export FOOOF measures as SPM volume and surface

rcassani:
- Add the function `GetFooofValues` in `bst_memory` to extract the a FOOOF measure. It can be externally called.
- Add FOOOF case in `process_export_spmsurf`  and `process_export_spmsurf`
- Validation was performed for the created `gii` and `nii` files


ftadel:
When we discussed about this topic, we agreed on implementing this as a 2-step procedure:
- a new process that "flattens" a FOOOF file into a regular file, similar to `process_extract_pthresh.m` for stat files
- then use the existing unmodified processes for exporting surfaces/volumes

This has the advantages of:
- Allowing running all of Brainstorm functions on the FOOOF results (eg. averaging, statistics, automated screen captures...) - as I'm sure someone will ask soon for reusing the FOOOF models for some other purposes
- Not adding complicated and verbose options to any existing process

Is there still a chance to convince you that working directly on a more versatile approach is better, or do you have other strong arguments in favor of this approach? :-)

rcassani:
I did not want to create an extra process, but it's correct that will have advantages.
I'll reorganize it, so to export the user needs to call two processes.

Do you think it's a good idea to keep the `GetFooofValues` function in bst_memory to be called from the flattening process?

ftadel:
It would be nice to have this illustrated at the end of the FOOOF tutorial as well.
Maybe something equivalent to the documentation of the process "Apply statistic threshold":
https://neuroimage.usc.edu/brainstorm/Tutorials/Statistics#Convert_statistic_results_to_regular_files

Thanks!

rcassani:
@ftadel, changes are done!
The process is in the `Frequency` category after `Group in time...`, though, this may not be best place.

ftadel:
I moved the menu a bit higher:
image

I also added a short explanation for the process:
image

Please merge the PR whenever you're happy with it.

PR #527, opened 2022-03-17
Connectivity: Dynamic load of input files for coherence

ftadel:
@rcassani @HosseinShahabi 

This PR has the objective of loading the input files within bst_coh_2021 instead of pre-loading everything in advance, and therefore filling up the memory. 

Instead reading the files in bst_connectivity, it saves the function call used to read the file, and passes it to bst_coh_2021, which executes the loading call when needed. This approach should be scalable to any number of trials, and should solve the issue mentioned by @rcassani in the tutorial: 
https://neuroimage.usc.edu/brainstorm/Tutorials/CorticomuscularCoherence#Coherence:_EMG_x_Scouts

I had to move around all the input checks in bst_coh_2021, because now we don't know the size of the data at the beginning of the execution of the function. We only know the number of samples and signals when we actually execute the call to load the data, and we never have all the trials loaded in memory at once anymore.

Does it seem like a good solution?
Could you please double-check it works on some examples?

### DIFF
--- a/toolbox/connectivity/bst_cohn_2021.m
+++ b/toolbox/connectivity/bst_cohn_2021.m
@@ -93,44 +93,19 @@ Messages = [];
 % Get current progress bar position
 waitStart = bst_progress('get');
 
-%% ===== Total number of windows =====
 % Number of Files
 nFiles = length(Xs);
 % Window length and Overlap in samples
 nWinLen  = round(WinLen * Fs);
 nOverlap = round(nWinLen * Overlap);
-% Pair files in Xs and Ys must have same nSamples  
-nSamplesXs = cellfun('size', Xs, 2);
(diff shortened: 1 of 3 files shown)


ftadel:
@rcassani 
Was it also the purpose of this branch?
https://github.com/brainstorm-tools/brainstorm3/tree/coherence2021-src
Can we delete it or is there is there some useful code pending for review in there?

rcassani:
@ftadel, 
Yes, it had the same purpose.
Yes, we can delete. Most of those changes have been merged in #472, #488 and 6217358

rcassani:
@ftadel, I ran some tests, and it works well

ftadel:
I removed the comment at the end of the CMC tutorial.
And now you can get back to illustrating the differences between scout before/after on this dataset :)

PR #530, opened 2022-03-26
Intan importer update (.rhd)

mpompolas:
Intan has inefficient loader (read_Intan_RHD2000). The entire recording is loaded every time by default.
The previous importer (2018) modified Intan's importer to provide efficient importing.
The new importer uses Intan's 2022 importer, but converts the recording to a .bst for efficient loading.

Discussion: https://neuroimage.usc.edu/forums/t/importing-events-from-intan-rhd-recordings-and-saving-them-to-brainvision-core-data-format-edf/34029/10


ftadel:
I made the following modifications:
- Renamed `read_Intan_RHD2000_*` functions to get a better idea of what is coming from where
- Fixed a typo (.pl2?)

PR #531, opened 2022-03-30
compute tissues volume from the FEM mesh

tmedani:
Hi @ftadel & @rcassani,

This PR includes an option that computes the brain tissues volume from the tetra mesh. 
It adds a menu item to the right-click on the FEM mesh, then > Compute mesh volume >  displays the results as text using the "view-text" functions 
 
image

This feature can be useful for some applications that compare the brain's volume (ages, ratio WM/GM, ...).  
(here is an example: https://iopscience.iop.org/article/10.1088/2057-1976/ac0547/pdf#:~:text=Accompanying%20this%2C%20the%20brain%2Dto,skull%2C%20rather%20than%20brain%20conductivity.)

We Could also implement similar computation from the segmented MRI ("Tissue"). 

Thanks

### DIFF
--- a/toolbox/process/functions/process_fem_mesh.m
+++ b/toolbox/process/functions/process_fem_mesh.m
@@ -1255,3 +1255,38 @@ function NewFemFile = SwitchHexaTetra(FemFile) %#ok<DEFNU>
         NewFemFile = fem_tetra2hexa(FemFullFile);
     end
 end
+
+
+
+%% ===== COMPUTE FEM MESH VOLUME =====
+function volume = ComputeFemVolume(FemFile)
+    % Get data in database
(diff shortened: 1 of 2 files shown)


tmedani:
> I'm merged the PR because it works technically and I guess you needed it.

Thank you :)

> Shouldn't the volumetry always be estimated from the voxel count of the segmented MRI volume, rather than from the heavily downsampled representation of the segmentation?

Your right, it's not the best solution to compute accurately the volume from the mesh, it's just an approximation. There are better methods to get the volume (using freesurfer/spm...). 
This feature is discussed in some papers using the FEM/TES/EEG, then I added it. It will use in some other applications.

PR #534, opened 2022-04-13
PCA fix sign 1PC

rcassani:
- Handle the sign of the 1st PC


Moo-Marc:
Hi, Sorry I missed the last meeting, but this scaling doesn't seem to me like the correct "fix" for this.  I'm going to spend more time looking into this this week, but I think there's basically an incorrect assumption of independence between sources by doing PCA like this. Like we had discussed, I'll also look at doing the forward or inverse model "orientation average". For now, I'd leave out the scaling proposed here.

Moo-Marc:
Right.  I mean "orientation" of forward/inverse solution vectors in high-dimensional sensor space.

rcassani:
The power scaling has been moved out of this PR

rcassani:
This change is currently handle in #559

Issue #536, opened 2022-04-28
Duplicate events process 


guiomar:
Hi! 

Is there an option to duplicate events through a process using the Run button? 
I see there is the option in the events panel_record, and most of these options can also be acessed through a process using the Run button, but not the duplicate. What would be the most efficient way of doing it?

rcassani:
Hi @guiomar, it is possible to use the Merge events process (`process_evt_merge.m`) to this end, with the following parameters:

- **Events to merge**: `event-to-copy, event-to-copy`
- **New event name**: `copy-of-event`
- Uncheck **Delete the original event**

I agree, it seems better to have specific process for this. @ftadel?

ftadel:
Process "Merge events" transformed into "Duplicate / merge": https://github.com/brainstorm-tools/brainstorm3/commit/0e87b06fdc97278f878e8eafaebe0a5625021d98

PR #537, opened 2022-04-28
Add process to duplicate events

guiomar:
Hi!

I have tried to do a `process_evt_duplicate` funtion, related to issue #536.
Could something like this work? It seems to work on my side.
(I didn't know how to set the new color)

Thanks!


toolbox/process/functions/process_evt_duplicate.m
```diff
@@ -0,0 +1,148 @@
+function varargout = process_evt_duplicate( varargin )
+% PROCESS_EVT_DELETE: Delete a list of events.
```
juangpc, line 2: Shouldn't this be: "ROCESS_EVT_DUPLICATE: Duplicate a list of events. ?

juangpc:
Hi Guio, ✋🏻 
Long time no see. Hope you are doing well.

guiomar:
Hola Juan!! Muchos besos :)

toolbox/process/functions/process_evt_duplicate.m
guiomar: Chenged now :)

ftadel:
Hi Guio,

I would like to propose an alternative to adding yet another process, while we can adjust the existing ones to do what we want.
The number of processes in Brainstorm is slowly becoming an annoying bottleneck: it takes more and more time to parse the process folders and to create the process menus in the Pipeline editor. On older computers (including my laptop), it's already becoming quite annoying. Therefore, whenever we can with little loss of usability, I would prefer to add features without adding process functions/menus.

I propose this change to transform the process "Merge events" into "Duplicate / merge events", as these are really in essence the same operations: https://github.com/brainstorm-tools/brainstorm3/commit/0e87b06fdc97278f878e8eafaebe0a5625021d98

image

I'll close the PR and issue for now.
Please feel free to reopen it if you think there is something that is not handled correctly with my commit.

guiomar:
Yes! Sounds very good! I didn't know this trick that @rcassani explained using the merge function :)
Clarifying it in the name will work nicely. For the duplicate funtion do you have to write the event twice above, or has it been modified to work just with one label? 
Just discard my PR. Thanks a lot!!

ftadel:
> For the duplicate funtion do you have to write the event twice above, or has it been modified to work just with one label?

I removed the test that was forcing the process to have two input events.
So you should be able to make it work just like on the screen capture I posted above.

guiomar:
Awesome!

PR #538, opened 2022-04-28
PCA fix scale power

rcassani:
- Scales the power of the 1stPC to be the same as the average power as done in MNE-Python:
https://github.com/mne-tools/mne-python/blob/31a83063557fbd54d898f00f9527ffc547888395/mne/source_estimate.py#L2818

### DIFF
--- a/toolbox/math/bst_scout_value.m
+++ b/toolbox/math/bst_scout_value.m
@@ -272,19 +272,13 @@ function [F, explained] = PcaFirstMode(F)
     [U,S,V] = svd(F, 'econ');
     S = diag(S);
     explained = S(1).^2 / sum(S.^2);
-    % Find where the first component projects the most over original dimensions
-    [tmp__, nmax] = max(abs(U(:,1))); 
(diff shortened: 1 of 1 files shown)


Moo-Marc:
I'm not sure it's very useful, other than looking more similar to mean. But if we want to keep a meaningful scale comparable to mean, here's what I'd suggest.
For mean, the sum of weights = 1, whereas for a (PCA) component/projection, the sum of squared component values = 1.
To compare mean to PCA, we can multiply the mean weights by sqrt(N), and we get sum(w^2) = 1, a valid component. To go the other way, we can divide PCA by sqrt(N). 
This is simple and maintains a meaningful relative power between the two options.

ftadel:
If you need additional inputs on this topic, put it on the agenda of the next meeting. 
Dimitrios, Richard or John might be able to help.

rcassani:
> we can divide PCA by sqrt(N).

Yeah, this makes more sense than scaling the 1PC to match the same power of all N signals.
Can you add it in #559?

PR #546, opened 2022-05-14
Recompute tess2mri_interp when changing default MRI

### LINKED ISSUE #543: Surface registration after resampling
When resampling an MRI (from 0.8 mm voxel size to 1mm voxel size), the coregistration between the resampled MRI and the surface extracted from the original MRI is lost. I think it might be expected but it might be good to display a warning / and or to remove the surface.

Edouard2laire:
fix #543

### DIFF
--- a/toolbox/tree/tree_callbacks.m
+++ b/toolbox/tree/tree_callbacks.m
@@ -3153,7 +3153,11 @@ end
 
 
 %% ===== CLEAN SURFACE =====
-function SurfaceClean_Callback(TessFile, isRemove)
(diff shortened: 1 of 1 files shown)


ftadel:
I realized you submitted this after fixing the code (see the issue).
Sorry...

Do you still need this after the other fix?
Otherwise, please close the PR.

Edouard2laire:
Hello Francois, 

no problem :) Your fix is actually better since it keeps the consistency between changing the default anatomy from the GUI or from the script (db_surface_default). My fix only works when changing from the GUI :)

PR #547, opened 2022-05-16
Enforce to display "No units" for SSP and ICA component plots 

rcassani:
The time series and the scalp maps of the components obtained with either ICA or SSP (PCA) are in arbitrary units. It's their product the one that has the same units as the original data. This should be explicit. 

See forum question: https://neuroimage.usc.edu/forums/t/flat-time-series-components/34408

This PR, enforces the display of "No units" in the colorbar (displayed for topography) and the scale in the time series figure. 

* An example of the time series of SSP components taking the units of the data can be seen in the introduction tutorial. 
  image
  Plot of the first 4 SSP components for `blink`

* An example of the scalp maps of ICA components taking the units of the data can be seen in the epilepsy tutorial:
  image
  Topoplot for 1st ICA component


ftadel:
I organized this fix in a different way: https://github.com/brainstorm-tools/brainstorm3/commit/fb9c40393f611179fe02eb87265bdb39feb8cf1f

PR #548, opened 2022-05-21
Force intermediate R matrix to [nA x nB x nTime x nFreq]

ftadel:
Issue detected in forum post: 
https://neuroimage.usc.edu/forums/t/envelope-correlation-n-x-n-2020-dimensions-of-arrays-being-concatenated-are-not-consistent/30358/3

In the large SWITCH block between the different connectivity measures, the dimensions of the R matrix that was generated was not same across measures. Possible dimensions included `[nA*nB, nTime, nFreq]`, `[nA, nB, nTime, nFreq]`, `[nA, nB, nFreq]`.

This made the behavior of the following operations ambiguous (handling unconstrained sources, final reshaping).
The error reported in the forum was due to the HENV measure returning a R matrix `[nA*nB, nTime, nFreq]`, which was not compatible with the processing of the unconstrained sources, which required `[nA, nB, nTime, nFreq]`.

The proposed change enforces the switch block to produce a R matrix with full dimensions `[nA, nB, nTime, nFreq]`.

@rcassani Can you please review the PR and double-check that this doesn't seem to break anything obvious?
Thanks!

### DIFF
--- a/toolbox/connectivity/bst_connectivity.m
+++ b/toolbox/connectivity/bst_connectivity.m
@@ -323,6 +323,7 @@ for iFile = 1:length(FilesA)
     
     
     %% ===== COMPUTE CONNECTIVITY METRIC =====
+    % The sections below must return R matrices with dimensions: [nA x nB x nTime x nFreq]
     switch (OPTIONS.Method)
         % === CORRELATION ===
         case 'corr'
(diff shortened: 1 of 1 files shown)


rcassani (approved):
It looks good!
Tested metric computation and visualizations

PR #549, opened 2022-06-01
Feature: Keep selected Subject node across exploration modes

rcassani:
This branch and PR draft was created for feature request by @Moo-Marc 

* c9ec74c: If one node is selected, remember that node for each of the exploration modes. This update seems to be missing, as the selected nodes are only updated if the tree is empty: [panel_protocols.m#L426](https://github.com/brainstorm-tools/brainstorm3/blob/a83914bc934a169bd6406d810b383b182584aadc/toolbox/gui/panel_protocols.m#L426)

* e54fcd3df6e5d9200b63c6516e264087d002554f: Add the field `Layout` to track the previous exploration mode

* f694993874b294fa901d1f75cf07d92f80894833: On change of exploration mode, it checks if the previous selected node was a subject, if so, the same subject should be selected in the new exploration mode.   
  Currently, `GetNode` returns no-nodes; @ftadel, could you take a look to this part of the code?


ftadel:
> Currently, GetNode returns no-nodes; @ftadel, could you take a look to this part of the code?

There is no issue with GetNode I think. 
What is not working is the part that saves a selected node. 
In "functional" mode, it was trying to look for a subject node, which exists only in the "anatomy" mode.

I only had a quick look, but it is not really clear to me what you are trying to do with the saving of "previous node". Is the idea to open the correct subject when switching from the "functional data" to the "anatomy" tabs?

Moo-Marc:
> Is the idea to open the correct subject when switching from the "functional data" to the "anatomy" tabs?

Yes.  I was switching back and forth between these tabs editing fiducials and viewing them on the head surface with the "check registration" figure (something I'd like to discuss Monday).  Right now, the tree always goes to the same selected files each time you switch tabs.  

If each mode remembers its last selected file, it would be good enough for me.  I'd have to switch subjects on each tab, but only once.  The added feature to keep the same subject when a subject is selected is a bonus, but not as obvious for users, and we rarely have subjects selected while working - usually a file.

ftadel:
It feels important to me that:
- when we change subject in one view, it should change it in the other views
- the same file gets selected in the two functional tabs (it helps a lot navigating between the two) - I don't see the added value of selecting different files in both views

I would extend #550 with your code from #549 to keep the selection of the actual files (not only the study/subject folder like I did so far).
If the subject doesn't change in the Anatomy view, we don't update the selection of the file in the functional views (and vice versa). If the subject changes, we make sure something from the same subject is selected in the other views.

Does it make sense?

ftadel:
I updated #550 and tried to integrate what you wanted to keep from here.
Am I missing something?

rcassani:
This feature is on development on #550

PR #550, opened 2022-06-02
Keep study/subject selection across display modes

ftadel:
Replaces PR https://github.com/brainstorm-tools/brainstorm3/pull/549


ftadel:
@Moo-Marc @rcassani 
If you have very large databases with lots of subjects and/or trials, it would be interesting that you check the computational impact of this change before accepting it, by putting a tic/toc around the call to UpdateSearchSelStudy in panel_protocols.m (line 841).
Thanks

Moo-Marc:
I didn't time it, but switching tabs still appears instantaneous with >300 subjects (OMEGA, which currently only has imported anatomy, head surfaces, and links to raw MEG files.  Only the initial display of each tab is a bit slow (5-10 seconds).

ftadel:
@Moo-Marc  @rcassani 
What about something like this?

rcassani:
More something like 8f5d3bd
@Moo-Marc can you test it to see if it's what we agreed?

* Anatomy <--> FunctSub (or FunctionalCon), only keep subject node if it was selected
* FunctSub <--> FunctCon, keep any selected file 
* FunctSub --> FunctCon, if selected Subject>Condition, select Condition>Subject 
* FunctCon --> FunctSub, if selected Condition>Subject, select Subject>Condition
* Any other combination, would remember the last selected node in each exploration mode

rcassani:
* 36a6ed8: revert 8f5d3bd

* e818bbe:
  * FunctSub <--> FunctCon, keep any selected file
  * FunctSub --> FunctCon, if selected `Subject>Condition`, select `Condition>Subject`
  * FunctSub --> FunctCon, if selected `Subject`, do nothing (keep previously selected node)
  * FunctCon --> FunctSub, if selected `Condition>Subject`, select `Subject>Condition` 
  * FunctCon --> FunctSub, if selected `Condition`, do nothing (keep previously selected node)

PR #552, opened 2022-06-09
View 3D head surface when editing MRI fiducials

Moo-Marc:
Additional button (next to zoom and cancel/save) in the MRI panel for computing and displaying a head surface while editing the MRI fiducials.

image

The yellow dot is linked to the crosshairs / current location, the green dots to "set" fiducials.


ftadel:
I improved the robustness of the code and added the possibility to click on the surface to select the slices in the MRIViewer.
Please test it a bit more and let me know when you think this is ready for merging.

Moo-Marc:
Love it!  I was also thinking to eventually add the surface selection.  It's great.
The only tiny thing I'm not sure if we can fix: when closing the MRI figure with the window X button, the cancel callback isn't triggered and the 3D head figure stays open.
Otherwise, I think it's ready to merge.

ftadel:
> The only tiny thing I'm not sure if we can fix: when closing the MRI figure with the window X button, the cancel callback isn't triggered and the 3D head figure stays open.

Done.

I also added a paragraph to the tutorials:
https://neuroimage.usc.edu/brainstorm/Tutorials/ImportAnatomy#Using_the_MRI_Viewer

PR #551, opened 2022-06-09
Udpate coil_def file with Fieldline and Kernel opm sensor defs

juangpc:
Hi, 

At @jcmosher attention, coil_def.dat was found to be missing Fieldline and Kernel sensors' definitions. 

These changes make the file equivalent to https://github.com/mne-tools/mne-python/commit/323ab3fd683d81be55bf328d7c2bc8083c4ec104#diff-679288d606bdfddab221f5dc41c7119faf600b7c8b8c93173120102dcad8a604, by @larsoner. 

For a more stable solution, a ci action could check for updates in mne-python's version of this file. @ftadel if you want, I could do that.

PR #553, opened 2022-06-10
Github Action to automate check for changes on coil_def file

juangpc:
A long term solution to automatically check for updates for MNE-Python's version of coil_def.dat seems interesting. This option takes advantage of the free cloud service provided by Github to check for changes in the file and generate an issue.

See discussion in #551. 

The scripts runs in about 10 seconds or so.
In the future, other parallel jobs can be created completely independent of this one.
If differences are detected between ```coil_def.dat``` stored in MNE-Python and the one stored in this repo, an issue is created similar to this one: https://github.com/juangpc/brainstorm3/issues/7

The job is right now configured to run every day of the week, at 9am UTC. This can be easily changed. See lines [3-4 of the yml file)](https://github.com/juangpc/brainstorm3/blob/44dce088a2e66c1295429d5f9797399eb90d481f/.github/workflows/coil_def_check.yml#L4)

If no differences are detected, nothing happens. 

If an error occurs during the execution of the Github Action, an email is received. This option can be optionally turned off in the settings menu.

This solution is stable and extensively used in many other repositories. Externally the only change is the appearance of a ```.github``` folder in the repository.

If this turns out to be not the best option, feel free to drop this pr. Hope it helps.

### DIFF
--- a/.github/workflows/coil_def_check.yml
+++ b/.github/workflows/coil_def_check.yml
@@ -1,0 +1,49 @@
+name: 'coil_def.dat Update Checker' 
+on: 
+  schedule:
+    # runs at 9:00h UTC, any day of the month, any month, on Monday.
+    - cron: '0 9 * * 1'
+
+permissions:
+  issues: write
+
+jobs:
+  Compare-coil_def-updates:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone brainstorm repo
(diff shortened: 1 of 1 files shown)


ftadel:
Looks great! Thanks!
I'm curious to see it working, but it might not happen before a few years from now...

PR #555, opened 2022-06-27
Bugfix: Handle `badSeg` when out-of or half-in the user-specified `TimeWindow`

rcassani:
Extend the bugfix added in #498, because if the user specified a `TimeWindow` it could lead to negative indices. 

In this PR: 
 * Bad segments out of the user-specified `TimeWindow` are ignored in making `Fmask`
 * Bad segments that are partially in the user-specified `TimeWindow` are constrained and used to make `Fmask`. However, if the resulting bad segment is not a segment but a point, it is ignored in making `Fmask`

### DIFF
--- a/toolbox/process/functions/process_evt_detect.m
+++ b/toolbox/process/functions/process_evt_detect.m
@@ -197,34 +197,20 @@ function OutputFiles = Run(sProcess, sInputs) %#ok<DEFNU>
         end
         
         % ===== BAD SEGMENTS =====
-        % If ignore bad segments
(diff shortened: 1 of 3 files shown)


ftadel:
It's now a lot of code copy-pasted between the three functions.
Maybe it would be nice to factorize the entire block of code "===== BAD SEGMENTS =====", and make it a function available in one of the three process functions. 
What do you think?

rcassani:
> What do you think?
It makes sense, I'll update this PR reflecting that

Issue #225, opened 2019-07-25
Leff ratio in dipole scanning and whitened recordings


ftadel:
@jcmosher suggested we change the scaling factor applied in two functions: 

**Dipole scanning**

https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/process/functions/process_dipole_scanning.m#L119
`Factor = sqrt(DataMatP.Leff / sResultP.Leff);`
Replaced with:
`Factor = sqrt(sResultP.Leff);`

Problem, when testing with the introduction tutorials, the goodness of fit of the dipoles drops to very small values, example for standard condition at 90ms (Leff=193):
![image](https://user-images.githubusercontent.com/6920058/61877817-13423180-aef0-11e9-87f6-b5c836e95fc6.png)

Current code (left) vs. John's suggestions (right):
![image](https://user-images.githubusercontent.com/6920058/61877875-294ff200-aef0-11e9-9665-e1289f0719c5.png)  
![image](https://user-images.githubusercontent.com/6920058/61877910-353bb400-aef0-11e9-9581-07b667b1f2d0.png)

**Whitened recordings**

https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/tree/tree_callbacks.m#L3280
`Factor = sqrt(DataMat.Leff / ResultsMat.Leff);`
Replaced with:
`Factor = sqrt(sResultP.Leff);`

This looks more reasonable. Below, figures obtained by right-clicking on the MN source link for the standard condition (Leff=193) > Model evaluation > Save whitened data (left=before, right=after modification):
![image](https://user-images.githubusercontent.com/6920058/61878151-c1e67200-aef0-11e9-8bb6-225a626031a8.png)   
![image](https://user-images.githubusercontent.com/6920058/61878938-4e456480-aef2-11e9-84d5-869969c24e64.png)

@jcmosher @dimitriospantazis please validate whether this is what you want to obtain.

jcmosher:
https://neuroimage.usc.edu/brainstorm/Tutorials/TutDipScan

ftadel:
@jcmosher  ??

ftadel:
@jcmosher @dimitriospantazis @sbaillet 
If you have time now, can you please have a look at this?

ftadel:
@jcmosher @dimitriospantazis @sbaillet
Do you just want to ignore this question until someone spots a possible bug?

ftadel:
This issue is still pending....

[...]

Issue #141, opened 2018-11-29
Group analysis: recommended workflow and statistics


ftadel:
The pipeline currently described on the website is not giving acceptable results, and we could not publish any statistics at the group level in the recent Frontiers article. This needs to be addressed urgently.

Two online tutorials to fix:
https://neuroimage.usc.edu/brainstorm/Tutorials/Workflows#Constrained_cortical_sources
https://neuroimage.usc.edu/brainstorm/Tutorials/VisualGroup#Group_analysis:_Sources

Unconstrained sources must be addressed as well. We are now recommending most EEG users to use unconstrained dipole orientations, but we have no option at all to offer for group analysis.

ftadel:
@richardmleahy @HosseinShahabi 
Weren't you working on this?
Statistics are still the major black hole of the software...

ftadel:
Ping

ftadel:
Still some major work needed on wrapping up the recommended pipeline for unconstrained sources...

ftadel:
Forum question: https://neuroimage.usc.edu/forums/t/unconstrained-to-flat-map-with-pca-method-details-on-the-process/36054

PR #559, opened 2022-07-12
Global PCA for scouts and/or unconstrained sources

Moo-Marc:
EDIT (2023-03-20): The PR has gone through a few iterations and the Implementation section here is now outdated.  See comments from 2023 below.

**Description**

This PR implements computing PCA across all files/epochs at once ("global pca") instead of separately for each file ("per-trial PCA").  It applies to extracting scout time-series, either through process_extract_scout or indirectly (e.g. for connectivity on scouts), and to "flattening" unconstrained source orientations to one source per vertex.

**Justification**

Following tests of various PCA procedures on the introductory tutorial data, it seems the feature of  has the following advantages:

1. It avoids sign flipping issues between trials.  In particular, currently, PCA for unconstrained sources fails because of this, and the [solution](https://github.com/brainstorm-tools/brainstorm3/pull/534) proposed by @rcassani for scouts does not work for unconstrained source flattening.
2. It seems on par with per-trial PCA for extracting source activity across different conditions (e.g. many "standard" vs few "deviant" trials in the introduction tutorial), and should be more robust with very few or very short trials (though tests show per-trial is pretty robust too).  
3. It may simplify interpretation of comparisons when the same component is applied across conditions.
4. It allows visualising the scout component maps.
5. It can be used to do both scout and flattening in a single step, which is slightly better than doing flattening separately first, and noticeably better than doing scout extraction on x,y,z separately first.  Here "better" means it keeps more source power, which is what PCA maximises.  It achieves this by taking into account the full scout source covariance matrix, whereas doing it in 2 steps ignores some of the source pairs (e.g. location 1 x vs location 2 y).

Among these points, only 1 is pretty serious, but I think it all adds up to a good case for including this feature, and even using it as the default instead of per-trial PCA.  A warning about the sign issue should also be added to per-trial PCA for flattening.

**Implementation**

This initial PR is a rough draft, how I implemented it for testing only.  In particular, it has a bunch of extra options that would be removed (pcagt, pcag3, pcagx).

Currently, the functions that do PCA for scouts or flattening often see only one file at a time.  Therefore, I implemented this using the data covariance matrix.  The user would first compute the data covariance on all trials, and that would be used to compute the global PCA and applied to each file or to file averages, depending on the use case.  Unfortunately, this means the same PCA decomposition is computed repeatedly each time it is applied to a file, but this was the simplest way to do it without creating new files (e.g. PCA kernels), or new fields (e.g. in the channel file, or inverse kernel file).  And it is reasonably fast enough, equivalent to the current per-trial PCA.

I'm open to suggestions on how else we could implement this.

**Documentation**

Do we want to save somewhere a summary of the tests and results that show how the various PCA procedures perform?  I have a couple slides now, but I could for example save it as a forum post which we could link to in the tutorials?


ftadel:
> Even though it's been like this for some time, I think the broken flattening PCA should be fixed asap.

The PCA option is not used in any of our recommended pipelines, so it is not an emergency to do anything with it.
All the things that are in the standard pipelines is much more urgent to fix (like the statistics for instance: https://github.com/brainstorm-tools/brainstorm3/issues/141, https://github.com/brainstorm-tools/brainstorm3/issues/225...)

> Given this, I propose that I simplify this PR to only keep the global PCA option for flattening as a first step. Once reviewed and merged, I can introduce the rest of the changes for later review. Once the code is cleaned, I think it shouldn't be too long to review. 

Sounds good.

Moo-Marc:
Really trimmed down the changes to only unconstrained source flattening as a first step, for easier review.
Only process_source_flat has access to the PCA function for orientations at the moment.

toolbox/inverse/bst_source_orient.m
```diff
@@ -12,14 +12,16 @@
 %                     GridAtlas.Scouts(i).Region(2) is the source type (V=volume, S=surface, D=dba, X=exclude)
 %                     GridAtlas.Scouts(i).Region(3) is the orientation constrain (U=unconstrained, C=contrained, L=loose)
 %    - SourceValues : [Nvertices x Nsensors] or [Nvertices x Ntime], source values
-%    - Function     : Name of the function to apply to group multiple components {'sum', 'sum_power', 'rms', 'max', 'pca', 'mean'}
+%    - Function     : Name of the function to apply to group multiple components {'sum', 'sum_power', 'rms', 'max', 'pca', 'pcag', 'mean'}
 %    - DataType     : Type of data being processed {'data', 'results', 'scouts', 'matrix'}
 %    - RowNames     : Description of signals being processed: {empty, array of doubles, array of cells}
+%    - SourceCov    : [Nvertices x Nvertices] Covariance matrix between rows of SourceValues. Used with Function='pcag' only.
```
ftadel, line 18: From what I understood, you aimed to pass a [Nsources x Nsources] matrix here.
This doesn't work: it's going to be way too big. Most users don't have 64Gb of RAM.
You need to optimize this input. Maybe passing the data covariance + the kernel, and optimize the computation so that it never computes the full source covariance

toolbox/math/bst_scout_value.m
```diff
@@ -234,17 +270,7 @@
...
-            switch (size(F,3))
-                case 2
-                    Fs(1:2:end) = F(:,:,1);
-                    Fs(2:2:end) = F(:,:,2);
-                case 3
-                    Fs(1:3:end) = F(:,:,1);
-                    Fs(2:3:end) = F(:,:,2);
-                    Fs(3:3:end) = F(:,:,3);
-            end
+            Fs = permute(Fs, [3,2,1]);
```
ftadel, line 273: Is what you removed and what you added really equivalent here?

toolbox/math/bst_scout_value.m
```diff
@@ -264,27 +290,20 @@
...
-    sign_omaxx = sign(F(nmax,i_omaxx));
-    % Sign of maximum in first component time series
-    [Vmaxx, i_Vmaxx] = max(abs(V(:,1)));
-    sign_Vmaxx = sign(V(i_Vmaxx,1));
-    % Reconcile signs
-    F = sign_Vmaxx * sign_omaxx * S(1) * V(:,1)';
-    % Add the weighted average back to the signal
-    % The mean is tderived from the original signal averages weighted by their respective contributions to the first singular vector
-    F = F + sign_Vmaxx * sign_omaxx * U(:,1)' * Fmean;
+    U = U(:,1);
```
ftadel, line 298: You're actually modifying the function, making the software not backward compatible, and older results not reproducible.

In our preliminary discussion, one request I had was to add new options and keep the old ones working with no modification.
Please reorganize your developments so that a previous execution of `process_source_flat` keeps on producing the exact same results as before even with the new version of the software, and that newly created scripts have an additional option to compute your new PCAg function.

toolbox/process/functions/process_source_flat.m
```diff
@@ -61,47 +63,78 @@
...
+                SharedFile = bst_process('GetNewFilename', bst_fileparts(sStudy.FileName), [FileType '_KERNEL_' fileTag]);
+                bst_save(SharedFile, ResultsMat, 'v6');
+                db_add_data(sInputs(iInput).iStudy, SharedFile, ResultsMat);
+                panel_protocols('UpdateNode', 'Study', sInputs(iInput).iStudy); 
+                % Find links to the result kernel that was just created
+                % This includes bad trials (same as process_inverse I think).
+                OutputFiles = db_links('Study', sInputs(iInput).iStudy);
+                OutputFiles(~contains(OutputFiles, file_short(SharedFile))) = [];
+                return;
+            end
```
ftadel, line 124: This logic doesn't work.
1) This is a very bad practice to have a return in a middle of a loop.
2) If the user passes multiple files that do not relate to the same ImagingKernel, you need to group them by inverse solution file, loop on each group, and produce one file in output per group.
Please reorganize the code of this function in a cleaner way.

toolbox/process/functions/process_source_flat.m
```diff
@@ -48,8 +49,9 @@
                                        'Method used to perform this conversion:'];
     sProcess.options.label1.Type    = 'label';
     sProcess.options.method.Comment = {'<B>Norm</B>: sqrt(x^2+y^2+z^2)', ...
-                                       '<B>PCA</B>: First mode of svd(x,y,z)'};
-    sProcess.options.method.Type    = 'radio';
+                                       '<B>PCA</B>: First mode of svd(x,y,z)<BR>(requires pre-computed data covariance across epochs)'; ...
```
ftadel, line 52: Use has to be warned that this is PCAg.
Also, Sylvain requested to have access to both PCA (trial-wise) and PCAg.

ftadel (changes requested):

toolbox/math/bst_scout_value.m
Moo-Marc: The option I add of not removing the baseline here is off by default.  Everything is still compatible and reproducible.

toolbox/process/functions/process_source_flat.m
Moo-Marc: This first PR was only to fix the problematic sign-flipping issues.  We don't want to keep previous bugged code just for reproducibility, right?  I will add all the options we discussed in the next PR.

toolbox/math/bst_scout_value.m
Moo-Marc: Anyway, it wasn't used yet in this PR so I guess it can be added back later.  But this option will be important: when users compute their data covariance matrix, they chose a baseline to remove the offset.  It's not a good idea to again remove the offset later on the whole trials without the user knowing about it.

toolbox/process/functions/process_source_flat.m
```diff
@@ -65,8 +67,9 @@
     OutputFiles = cell(1, numel(sInputs));
     % Get options
     switch(sProcess.options.method.Value)
-        case 1, Method = 'rms';  fileTag = 'norm';
-        case 2, Method = 'pcag';  fileTag = 'pca';
+        case {1, 'norm'}, Method = 'rms';  fileTag = 'norm';
+        case {2, 'pca'},  Method = 'pca';  fileTag = 'pca';
+        case 'pcag',      Method = 'pcag'; fileTag = 'pcag';
```
Moo-Marc, line 72: I don't like pcag as a file tag.  "global PCA" is vague (it was originally about doing flattening and scout together, which didn't have an obvious name).  We should simply call this option PCA across trials (PCA-a?) vs PCA per trial (PCA-t?).

toolbox/math/bst_scout_value.m
```diff
@@ -1,8 +1,7 @@
-function [Fs, Comp] = bst_scout_value(F, ScoutFunction, Orient, nComponents, XyzFunction, isSignFlip, scoutName, Cov)
+function [Fs, PcaOrient] = bst_scout_value(F, ScoutFunction, Orient, nComponents, XyzFunction, isSignFlip, scoutName, Cov)
```
Moo-Marc, line 1: It doesn't matter much, but this name won't make as much sense once I add back PCA across trials for scouts.  It also outputs a PCA component, but it's no longer an orientation.

toolbox/math/bst_scout_value.m
Moo-Marc: Yes.  Here there's a single source (scout function was applied), so it's size 1,nTime,nComp.  Examples: 
```
>> a = 1:5; a(:,:,2)=6:10;
>> a = permute(a,[3 2 1])
a =
     1     2     3     4     5
     6     7     8     9    10
>> a = 1:5; a(:,:,2)=6:10; a(:,:,3)=11:15;
>> a = permute(a,[3 2 1])
a =
     1     2     3     4     5
     6     7     8     9    10
    11    12    13    14    15
```

toolbox/math/bst_scout_value.m
ftadel: Please feel free to change it to something else that makes more sense.
I just felt that `Comp` was not clear enough, I tried to make it more readable.

toolbox/inverse/bst_source_orient.m
ftadel: I'm not sure I understand why you marked this as resolved.
Can you please explain? 
Thanks

toolbox/math/bst_scout_value.m
ftadel: I'm sorry, I don't understand your reply.

I was pointing at the fact that you replaced:
```
    S = diag(S);
    explained = S(1).^2 / sum(S.^2);
    % Find where the first component projects the most over original dimensions
    [tmp__, nmax] = max(abs(U(:,1))); 
    % What's the sign of absolute max amplitude along this dimension?
    [tmp__, i_omaxx] = max(abs(F(nmax,:)));
    sign_omaxx = sign(F(nmax,i_omaxx));
    % Sign of maximum in first component time series
    [Vmaxx, i_Vmaxx] = max(abs(V(:,1)));
    sign_Vmaxx = sign(V(i_Vmaxx,1));
    % Reconcile signs
    F = sign_Vmaxx * sign_omaxx * S(1) * V(:,1)';
    % Add the weighted average back to the signal
    % The mean is tderived from the original signal averages weighted by their respective contributions to the first singular vector
    F = F + sign_Vmaxx * sign_omaxx * U(:,1)' * Fmean;
```

with:
```
    % Signal decomposition / Remove average over time for each row
    [U, S] = svd(bst_bsxfun(@minus, F, mean(F,2)), 'econ');
    S = diag(S);
    explained = S(1).^2 / sum(S.^2);
    U = U(:,1);
```

I have a hard time believing that "Everything is still compatible and reproducible".
If it is not possible to produce previously obtained results with an unmodified script, it is not reproducible.

toolbox/process/functions/process_source_flat.m
ftadel: > We don't want to keep previous bugged code just for reproducibility, right?

Yes, I think we do, whenever we can.
In my understanding, this is what reproducibility means.
For the past 10 years, we continuously updated this code with newer version, that invariably was declared "bugged" a few months or years later. Whatever code you implement now will probably be declared "bugged" as well, at some point. It is important to keep track of what we do with unstable code like this, so that: 1) old scripts keep on producing the same results, and 2) a future student in 2030 will be able to re-run all the previous attempts and finally decide on what is the best solution that was proposed over the years.

But maybe I'm not aiming at the right targets.
Since at the moment we don't recommend using PCA, I'd be less worried to modify the existing methods. But still, that would break the notion of reproducibility.

@sbaillet @rcassani 
What is your opinion on this topic?

toolbox/process/functions/process_source_flat.m
ftadel: I have no opinion on the tag that is used.
Please decide on what you find is best, or ask @rcassani or @sbaillet for suggestion.

But if the old tag is "pca", then the new one should be something else to be able to differentiate the files in the database and the methods in the scripts.

ftadel (changes requested):

toolbox/inverse/bst_source_orient.m
Moo-Marc: Oh, sorry, my comment didn't save.  I will change the comment.  It's actually 3x3xNvertices here.

toolbox/math/bst_scout_value.m
Moo-Marc: I see.  I thought it was about the non-baseline removal option which you removed here.

This change is the sign-flip fix from Raymundo.  I agree with keeping many things as (sometimes hidden) options, but not what is clearly broken.  For reproducibility, with version control, users can use any previous version.  Of course we don't currently have version numbers/tags, so it would be the responsibility of users to note the Brainstorm version they used.

toolbox/process/functions/process_source_flat.m
Moo-Marc: > We don't want to keep previous bugged code just for reproducibility, right?

I meant that with the understanding that everything exists in the version control system, so reproducibility is always possible.  I would suggest only keeping "live", things that should still be used.  In a case such as this, it's not just a better implementation or newer method, it was not doing what users (and we) expected.  I call that a bug.  Even if a user has started doing PCA and wants to continue, we have a responsibility to let them know that the previous implementation was broken and shouldn't be used.

toolbox/math/bst_scout_value.m
ftadel: As much as I can, for the standard recommended pipelines, I would like the same script to be able to run identically even with different versions of Brainstorm. Except indeed for obvious bugs. 
However, this PCA thing is not that clearly **a bug**, as it's been modified multiple times with the approval of all the team, and there is no clear ground truth to any dimension reduction technique.

I'm fine with considering this as a bug, and just replace the code as you suggested, but I'd like to get the explicit approval from the authors of this code (@sbaillet @dimitriospantazis @jcmosher @richardmleahy) that they all agree that what they wrote is _a bug_ and that your modified version is _not a bug_.

Unfortunately, I don't think this can be discussed over github, as they would not read through all this and dive in the code.
Could you please prepare a demo to illustrate your suggested code replacement, and present it at the meeting next Wednesday? (you could send an email to at least Sylvain and Richard to put it officially on the agenda).
Thanks

Moo-Marc: Sure let's confirm in a meeting. But this was already presented in detail to the team by @rcassani a while ago, with https://github.com/brainstorm-tools/brainstorm3/pull/534. I don't think anybody remembered why it was done the way it was previously, but the goal was to get a consistent sign. Raymundo clearly showed the sign flips between epochs, leading to the bad results he first saw with PCA as scout function in connectivity. And he showed his solution works well on scouts.

> no clear ground truth to any dimension reduction technique

True, but that's not the issue here. PCA has a sign ambiguity, independently of what it is used for (here dimension reduction). If we apply it on each epoch separately, we have to be consistent in the choice of sign. Otherwise, any further computation that combines the epochs in any way (averaging, connectivity, etc.) will not give the expected result.

toolbox/process/functions/process_source_flat.m
Moo-Marc: Agreed. I'll add the two options in this PR. I'm working on it. 
Recall that the per-trial option will still change since it also requires a sign flipping solution, different than the one for scout PCA.

toolbox/process/functions/process_source_flat.m
Moo-Marc: Changed as requested.  
I wasn't sure if there was a better way to get the kernel file than extract it from the "link|kernel|data" composite FileName. Would be nice if it were saved as its own field.

Moo-Marc:
Requested changes were made, so ready for another review once the "bugness" question is decided. 
I added back flattening PCA per epochs, with sign correction based on PCA across epochs.
Some changes were made for efficiency, but it still takes quite a while to run flattening PCA per epochs: ~10 minutes for 192 epochs, vs seconds for across epochs which only runs once and saves a kernel for all files.
Tested, and confirmed similar results (left per epoch, right across epochs):
Flattening_160ms_20220919

toolbox/process/functions/process_source_flat.m
ftadel: Please use `file_resolve_link`, it returns the kernel file + the data file.

toolbox/process/functions/process_source_flat.m
```diff
@@ -61,47 +68,115 @@
...
+        sStudy = bst_get('Study', sInputs(iInput).iStudy);
+        % Error: PCA requires kernel and data covariance
+        if strncmpi(Method, 'pca', 3) 
+            if numel(sStudy.NoiseCov) < 2
+                bst_report('Error', sProcess, sInputs(iInput), 'PCA flattening requires data covariance matrix to be computed first.');
+                return;
+            end
+            % Load kernel
+            ResK = in_bst_results(sInputs(iInput).FileName, 0);
+            ResultsMat.ImagingKernel = ResK.ImagingKernel;
```
ftadel, line 124: In this case, you are loading twice the file, and adding and `ImagingKernel` field to a structure that already contains a `ImageGridAmp` field, making it an invalid data structure (it's one or the other).

In the cases you need the ImagingKernel, please load the file only once, with `ResultsMat = in_bst_results(sInputs(iInput).FileName, 1);`

toolbox/process/functions/process_source_flat.m
```diff
@@ -122,8 +197,34 @@
...
+                    error('PCA flattening only available for imaging kernel files.');
+                end
+                Ker = reshape(ResultsMat.ImagingKernel, ResultsMat.nComponents, [], size(ResultsMat.ImagingKernel,2)); % (nComp, nSource, nChan)
+                % Get data covariance matrix for provided data file.
+                if numel(sStudy.NoiseCov) < 2
+                    error('PCA flattening requires data covariance matrix to be computed first.');
+                end
+                DataCov = load(file_fullpath(sStudy.NoiseCov(2).FileName)); % size nChanAll
+                % Keep 3x3 covariance per location: 3x3xnSource instead of (3*nSource)^2
+                OrientCov = squeeze( sum(bsxfun(@times, sum(bsxfun(@times, permute(Ker, [3,4,1,2]), DataCov.NoiseCov(ResultsMat.GoodChannel, ResultsMat.GoodChannel)), 1), ... % (1, nChan, nComp, nSource)
```
ftadel, line 213: On the tutorial dataset, this line generates the following error on my computer:
```
Error using bsxfun
Requested 274x274x3x15002 (25.2GB) array exceeds maximum array size preference (15.9GB). This might cause MATLAB to become unresponsive.
```

We can't assume that users have 25Gb of RAM available for this process.

toolbox/math/bst_scout_value.m
```diff
@@ -13,6 +12,12 @@
 %     - XyzFunction    : String, function used to group the the 2 or 3 components per vertex: return only one value per vertex {'norm', 'pca', 'none'}
```
ftadel, line 12: Add pcaa in the list.
And maybe some description explaining what is happening with the interaction of the two options ScoutFunction/XyzFunction, at the end of the header?

toolbox/process/functions/process_source_flat.m
```diff
@@ -61,47 +68,115 @@
...
+        end
+        % Save as kernel if possible.
+        if ~isempty(ResultsMat.ImagingKernel) 
+            ResultsMat.ImageGridAmp = [];            
+            % For PCA method, save a single shared kernel.
+            if strcmpi(Method, 'pcaa')
+                ResultsMat.DataFile = '';
+                % db_add doesn't make it a shared kernel.
+                %SharedFile = db_add(sInputs(iInput).iStudy, ResultsMat);
+                SharedFile = bst_process('GetNewFilename', bst_fileparts(sStudy.FileName), [FileType '_KERNEL_' fileTag]);
```
ftadel, line 157: Input kernel file: `results_dSPM-unscaled_MEG_KERNEL_220921_1723.mat`
Output kernel file: `results_KERNEL_pcaa_220921_1727.mat`

It would be better if the output filename could keep the tags from the input file, it would help tracking where is the flattened file came from.
Maybe just append "pcaa" at the end of the existing kernel filename?

toolbox/math/bst_scout_value.m
```diff
@@ -214,17 +225,49 @@
...
+            explained = 0;
+            for i = 1:nRow
+                Fi = permute(F(i,:,:), [3,2,1]); % permute faster than squeeze
+                % Here, we will probably want to remove the offset removal, but keeping as before for now.
+                [U, S] = svd(bsxfun(@minus, Fi, sum(Fi,2)./size(Fi,2)), 'econ', 'vector'); % sum faster than mean
+                explained = explained + S(1).^2 / sum(S.^2);
+                PcaFirstComp(:,i) = U(:,1);
             end
+            explained = explained / nRow;
+            %warning('This older PCA implementation suffers from sign flipping between epochs.');
```
ftadel, line 246: If this is not the old implementation: remove the comment

toolbox/math/bst_scout_value.m
```diff
@@ -214,17 +225,49 @@
...
+            % Use PCA across epochs (here in OrientCov) to consistently select the component sign for each epoch. 
+            CompSign = sign(sum(PcaFirstComp .* OrientCov));
+            PcaFirstComp = bsxfun(@times, CompSign, PcaFirstComp);
+            Fs = sum(bsxfun(@times, permute(PcaFirstComp, [2,3,1]), F), 3); % dot product of Comp with F on 3rd dim, gives size (nRow, nTime)
```
ftadel, line 250: Is the goal of this only to be reproducible between two identical calls of the same process on the same files?
Otherwise, for what reason would we enforce two epochs to have their PCA oriented along a similar direction?

It prevents from calling the PCA flattening on a single file, e.g. a subject-level average, resulting of the averaging of multiple files. It is OK, but since this is a one of the recommended workflows (https://neuroimage.usc.edu/brainstorm/Tutorials/Workflows#Unconstrained_cortical_sources), it requires some adjustments in the procedures if you want to make this modified process compatible with them.

ftadel (dismissed):
Some more things to fix.

Another thing to think about:
Where would you see this documented in the tutorials?
(at the moment, it is not documented at all...)

toolbox/process/functions/process_source_flat.m
```diff
@@ -133,6 +234,7 @@
     switch(Method)
         case 'rms',  fileTag = 'norm';
         case 'pca',  fileTag = 'pca';
+        case 'pcaa',  fileTag = 'pcaa';
```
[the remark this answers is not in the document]
Moo-Marc: **A**cross epochs

toolbox/math/bst_scout_value.m
```diff
@@ -32,10 +37,14 @@
 % For more information type "brainstorm license" at command prompt.
 % =============================================================================@
 %
-% Authors: Sylvain Baillet, Francois Tadel, John Mosher, 2010-2016
+% Authors: Sylvain Baillet, Francois Tadel, John Mosher, 2010-2021
+%          Marc Lalancette, 2022
 
 % ===== PARSE INPUTS =====
-if (nargin < 7) || isempty(scoutName)
```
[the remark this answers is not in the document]
Moo-Marc: Right, noted.

toolbox/math/bst_scout_value.m
Moo-Marc: Kept it commented for now if we need to keep the old implementation, since it's not decided yet.

toolbox/math/bst_scout_value.m
Moo-Marc: This is the PCA sign ambiguity question, which we need to make consistent both between calls and importantly between epochs so that averaging, connectivity, etc. doesn't cancel out when combining epochs.

> It prevents from calling the PCA flattening on a single file

Not really. Recall that it uses the data covariance to compute the pcaa. The call just needs to provide it, which process_source_flat now does. I'll search if there are other ways this gets called for flattening.  Maybe we'll have to remove "shortcuts" where it's incorporated into other processes and require it being done separately.

toolbox/math/bst_scout_value.m
```diff
@@ -265,26 +299,18 @@
...
-    % Add the weighted average back to the signal
-    % The mean is tderived from the original signal averages weighted by their respective contributions to the first singular vector
-    F = F + sign_Vmaxx * sign_omaxx * U(:,1)' * Fmean;
+    U = U(:,1);
+    % Correct sign of the first PC to be consistent across epochs as best we can.
+    % For scouts: choose sign to have positive correlation with orientation-corrected (isSignFlip=true) mean timeseries.
+    %sign_meancorr = sign(mean(F,1) * (U' * F)');
+    % Mathematically equivalent simpler expression, using the definition of the eigen-decomp: 
+    % U*S^2*U' = F*F' => U1' * F * F' * ones = S1^2 sum(U1)
+    %sign_meancorr = sign(sum(U(:,1)));
```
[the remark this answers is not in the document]
Moo-Marc: I'll try.  But it's all an explanation of the sign correction in the following line. Since the math simplifies a lot, it's not clear why sign(sum(U)) makes sense, so it's important to keep this.

toolbox/process/functions/process_source_flat.m
```diff
@@ -48,9 +49,15 @@
...
     sProcess.options.label1.Type    = 'label';
     sProcess.options.method.Comment = {'<B>Norm</B>: sqrt(x^2+y^2+z^2)', ...
-                                       '<B>PCA</B>: First mode of svd(x,y,z)'};
-    sProcess.options.method.Type    = 'radio';
-    sProcess.options.method.Value   = 1;
+                                       '<B>PCA</B>: First mode of svd(x,y,z)<BR>(requires pre-computed data covariance)'; ...
+                                       'norm', 'pca'};
+    sProcess.options.method.Type    = 'radio_label';
+    sProcess.options.method.Value   = 'norm';
+    sProcess.options.pcamode.Comment = {'once across epochs', 'separately for each epoch', 'PCA computation:'; ...
```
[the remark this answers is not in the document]
Moo-Marc: > 1. and you should therefore make it conditional to the selection of "pca" in the first option

I wanted to, but I think the conditional stuff Martin added only works with checkboxes as a controller, right?

toolbox/process/functions/process_source_flat.m
```diff
@@ -61,47 +68,115 @@
 
 
 %% ===== RUN =====
-function OutputFiles = Run(sProcess, sInput) %#ok<DEFNU>
-    OutputFiles = {};
+function OutputFiles = Run(sProcess, sInputs) %#ok<DEFNU>
+    OutputFiles = cell(1, numel(sInputs));
```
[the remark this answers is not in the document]
Moo-Marc: {} Even if it ran successfully for some of the files before the error?  Or you meant it shouldn't contain empty cells, i.e. grow as we loop?

toolbox/math/bst_scout_value.m
Moo-Marc: > what is happening with the interaction of the two options ScoutFunction/XyzFunction

Yes that will be important, but let's wait until the next PR that will add changes to the scout function PCA.  For both PCAs, flattening should be done first, which is the opposite of what this function currently does.  So there's more to think about.

toolbox/math/bst_scout_value.m
Moo-Marc: I verified that no other function can currently call this with XyzFunction 'pca'.  I'm making a note that we should eventually review each of these to see what options we want, once PCA is settled.

- view_scouts, process_extract_scout, process_source_atlas: norm or none
- bst_timefreq: sum (but TODO for complex values)
- process_canoltymap: sum for TF, mean for ERP
- process_pac, process_pac_dynamic: mean
- bst_connectivity: max
- bst_memory: min for p-values, absmax for t-values, max for stat, rms for results

toolbox/process/functions/process_source_flat.m
Moo-Marc: Right, I can use kernel only for pcaa now that I only save a kernel. And for pca I can load the full data and discard the kernel after I have PcaAcross.

toolbox/process/functions/process_source_flat.m
```diff
@@ -61,47 +68,115 @@
...
-    % Save on disk
-    bst_save(OutputFiles{1}, ResultsMat, 'v6');
-    % Register in database
-    db_add_data(sInput.iStudy, OutputFiles{1}, ResultsMat);
 end
 
 
 %% ===== COMPUTE =====
-function ResultsMat = Compute(ResultsMat, Method, Field)
+function [ResultsMat, PcaOrient] = Compute(ResultsMat, Method, Field, sStudy, PcaOrient)
```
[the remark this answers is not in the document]
Moo-Marc: Will do.

toolbox/process/functions/process_source_flat.m
Moo-Marc: Done (in next commit)

toolbox/process/functions/process_source_flat.m
```diff
@@ -122,8 +197,34 @@
...
-            % Set the number components
+            OrientCov = [];
+            Ker = [];
+            if strcmpi(Method, 'pcaa') 
+                if ~isfield(ResultsMat, 'ImagingKernel') || isempty(ResultsMat.ImagingKernel)
+                    error('PCA flattening only available for imaging kernel files.');
+                end
+                Ker = reshape(ResultsMat.ImagingKernel, ResultsMat.nComponents, [], size(ResultsMat.ImagingKernel,2)); % (nComp, nSource, nChan)
+                % Get data covariance matrix for provided data file.
+                if numel(sStudy.NoiseCov) < 2
```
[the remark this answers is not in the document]
Moo-Marc: That was in case it were called from elsewhere.  Removed in next commit.

toolbox/process/functions/process_source_flat.m
```diff
@@ -122,8 +197,34 @@
         % Unconstrained source models
         case 3
             % Apply source orientations
-            ResultsMat.(Field) = bst_source_orient([], ResultsMat.nComponents, [], ResultsMat.(Field), Method);
-            % Set the number components
+            OrientCov = [];
+            Ker = [];
+            if strcmpi(Method, 'pcaa') 
+                if ~isfield(ResultsMat, 'ImagingKernel') || isempty(ResultsMat.ImagingKernel)
+                    error('PCA flattening only available for imaging kernel files.');
```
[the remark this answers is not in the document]
Moo-Marc: Also tested in the calling function.  Removed in next commit.

toolbox/process/functions/process_source_flat.m
Moo-Marc: Indeed.  We'd need pagemtimes here.  I found a way to loop once on components instead. It should stay under 100MB.

toolbox/math/bst_scout_value.m
ftadel: > > It prevents from calling the PCA flattening on a single file
> Not really. Recall that it uses the data covariance to compute the pcaa. The call just needs to provide it, which process_source_flat now does. 

The two problems I see with the current process:
- It requires a kernel, and is therefore not applicable to any source file that was post-process, or averaged across runs (which only have ImageGridAmp and no ImagingKernel, e.g. Z-scored maps)
- In the case of averages between multiple runs, there is no data covariance available in the folder, the process crashes, while in principle, the method you propose should be applicable to single files, with no data covariance and no imaging kernel.

I'm sure this limitation would cause problems in some pipelines, and would cause misunderstanding and productivity  because this limitation cannot be anticipated.
I think that for the versatility of the process, the processing stream should be refined, for example:
- If there is no ImagingKernel: use the ImageGridAmp, ignores the data covariance, and saves only one single flattened file
- If there is only one file in input (or two? or less than half of the trials of the folder? or any other heuristic?): save independent files instead of kernel files that create again tons of links. 

Use case: for connectivity, you typically want to flatten individually all the trials. But for group-level statistics, you need to flatten only the averages. It would get very quickly annoying if when processing only the average, either it crashes (because there is no Kernel) or it recreates hundreds of new unwanted links for all the trials, making the source files more difficult to select programmatically.

toolbox/process/functions/process_source_flat.m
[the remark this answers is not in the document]
ftadel: No, it works with radio buttons as well (at least since I fixed it).

But I recommend you don't lose time in doing this. The usability would be definitely improved with only ONE flat list of THREE radio buttons, instead of 2+2 radio buttons with conditional behaviors.

toolbox/process/functions/process_source_flat.m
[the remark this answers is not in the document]
ftadel: If there is one input file that can't be processed, you can decide on one of the two behaviors below:
1) It triggers an error: the process stops immediately calls `return`, no files are returned so that the calling script is informed that a blocking error occurred. OutputFiles must be set to {} before the function returns.
2) It triggers a warning: the process doesn't stop, it moves to the next iteration of the `for` loop with `continue` instead of `return`, reports a `Warning` instead of an `Error`. The corresponding item in the OutputFiles cell-array must be set to the original file (no empty cell allowed in the returned cell-array). 

The second option is less restrictive, but might not correspond to what we need here. In a full processing script, if the file can't be processed and is still with 3 components, you just can't move on as if nothing happened because the next process might crash in an ugly way or produce weird/unwanted results, causing a bug that might be difficult to detect. It's usually safer to generate clear errors, forcing the user to fix the issues, instead of allowing uncontrolled behaviors.

The only case where you can trigger a warning instead of an error is if the file is already flat (nComponents=1).

toolbox/process/functions/process_source_flat.m
ftadel: I think you should implement the two streams in parallel: 
- reading a kernel => saving a kernel
- reading a single file => saving a single file (no use of ImagingKernel)

toolbox/process/functions/process_source_flat.m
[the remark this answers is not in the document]
Moo-Marc: Ok I'll reset it before error returns.  
Note that yesterday I randomly looked at process_timefreq and bst_timefreq (I had them open) and it looked to me that it was returning a partially filled OutputFiles on an error.

toolbox/process/functions/process_source_flat.m
Moo-Marc: It's not that simple.  We now always need the kernel to apply to the covariance matrix, even with pca where we use pcaa to fix the sign.  So both pca/a will no longer be possible on non-kernel files.  pcaa saves a shared kernel, pca saves a single-file kernel.  Next commit coming today.

toolbox/math/bst_scout_value.m
Moo-Marc: > But for group-level statistics, you need to flatten only the averages.

It's true that would be ok to do.  But if we allow running on single trials, it becomes very difficult to properly warn the user to never do it if there's subsequent steps that depend on the individual trial sign consistency.  (Actually, if doing group averaging, there's a question of between-subject sign consistency as well.  This requires further thought.)  Instead of allowing single trials because it's valid in some cases, I'd suggest not allowing and instead the flattening would have to be performed earlier in the pipeline, with a kernel, if users want to use PCA.

Moo-Marc: > save independent files instead of kernel files

We only save a **shared** kernel when doing pcaa, which implies it's valid on all the data from the data covariance matrix.  Yes users can run it on a single file (actually makes sense since it only needs to run once) and it creates all the links.  For selecting the sources, there's now the added 'pcaa' part in the files.  I see these as minor inconveniences.  I believe this fits the philosophy of Brainstorm to not aim to maximize versatility, but ensure proper use of few valid options, right?

In terms of user friendliness and having fewer steps and links etc, I still think the best approach down the road is to have this PCA flattening as an option in the inverse process source model.  For now, I'd stick to have it as is (kernel only).

toolbox/process/functions/process_source_flat.m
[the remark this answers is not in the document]
Moo-Marc: ok for flat. I can show it in our next Monday meeting and get feedback.

Moo-Marc:
Thanks Francois for all the useful comments. 

Regarding tutorials, how about a section for flattening right after the [Unconstrained orientations](https://neuroimage.usc.edu/brainstorm/Tutorials/SourceEstimation?highlight=%28unconstrained%29#Unconstrained_orientations) section in the source estimation tutorial, or in the advanced section?

toolbox/process/functions/process_source_flat.m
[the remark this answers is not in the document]
ftadel: process_timefreq/bst_timefreq is indeed very messy in terms of error handling.
But it never returns empty cells in OutputFiles.

If you prefer returning partially filled OutputFiles (until it crashes), it would not change anything.
But please avoid risking having only warnings and empty cells: this would cause problems in bst_process.

toolbox/process/functions/process_source_flat.m
ftadel: We should be able to compute the pca without any data covariance.
The data covariance/kernel should be only an optimization of the computation/storage, not a requirement.

toolbox/math/bst_scout_value.m
ftadel: So we wouldn't be able to flatten a single non-kernel file (a Z-scored average, possibly imported directly from another program) with PCA? The current version of the PCA "works well" in this case. Are you suggesting to remove this option from the possible pipelines?

It looks like there is some more work to be done to explain how this PCA fits in all the possible workflows involving unconstrained sources.

ftadel:
> Regarding tutorials, how about a section for flattening right after the [Unconstrained orientations](https://neuroimage.usc.edu/brainstorm/Tutorials/SourceEstimation?highlight=%28unconstrained%29#Unconstrained_orientations) section in the source estimation tutorial, or in the advanced section?

Maybe an advanced section just after: https://neuroimage.usc.edu/brainstorm/Tutorials/SourceEstimation#Averaging_normalized_values

toolbox/math/bst_scout_value.m
Moo-Marc: Ok, I thought a bit more about it and we can add non-kernel files.  

There will be more options.  I think it deserves a separate "PCA options" panel, as it will apply to scout PCA too:

- PCA "type": across epochs, per epoch with sign consistency, (per epoch with arbitrary sign - if we want to keep it)
- Use pre-computed data covariance (much faster, but requires kernel files only)
- For covariance: baseline and active time windows, DC offset.
  Possibly some options like the averaging process, deciding how to group:
- Eventually, for multiple subjects: "group PCA" (like group ICA)? or separately per subject
- For multiple studies: weighted or equal averaging

Is there a recommended way to sort inputs so they are ordered by subject, study, and condition?  I was thinking text sorting by FileName. 

For single non-kernel files though, we'd have to give a strong warning.  If the user wants to contrast two averaged conditions for example, they should not run them separately or the sign might not match. 

For group averaging, it's another level of difficulty for matching signs.  Something to be investigated further.  Probably we want to warn about this for now, at least in the tutorials.

ftadel: > PCA "type": across epochs, per epoch with sign consistency, (per epoch with arbitrary sign - if we want to keep it)

This can added as 2 (or 3) options for flattening or for scout function, no need to add any extra panel.

> Use pre-computed data covariance (much faster, but requires kernel files only)
> For covariance: baseline and active time windows, DC offset.
> Possibly some options like the averaging process, deciding how to group:
> For multiple studies: weighted or equal averaging
> 
> For single non-kernel files though, we'd have to give a strong warning. If the user wants to contrast two averaged conditions for example, they should not run them separately or the sign might not match.

This sounds too complicated. We can't ask the users to fill in a full window of options in order to get their scouts time series. What we propose must remain accessible to the average user, unless you want to make this an advanced option only accessible to experts and not recommended anywhere.
Can't you simply re-use the data covariance available in the database for this file, instead of asking all the input parameters to compute it again? And does it make sense to compute a data covariance from an average anyway?

If you can't make it easily accessible without a full panel of options, just drop this option of making it accessible to non-kernel files.

> Eventually, for multiple subjects: "group PCA" (like group ICA)? or separately per subject
> For group averaging, it's another level of difficulty for matching signs. Something to be investigated further. Probably we want to warn about this for now, at least in the tutorials.

For group analysis, the source signals are always flattened and **rectified** - no sign problem anymore. Please try to limit the scope of the solutions you're exploring to what we actually need, which is documented in the Workflows page:
https://neuroimage.usc.edu/brainstorm/Tutorials/Workflows#Unconstrained_cortical_sources

> Is there a recommended way to sort inputs so they are ordered by subject, study, and condition? I was thinking text sorting by FileName.

You can group the files using the field SubjectFile or ChannelFile of the input structures. All the possible grouping solutions are coded in `process_average.m`.
Simple examples grouping by channel file (which is maybe the only thing you need) are available in various processes, like `process_headpoints_add` or `process_montage_apply`. Search for `{sInputs.ChannelFile}`.

Moo-Marc: > Can't you simply re-use the data covariance available in the database for this file

No, not without a kernel.  That's why I had it this way to start.

> This sounds too complicated.

It's not that bad, but we could hide the last two things for now, so just have the covariance options.  I wanted to add the baseline/active window anyway to replace the automatic DC offset removal that used to be done for PCA.

> compute a data covariance from an average

No, I average the covariances. 

> For group analysis, the source signals are always flattened and **rectified**

That's one solution, and the one I'd prefer in many cases.  But is that just the recommended workflow or is it enforced in group averaging?  Would it for example prevent averaging of signed connectivity results without applying abs?


I'm almost done with this change.  It wasn't small, but it will now properly deal with files from multiple subjects (treat them separately for now), and multiple conditions, which I hadn't considered yet but it's good to combine them.  It will save a shared kernel for conditions where all files are links, or single-file kernels or full results for other files.  It's longer if there are non-links because we have to loop over those files twice.

ftadel: > But is that just the recommended workflow or is it enforced in group averaging?

Heavily recommended, only solution documented, but not enforced.

>  Would it for example prevent averaging of signed connectivity results without applying abs?

It would not prevent any averaging or statistical tests on signed data, not even generate a warning.
It is only for source-level data that this is recommended.
For the moment, there is no clear workflow documented for group-level statistics on connectivity measures.

Moo-Marc:
Hi @ftadel,
The code for PCA flattening is ready for review.  It now works for non-kernel files, and across conditions/studies. (recall I kept non-kernel files for now since they'll be needed for scouts).  If you can have a look to see how the flattening process is now organized and think a bit about scouts, I'd appreciate your input.  I'm not sure yet what's the best approach since there are 2 extra levels of looping in process_extract_scout (scouts and frequencies).

There's now a new PCA options panel:
PcaFlatProcess
PcaOptions

There's one useability bug I haven't yet figured out: when modifying options in the panel and clicking ok, it goes back to the process window. If we click the Edit button again, the panel options are reset, it doesn't keep the previous settings.

Moo-Marc:
Note that I did pretty thorough tests of various inputs and settings.  In particular, I confirmed the results are identical when using pre-computed data covariance or not, with the same settings.

Moo-Marc:
Finally the PR is ready for review again.

Scout and flattening PCA are implemented and all connectivity processes updated.  Most of the file manipulation for PCA across files is now done in a new bst_pca function.

Still to do: 

- Documentation.
- Possibly: change bst_process to do the temporary scout extraction instead of each process.
- Time-frequency processes that use scouts.

toolbox/math/bst_scout_value.m
```diff
@@ -1,290 +1,485 @@
...
+%     - Fs           : Combined time series. [Ncomponents x Ntime] for ScoutFunction only, [Nsources x Ntime] for XyzFunction, or [1 x Ntime] for both.
+%     - PcaFirstComp : First mode of the PCA, as column(s). [Nsources, Ncomponents] for ScoutFunction only, [3 x Nsources] for XyzFunction, or [Nsources * Ncomponents, 1] for both.
+%
+% NOTES: 
+%     ScoutFunction is applied before XyzFunction. But when both are PCA, they are done simultaneously.
+%     For XyzFunction 'pca', it should be applied before extracting scouts, or at the same time with the same PCA function.
+%     For ScoutFunction 'pca' with nComponents > 1, XyzFunction must match, unless the old deprecated 
+%       behavior is desired where scouts are extracted separately on x,y,z.
+%     For ScoutFunction 'pca', the PCA component (combination of sources) is scaled to 1/sqrt(nVertices)
+%       to match the scaling of 'mean', and to be more easily comparable between scouts.
```
rcassani, line 29: It seems such scaling is not applied for `PCA` nor for `FastPCA` ScoutFunctions

toolbox/math/bst_noisecov.m
```diff
@@ -86,6 +90,8 @@
     strComment = 'Noise covariance';
 end
 
+% Don't show progress bar if called with 1 file
+isProgress = length(DataFiles) > 1;
```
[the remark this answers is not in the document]
Moo-Marc: The progress increments as it loops through the input files. For one file it just appears and disappears (rather quickly), which is just not useful. Plus it may interfere with any other progress bar that was ongoing: this was the case when I was using it for pca, but no longer.

toolbox/process/functions/process_extract_scout.m
```diff
@@ -50,23 +50,33 @@
...
     sProcess.options.isnorm.Comment = 'Unconstrained sources: Norm of the three orientations (x,y,z)';
     sProcess.options.isnorm.Type    = 'checkbox';
     sProcess.options.isnorm.Value   = 0;
     sProcess.options.isnorm.InputTypes = {'results'};
+    sProcess.options.isnorm.Class   = 'notpca';
     % === CONCATENATE
     sProcess.options.concatenate.Comment = 'Concatenate output in one unique matrix';
     sProcess.options.concatenate.Type    = 'checkbox';
     sProcess.options.concatenate.Value   = 1;
+    sProcess.options.concatenate.Class   = 'notpca';
```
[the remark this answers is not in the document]
Moo-Marc: I actually don't know when it's ever useful, why it was added.  It could eventually be added back, but it complicates the already many possibilities of output files with PCA...  So I thought we should discuss if it would actually be useful before doing it.

toolbox/math/bst_pca.m
```diff
@@ -0,0 +1,777 @@
...
+if isScout && ~isempty(sResults.Atlas)
+    Message = 'Scout PCA already present in first input file. Returning all inputs unchanged.';
+    bst_report('Warning', sProcess, sInputs, Message);
+    OutputFiles = {sInputs.FileName};
+end
+
+%____________________________________________________________________________________________
+% First loop over files for 2 reasons: for pcaa and pcai we accumulate covariances in source
+% space (NxN for scouts or 3x3 xyz covariance matrices) across files for the reference
+% components, and for pca and pcai we compute individual file PCA components. This however means
```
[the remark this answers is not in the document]
Moo-Marc: You mean in the history?  I guess we could, like it's done for concatenated output in extract_scout.  Yet this is not done for other processes, like covariance or inverse models, right?

toolbox/math/bst_scout_value.m
Moo-Marc: Not for FastPCA on purpose: I think we'll get rid of it (from the interface).  
For the old PCA, it should be applied.  It's my mistake: I had to temporarily disable it to confirm it was otherwise the exact same result as the old code.  Turned back on in a new commit.

toolbox/math/bst_pca.m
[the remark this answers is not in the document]
rcassani: Yes, something like in History, or in the Options.
Other kernels save information of the noise and data covariance matrices in Options.NoiseCovMat.History and Options.DataCovMat.History respectively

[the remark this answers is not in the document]
Moo-Marc: Ok, I didn't know it was in the covariance files.  Lets do it the same way then, directly in History, to save the list of input files.

toolbox/connectivity/bst_connectivity.m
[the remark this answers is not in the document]
Moo-Marc: > It would be necessary to keep the original `sInputB` to properly set the parent file.

Or we don't delete the result files in that case?  But if we're really not allowed to have those types of result files, I guess we need to.  But bst_connectivity didn't have sInputs until now, only the file names; and we still have the DataFile in the result sInput, so that should be fine using isTempPcaB to check.  I'd have to look more closely at how you've changed FilesB to sInputB though.  

By the way, if we don't keep the function compatible with both Files and sInput as inputs, I think at least there should be an explicit check and clear warning message of how this was changed, just to be safe. 

> It seems the L282 is broken when old PCA is requested

It works for me with PLV for example.  When exactly does it fail?

> At this moment, only the PCA scout function implies flattening before connectivity. Will this be the general approach for other scout functions?

I believe this is indeed what we decided following your tests, to have it as the general approach.  I just didn't consider it part of this PR's scope.  And we need to decide how to implement this: reject unconstrained inputs, or perform it from within the connectivity processes, and if the latter, whether or not to save the flattened files permanently in the tree.

toolbox/process/functions/process_cohere2_2021.m
Moo-Marc, line 1: Needs to be reverted following bst_connectivity accepting sInputs.

toolbox/connectivity/bst_connectivity.m
```diff
@@ -127,8 +132,112 @@
...
+    % FilesA/B are replaced by temporary files as needed by RunTempPca.
+    [FilesA, isTempPcaA, FilesB, isTempPcaB] = process_extract_scout('RunTempPca', OPTIONS.ProcessName, OPTIONS.PcaOptions, ...
+        FilesA, OPTIONS.TargetA, FilesB, OPTIONS.TargetB, strcmpi(OPTIONS.UnconstrFunc, 'pca'));
+    if isTempPcaA
+        sInputToDel = FilesA;
+    end
+    if isTempPcaB
+        sInputToDel = [sInputToDel, FilesB];
+    end
+end
```
rcassani, line 183: Connectivity A×B: cases where flatten is required, but only FilesB are results need to be handled. 
Currently an error in `process_extract_scout>LoadFile` occurs.

rcassani: @Moo-Marc, at first glance, this could be addressed as below. Although you may want to handle it in a different way
```
--- a/toolbox/process/functions/process_extract_scout.m
+++ b/toolbox/process/functions/process_extract_scout.m
@@ -590,6 +590,9 @@ function [isUnconstrained, nComponents] = CheckUnconstrained(sProcess, sInputs,
     isUnconstrained = [];
     nComponents = [];
     if nargin < 3 || isempty(sResults)
+        if ~strcmpi(sInputs(1).FileType, 'results')
+            return
+        end
         % Load first file, without data.
         sResults = LoadFile(sProcess, sInputs);
         if isempty(sResults)

```

toolbox/connectivity/bst_connectivity.m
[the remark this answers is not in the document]
Moo-Marc: For posterity: we decided to add flattening with PCA, before scouts and connectivity, as an option independently of the scout function.

toolbox/connectivity/bst_connectivity.m
```diff
@@ -127,7 +132,130 @@
...
+% Extract scouts with PCA, and save time series in temp files.
+if strcmpi(OPTIONS.ScoutFunc, 'pca') && ~isempty(OPTIONS.PcaOptions) && ~strcmpi(OPTIONS.PcaOptions.Method, 'pca')
+    % Check inputs
+    if (~isempty(FilesA) && ~isstruct(FilesA)) || (~isempty(FilesB) && ~isstruct(FilesB))
+        bst_report('Error', OPTIONS.ProcessName, [], 'When selecting PCA, bst_connectivity now requires sInput structures instead of file names as inputs.');
+        return;
+    end
+    % FilesA/B are replaced by temporary files as needed by RunTempPcaScout.
+    [FilesA, isTempPcaA, FilesB, isTempPcaB] = process_extract_scout('RunTempPcaScout', ProcessName, ...
+        OPTIONS.PcaOptions, FilesA, OPTIONS.TargetA, FilesB, OPTIONS.TargetB);
```
rcassani, line 196: Produces error en case of `1xN`, with `1` = A channel of data, and `N` = N Scouts:

`Unexpected scout definition for running PCA with temporary files.` L1449 in process_extract_scout, as `AtlasListA` is not an Atlas but the channel name for files A.

PR #560, opened 2022-07-15
Faster implementation for wPLI

danielemarinazzo:
I have implemented a version of wPLI that is faster (and more in line with the time resolved one).


ftadel:
Much faster indeed, about 10x!
(with identical results)

Thanks!

PR #563, opened 2022-07-21
resize HA and HB directly instead of making new indices

danielemarinazzo:


ftadel:
It is indeed significantly faster.
Thanks!

PR #564, opened 2022-07-26
Hilbert transform in the frequency domain

danielemarinazzo:
This modification does a zero phase filtering (equivalent of matlab `filtfilt.m` and then performs Hilbert transform in the frequency domain), using some code provided by Ricardo Bruña.
In my opinion this is more agile, makes sure that the filter is zero-phase (not sure about the current one), and is also faster.
Furthermore it eliminates the need to check whether the signal processing toolbox is present, avoiding possible issues in case the implementations should diverge in the future.

On the other hand this would bypass all the possible existing options for filtering (even though I am pretty confident that only the default option is actually used).

So I leave it up to you to see what to make of this.


ftadel:
Our colleagues @HosseinShahabi @richardmleahy @jcmosher worked a lot on designing the specifications of these filters: 
https://neuroimage.usc.edu/brainstorm/Tutorials/ArtifactsFilter#Filter_specifications:_Low-pass.2C_high-pass.2C_band-pass

The idea was initially to use the same filters everywhere in the application, but I guess this is not a necessity.
The authors of these filters should comment on this topic, but it might take a while before we get an answer.

Note that the function `butter` is part of the signal processing toolbox and should be replaced with `oc_butter` if the toolbox is not available or disabled.

Moo-Marc:
My comment from an email discussion about this PR:
Thanks for the suggestion, but we will keep filtering as a separate process for Hilbert.  I understand the logic, but we benefit more in reusing existing (stable) functions where possible; our filtering function was revamped in 2019 and is used everywhere in Brainstorm.  Note that there are already major performance improvements with the recent PR about connectivity (phase metrics and averaging).  The remaining slower steps are mainly those involving file operations and to a lesser extent one or two computation steps requiring large amounts of memory (but not filtering).

PR #565, opened 2022-08-05
Export scout as label

Edouard2laire:


ftadel:
I implemented it like this instead: https://github.com/brainstorm-tools/brainstorm3/commit/0709bc683fc7278938bdeb61c68b446b0a86bf42

The folder external/mne is an external toolbox, we can't add files to it from this repository.

Edouard2laire:
> The folder external/mne is an external toolbox, we can't add files to it from this repository.

The file is coming from MNE matlab GitHub: https://github.com/mne-tools/mne-matlab/blob/master/matlab/mne_write_label_file.m

shouldn't it be part of this toolbox ?

ftadel:
Oops... of course.... I should have checked that before........ I don't why I thought that you created this function yourself....
Sorry for the added mess, your solution was better... anyway it's in there now :)

PR #566, opened 2022-08-07
Scout Projection from one hemisphere to another

Edouard2laire:
Hello, 

This PR is here to implement the scout projection from one hemisphere to another using FreeSurfer Sphere (https://surfer.nmr.mgh.harvard.edu/fswiki/recon-all#ContralateralSurfaceRegistation.28SphericalMorph.29.28-.3Cno.3Econtasurfreg.29) .

The changes are the following : 
Importation from FreeSurfer: 
- Import FreeSurfer sphere (rh.lh.sphere.reg)
- Keep track of how the sphere is downsampled from high-resolution sphere (to be able to reimport other sphere in the future without reimporting all the anatomy)

Projection of scout 
- Add the projection from one hemisphere to the other using FS sphere

This example show projection of the red scout from right to left : 

Capture d’écran 2022-08-06 à 22 20 57

NOTE: 

By default, rh.lh.sphere.reg and lh.rh.sphere.reg doesn't seems to be computed.  -contrasurfreg needs to be specified when launching FreeSurfer. If a subject is already analysed using FreeSurfer, it is possible to relaunch only -autorecon3 (eg : 
recon-all   -subjid subjid -autorecon3 -contrasurfreg [+ OPTIONS]  
It might be possible to achieve this by only calling "mris_register"; but I haven't found how to yet :)

NOTE 2:
Here is the FS folder of the subject (ICBM512) if you want to do some tests without needing to launch FreeSurfer:
https://drive.google.com/file/d/1X2gwZP-gUfycOwJWoyQ-ZZGkUXAA1vyN/view?usp=sharing

Regards, 
Edouard


ftadel:
I was not aware of this contralateral registration in FreeSurfer, it would be a nice feature to add to Brainstorm!

However, it looks like keeping the indices is not needed for projecting the scouts to the contralateral hemisphere: the variable `SamplingIdx` is not used anywhere in a useful way.  Could you please remove that everywhere?
Also: add back the `.Vertices` to `tess_average.m`, no need to change that, then.

Please move the menu inside "Project to...", with the name "Contralateral hemisphere".
The menu must be disabled in case of volume atlases. The function must intercept correctly when this is requested on a non-cortical scout, and report the error with a meaningful error message to the user.

Can this option be used in all cases? Does it add significant computation time?
If yes and no: We should add this option to `process_segment_freesurfer.m`. 
Could you add this to the PR?

When this is merged, we should also edit the tutorials (I'll do that): 
- https://neuroimage.usc.edu/brainstorm/Tutorials/Scouts#Menu:_Scout
- https://neuroimage.usc.edu/brainstorm/Tutorials/LabelFreeSurfer

Thanks!

Edouard2laire:
Hello Francos, 

Thank you for having provided a review so quickly :). I think I addressed all the comment. Let me know if you have any other comment on the PR. 


> However, it looks like keeping the indices is not needed for projecting the scouts to the contralateral hemisphere: the variable SamplingIdx is not used anywhere in a useful way. Could you please remove that everywhere?
> Also: add back the .Vertices to tess_average.m, no need to change that, then.

It has been removed. I think I will open a new PR after this one as I think it might make the importation of other sphere in the future more easy . But it's not mandatory for this PR. 

> Please move the menu inside "Project to...", with the name "Contralateral hemisphere".
> The menu must be disabled in case of volume atlases. The function must intercept correctly when this is requested on a non-cortical scout, and report the error with a meaningful error message to the user.

Done; I am not sure about how to test for non-cortical sources. I tried to look at the function but_project_scout to see how it was handled but I have not found anything useful. 


> Can this option be used in all cases? Does it add significant computation time?
> If yes and no: We should add this option to process_segment_freesurfer.m.
> Could you add this to the PR?

Yes. According to FS website it is adding ~1 hour of computation per hemisphere so I think it is reasonable. With the current code, it can be deactivated by specifying -nocontrasurfreg in the options


> When this is merged, we should also edit the tutorials (I'll do that):

I think one thing that we will get asked (and that the question I am already facing) is : what steps do we suggest for people who already have a Brainstorm database with source analysis and who would like to use this feature ? 

Computing the sphere can be done outside Brainstorm with : `recon-all -subjid subjid -autorecon3 -contrasurfreg [+ OPTIONS]`

The main question is then how to import them in Brainstorm without breaking already done analysis ?  If someone reimport the anatomy with the exact same fiducials; are the previous source-map still valid ? 


Edouard

ftadel:
> Here is the FS folder of the subject (ICBM512)

Do you need this `xhemi` subfolder? Is it generated when using the option `contrasurfreg`?

> I also fixed a bug that was causing cortex-cereb to have a sphere attached to it; even if it should not. I saw in the code that there was already a code having a similar function but that was commented. Let me know if you prefer to keep the sphere even for cortex-cereb.

I'm not sure why this was disabled... 
It is indeed dangerous to have spheres in surfaces that include more than just the two hemispheres, as it would be easy to get the indices mixed up.
Maybe what you did breaks something existing, but I can't recall what, and right now the change seems reasonable to me. We'll change it back it someone complains :)

> It has been removed. I think I will open a new PR after this one as I think it might make the importation of other spheres in the future easier.

I don't think there would be any workflow that would involve importing more spheres after downsampling the surfaces. This would be too complex to track. What is enforced in the current pipeline is: first you run FreeSurfer, then you import the output in Brainstorm. There is no going back and forth between the two.

> Done; I am not sure about how to test for non-cortical sources

By construction, now we shouldn't have spheres if there are non-cortical sources, so ignore this comment.

> I think one thing that we will get asked (and that the question I am already facing) is: what steps do we suggest for people who already have a Brainstorm database with source analysis and who would like to use this feature?
> The main question is then how to import them into Brainstorm without breaking already done analysis ? If someone reimport the anatomy with the exact same fiducials; are the previous source-map still valid ?

Messing around with the existing files is not a good idea, this would be calling for trouble, in case the cortex surfaces are not downsampled in the same way. For the general users, I would say that this is not possible and would recommend to reimport the anatomy and restart the analysis.

For advanced users, maybe something like:
- Creating a new subject,
- importing the anatomy folder with the extra contralateral spheres, 
- copy the field SphereLR to the old cortex surface structure (export the old cortex to Matlab, export the new cortex to Matlab, copy the field, right-click on the old cortex > Import from Matlab > select the modified old structure)
Would that work?

> Maybe a less important point; but are you ok with storing those extra spheres in reg.sphereLR?

I thought about renaming it to "SphereContra", but that was not going to be much clearer...
I think it's OK this way.
The good thing is that by construction we can add any field we want to the "Reg" field, so no worries with the added variable.

ftadel:
One thing I don't like much is the modified visualization.
I think this display with the 4 spheres is not intuitive at all. Regular users will never need to see this, what is interesting is to see how the hemisphere is matched to a sphere. How the two hemispheres are projected between each other is some internal thing that do not need any visual representation (or at least not this one).
I'd prefer to revert `view_surface_sphere.m` to its original version, if that's OK with you.

Can you please double-check my other modifications?

Edouard2laire:
> Do you need this `xhemi` subfolder? Is it generated when using the option `contrasurfreg`?


No: it was based on another attempt I did using the chemo toolbox. I will recompute a clean version of ICBM to make sure xhemi didn't create any other files. 

I am not sure about the changes made in https://github.com/brainstorm-tools/brainstorm3/pull/566/commits/092c0062dc52f998e522ef9f3dd9f415e28a7433

I think we should keep: 
```
%% Add extra-Parameter 
     if ~contains(param,'-contrasurfreg') && ~contains(param,'-nocontrasurfreg') 
         param = [param ' -contrasurfreg'];
     end
```
to make sure contrasurfreg is added to pare even when FreeSurfer is not called from the process

> For advanced users, maybe something like:
> 
> Creating a new subject,
> importing the anatomy folder with the extra contralateral spheres,
> copy the field SphereLR to the old cortex surface structure (export the old cortex to Matlab, export the new cortex to Matlab, copy the field, right-click on the old cortex > Import from Matlab > select the modified old structure)
> Would that work?

Yes, I think that should work.  another way I was thinking was : 
> Creating a new subject,
> importing the anatomy folder with the extra contralateral spheres,
> Transfert the scout from the old cortex to the new one
> Project to the contralateral hemisphere
> Transfert back the scout from to the old cortex

Transfert can be done either by simply doing export/import from Matlab; or if it doesn't work by using projection between subjects. 

> I'd prefer to revert view_surface_sphere.m to its original version, if that's OK with you.

Yes. no problem with me. 


> It is indeed dangerous to have spheres in surfaces that include more than just the two hemispheres, as it would be easy to get the indices mixed up.

Actually, I was wondering if we would not like to preserve the sphere but complete it with NAN for the surface that doesn't have a sphere. The idea is that for cortex_cereb some users might still want to project a scout from the left to the right hemisphere of the cortical surface. Same when projecting from a subject to a template, we can use the sphere for the cortical surface and approximation for the cerebellum. But it can be done in another PR. Especially we would need a function similar to tess_hemisplit but to separate between cerebellum and cortex.

ftadel:
> I think we should keep:
> > %% Add extra-Parameter 
> >      if ~contains(param,'-contrasurfreg') && ~contains(param,'-nocontrasurfreg') 
> >          param = [param ' -contrasurfreg'];
> >      end
> to make sure contrasurfreg is added to param, even when FreeSurfer is not called from the process

This is not a standard parameter for people running FreeSurfer recon-all, therefore it would be weird to force it for everybody. 
When calling directly the `Compute` function without the process, expert users can add in input `param='-contrasurfreg'`, just like in the process options.
The important thing to do is to document correctly this parameter in the tutorials.

> Actually, I was wondering if we would not like to preserve the sphere but complete it with NAN for the surface that doesn't have a sphere. The idea is that for cortex_cereb some users might still want to project a scout from the left to the right hemisphere of the cortical surface. Same when projecting from a subject to a template, we can use the sphere for the cortical surface and approximation for the cerebellum. But it can be done in another PR. Especially we would need a function similar to tess_hemisplit but to separate between cerebellum and cortex.

Indeed. And at the moment, the projections to/from `cortex_cerb` work well, so I reverted to the original logic of `tess_concatenate`: https://github.com/brainstorm-tools/brainstorm3/pull/566/commits/a576ee674f958acbd62388b70972dead3eb2567b

It would require some more work to make a cleaner version of this, but at the moment it seems to work, and we can probably leave it like this for now.

Is there anything else to change before I merge this?

ftadel:
I edited the tutorials, could you have a look to the following sections?
- https://neuroimage.usc.edu/brainstorm/Tutorials/LabelFreeSurfer#Contralateral_registration
- https://neuroimage.usc.edu/brainstorm/Tutorials/LabelFreeSurfer#Run_FreeSurfer_manually
- https://neuroimage.usc.edu/brainstorm/Tutorials/LabelFreeSurfer#Run_FreeSurfer_from_Brainstorm
- https://neuroimage.usc.edu/brainstorm/Tutorials/Scouts#Menu:_Scout

Edouard2laire:
Ok. Looks good to me. 
Thanks for having written the tutorials. 

I just relaunched free surfer on the Template. Will send you the file tomorrow

Edouard2laire:
Hello Francois, 

I just realized something weird when reimporting the output of Freesurfer segmentation of ICBM152 template. The head mesh doesn't seem to correspond to the one imported by Brainstorm.  If we look at the MRI it seems that it is cut at the nose level; but the headmesh still goes down to the neck. So i was wondering how was the head mesh generated ? 

Capture d’écran 2022-08-10 à 00 01 12

(top : Freesurfer output and MRI / down: Brainstorm head mesh)

ftadel:
> So I was wondering how was the head mesh generated?

This is a old surface generated by Matti Hamalainen in MNE-C many years ago.
I use this one instead of the truncated one because otherwise most EEG caps have electrodes that are not located on the head surface, which causes all sorts of problems.

ftadel:
Thanks for this great contribution!

Edouard2laire:
> > So I was wondering how was the head mesh generated?
> 
> This is a old surface generated by Matti Hamalainen in MNE-C many years ago. I use this one instead of the truncated one because otherwise most EEG caps have electrodes that are not located on the head surface, which causes all sorts of problems.

Here is the FS output for ICBM 152: https://drive.google.com/file/d/10YyLnwbwFOgm61UWcERJt7e2WQM6pzy6/view?usp=sharing . Let me know if you would like me to compute the new sphere on other templates.  Do you know if the non-truncated MRI can be found somewhere ? it might be better to have it rather than the truncated one. 


Have a good day :)

ftadel:
> Here is the FS output for ICBM 152

Thanks.
I will add the spheres when I update the ICBM152 template (in a few months): https://neuroimage.usc.edu/brainstorm/Next#Anatomy

> Do you know if the non-truncated MRI can be found somewhere ?

No it doesn't exist.

Edouard2laire:
Hello Francois, 


I just made a small script for a collegue to be able to project to the controlateral hemisphere even if the original freesurfer processing didnt include the --contrasurfreg option : https://gist.github.com/Edouard2laire/513d198888642a67e90fdfe73030cb38 


The idea is to create a new subject and to import the recomputed freesurfer segmentation with contrasurfreg; then project the scout from the first subject to this new subject, do the controlateral projection before projecting back to the first subject. Seems to work quite well.

Edouard

ftadel:
> I just made a small script for a collegue to be able to project to the controlateral hemisphere even if the original freesurfer processing didnt include the --contrasurfreg option : https://gist.github.com/Edouard2laire/513d198888642a67e90fdfe73030cb38
> The idea is to create a new subject and to import the recomputed freesurfer segmentation with contrasurfreg; then project the scout from the first subject to this new subject, do the controlateral projection before projecting back to the first subject. Seems to work quite well.

This makes complete sense.
I have this objective of updating the ICBM152 template, including the contralateral registration.
Once this is done, all the new protocols created in Brainstorm would have at least one subject with which the left-right projection is possible, and therefore we could always offer your solution for all the subjects, even if they were not processed with the contrasurfreg option.

Let's link this to this other forum post:
https://neuroimage.usc.edu/forums/t/icbm-default-anatomy-with-coregistered-hemispheres-available/37088

ftadel:
@Edouard2laire 
I've been trying to run a new segmentation of the ICBM152 MRI with FreeSurfer 7.3.2, in order to generate the contralateral registration. But it seems like the option '-contrasurfreg' is not accepted anymore by recon-all, while it is still documented on the FreeSurfer website.
What version of FreeSurfer are you using?
Am I missing something obvious?

https://www.mail-archive.com/freesurfer@nmr.mgh.harvard.edu/msg73768.html

Edouard2laire:
Hello, 

I am using Freesurfer 6.0.0.  I didn't saw anything in the release not about removing the option: https://surfer.nmr.mgh.harvard.edu/fswiki/ReleaseNotes . 


Can you run recon-all --help and check if you still have : "
Contralateral Surface Registation (Spherical Morph) (-<no>contrasurfreg)

Same as ipsilateral but registers to the contralateral atlas.
Creates lh.rh.sphere.reg and rh.lh.sphere.reg."

Edouard2laire:
Hello Francois, 

I attached the script that is executed when running Freesurfer on my computer. Espetially the interesting lines are: 

```
#--------------------------------------------
#@# Contra Surf Reg lh Tue Aug  9 21:57:56 EDT 2022

 mris_register -curv -reverse ../surf/lh.sphere /util/packages/FreeSurfer/6.0.0/average/rh.folding.atlas.acfb40.noaparc.i12.2016-08-02.tif ../surf/lh.rh.sphere.reg 

#--------------------------------------------
#@# Contra Surf Reg rh Tue Aug  9 21:57:56 EDT 2022

 mris_register -curv -reverse ../surf/rh.sphere /util/packages/FreeSurfer/6.0.0/average/lh.folding.atlas.acfb40.noaparc.i12.2016-08-02.tif ../surf/rh.lh.sphere.reg 
```

Can you try to 
- execute this script as is; 
- execute 
- the following: 
```
recon-all -all -subjid ICBM152_2022
cd subject_directory

 mris_register -curv -reverse ../surf/lh.sphere /util/packages/FreeSurfer/6.0.0/average/rh.folding.atlas.acfb40.noaparc.i12.2016-08-02.tif ../surf/lh.rh.sphere.reg 
 mris_register -curv -reverse ../surf/rh.sphere /util/packages/FreeSurfer/6.0.0/average/lh.folding.atlas.acfb40.noaparc.i12.2016-08-02.tif ../surf/rh.lh.sphere.reg 
```


[recon-all.cmd.txt](https://github.com/brainstorm-tools/brainstorm3/files/9956760/recon-all.cmd.txt)

Edouard2laire:
I tried to anwser in the FS list but in case it didn't worked: 

Hello Francois, 

It seems that the option was removed in 2019 in this commit: https://github.com/freesurfer/freesurfer/commit/c827a2dd06ca636afb5cd0c62c49f638418915a7 . I am curious to know the reason for this removal.  

I guess the sphere can still be obtained manually using mris_register : 
mris_register -curv -reverse {$subject_dir}/surf/lh.sphere {$FreeSurfer_home}/average/rh.folding.atlas.acfb40.noaparc.i12.2016-08-02.tif {$subject_dir}/surf/lh.rh.sphere.reg 
mris_register -curv -reverse {$subject_dir}/surf/rh.sphere {$FreeSurfer_home}/average/lh.folding.atlas.acfb40.noaparc.i12.2016-08-02.tif {$subject_dir}/surf/rh.lh.sphere.reg


Edouard

ftadel:
This solution seems to be working well with FreeSurfer 7.X !
Thanks for looking into the scripts!

The discussion on the FreeSurfer mailing list doesn't seem to go anywhere.
The reason for the commit you pointed out is: "I think I took it out because it was a bad idea (can't really remember). "
https://www.mail-archive.com/freesurfer@nmr.mgh.harvard.edu/msg73814.html

I tried with Xhemi, but it doesn't work well.
https://www.mail-archive.com/freesurfer@nmr.mgh.harvard.edu/msg73824.html

I'll need some more time to finish a new ICBM152 template, as I realize there are a lot more issues to bring everything in the current ICBM152 space in Brainstorm (the MRI dimensions are not 256x256x256...) 
But I'll get there, eventually :)

Edouard2laire:
Hello :) 

> This solution seems to be working well with FreeSurfer 7.X ! Thanks for looking into the scripts!
> 
> The discussion on the FreeSurfer mailing list doesn't seem to go anywhere. The reason for the commit you pointed out is: "I think I took it out because it was a bad idea (can't really remember). " https://www.mail-archive.com/freesurfer@nmr.mgh.harvard.edu/msg73814.html

Yes; i followed the conversation (not sure still on how to contribute to the discussion in FS mailing list though). I would love to know the reason why he is saying it was a bad idea. 


> I tried with Xhemi, but it doesn't work well. https://www.mail-archive.com/freesurfer@nmr.mgh.harvard.edu/msg73824.html

Yes; i also tried Xhemi when i started to investigate this but was also not very convinced by the resutls. 


> I'll need some more time to finish a new ICBM152 template, as I realize there are a lot more issues to bring everything in the current ICBM152 space in Brainstorm (the MRI dimensions are not 256x256x256...) But I'll get there, eventually :)

Which MRI are you using? Also, I think one of the first steps of recon-all is to move the MRI to free-surfer space (resampling the MRI at 1x1x1 mm resolution and 256x256x256 matrix  

I recently came across a website called Template Flow that is used by fMRIprep for their template: https://www.templateflow.org and it seems there is multiple version of the ICBM template. Do you think we should propose them all in Brainstorm ? 

Edouard

ftadel:
> I would love to know the reason why he is saying it was a bad idea.

It would have been interested too, but I guess we'll never know.
This email list seems to bury any non-adressed topic very quickly.

> Which MRI are you using? 

I've been using the T1 ICBM152 2009c: https://www.bic.mni.mcgill.ca/ServicesAtlases/ICBM152NLin2009

> it seems there is actually multiple version of the ICBM template; so it might be good anyway to clarify which one we are using :)

The documentation is quite explicit:
https://neuroimage.usc.edu/brainstorm/Tutorials/DefaultAnatomy#Scenarios

> Do you think we should propose them all in Brainstorm ?

No. The differences between these models are much smaller than what we care about in non-invasive imaging.
One unique reference, and possibly conversion solutions if needed some day.

> I think one of the first steps of recon-all is to move the MRI to free-surfer space (resampling the MRI at 1x1x1 mm resolution and 256x256x256 matrix

Yes, but I'd like to keep the original MRI size for optimization reasons: lots of display and computation functions would work much faster if the volume has 50 less voxels in each direction. And the Brainstorm package would be smaller.
It's not a problem to keep the original volume and import the FS surfaces on top of it, we only need the fiducials to be set correctly. But since there will be volume parcellations now: I also need to cut these volumes produced by FreeSurfer...
I think I know what to do, but I'm not sure I'll have to work on it before next week.

Edouard2laire:
> The documentation is quite explicit: https://neuroimage.usc.edu/brainstorm/Tutorials/DefaultAnatomy#Scenarios

Indeed but then why use the asymmetric version and not the symmetric one? :) 

> > Do you think we should propose them all in Brainstorm ?
> 
> No. The differences between these models are much smaller than what we care about in non-invasive imaging. One unique reference, and possibly conversion solutions if needed some day.

Then maybe the extended version with the neck might be a good option :)


> Yes, but I'd like to keep the original MRI size for optimization reasons: lots of display and computation functions would work much faster if the volume has 50 less voxels in each direction. And the Brainstorm package would be smaller. It's not a problem to keep the original volume and import the FS surfaces on top of it, we only need the fiducials to be set correctly. But since there will be volume parcellations now: I also need to cut these volumes produced by FreeSurfer... I think I know what to do, but I'm not sure I'll have to work on it before next week.

Maybe https://surfer.nmr.mgh.harvard.edu/fswiki/FsAnat-to-NativeAnat can help. However, this is in contradiction with the rest of the software that seems to prefer to have the MRI in the FS space (ie when we import a folder from FreeSurfer).  Or we should have the option when importing / computing segmentation using FreeSurfer to import MRI / segmentation in FS space or in native space ?

ftadel:
> Indeed but then why use the asymmetric version and not the symmetric one? :)

Because most of the other programs using the ICBM152 2009c templates are using the asymmetric ones.

> Then maybe the extended version with the neck might be a good option :)

There is no ICBM152 2009c volume including a neck.
I don't remember how I created the head surface currently in the anatomy template in Brainstorm...

ftadel:
I'm finally moving forward...
There is now a new ICBM152 2022 template available, that includes the contralateral registration:
https://neuroimage.usc.edu/brainstorm/Tutorials/DefaultAnatomy#FreeSurfer_templates
https://github.com/brainstorm-tools/brainstorm3/commit/e9c9f0532c4b9ebe9241f581cf1a410533e808ac

It is not part of the Brainstorm main distribution yet, but it will be soon.
I'm planning to remove the default anatomy from the GitHub repository as well, it would be downloaded/updated if needed at the startup of Brainstorm (like the java package).

In the meantime, you can set a subject to use it manually: update Brainstorm, then right-click on a subject (or the default anatomy) > Use template > MNI > ICBM152 2022.

**DO NOT DO THIS ON A EXISTING PROTOCOL!**
Since the surface tesselations for the cortex are changing (new FreeSurfer version), this would break all the existing source maps using the template anatomy.

ftadel:
I wrapped up the projection between hemispheres through the anatomy template:
https://github.com/brainstorm-tools/brainstorm3/commit/0bf18d5b71f280f4776d4a08b2102c718e5bc8d4

The last thing I need to do is to update the ICBM152 template in the Brainstorm distribution. I'll work on this soon.

Edouard2laire:
Great. Thanks a lot !

PR #573, opened 2022-09-07
Set windows title

Edouard2laire:
Small change so the name of the window (Brainstorm) appears in the task bar.


Edouard2laire:
On Linux, without the line, it displays nothing. With the line, surprisingly it shows Brainstorm versionDate as on windows. I don't really know why however.

rcassani:
@Edouard2laire, my guess is that it's related to the DE, (sometimes) I have similar issues on Cinnamon.

PR #575, opened 2022-09-08
ephys: Use relative paths in spike (`data_0ephys`) files

rcassani:
Relative paths are saved in the `data_0ephys` files, instead of absolute paths.
Having absolute paths in these files breaks the protocol when exported to other PCs.


ftadel:
Could this be coded with `file_short` and `file_fullpath` instead that direct references to .STUDIES?
I had a vague objective of progressively removing these direct manipulations of the folders in ProtocolInfo.
(maybe that would require editing these functions to work with folders instead of files?)

If this comment does not make sense, please let me know and I'll review the PR again :)

rcassani:
> Could this be coded with `file_short` and `file_fullpath` instead that direct references to .STUDIES?
> I had a vague objective of progressively removing these direct references to the folders in ProtocolInfo.
> (maybe that would require editing these functions to work with folders instead of files?)

Yes, it can be coded that way, It will imply to add the type `folder_spikes` into `file_gettype()`.
I'll prepare it this way and let you know when ready.

rcassani:
@ftadel, ready for review 👀

ftadel:
Much easier to read through! Thanks

PR #577, opened 2022-09-10
Time series figure: Use DisplayUnits from the files (fix NIRS)

Edouard2laire:
This small PR is here to display delta OD a bit more nicely and fix a bug where: when displaying a time course the unit would be delta OD (10^-3) but becomes delta OD (10^-3)(10^-3)  if displaying a particular timecourse


ftadel:
There was no real need of removing the regex, but as you didn't like it, it's gone :)

Edouard2laire:
> There was no real need of removing the regex, but as you didn't like it, it's gone :)

thank you :). looks better indeed

PR #578, opened 2022-09-16
add oc_filtfilt

Edouard2laire:
Add oc_filtfilt to replace the use of filtfit when signal toolbox is not available https://octave.sourceforge.io/signal/function/filtfilt.html 

Cf: https://neuroimage.usc.edu/forums/t/fnirs-data-import/36740/12

merge https://github.com/Nirstorm/nirstorm/pull/205

Edit: this might be a usefull replacement here too: https://github.com/brainstorm-tools/brainstorm3/blob/6cc6addc8e084ce5209ed6e67f619953199981d8/toolbox/math/bst_bandpass_filtfilt.m#L165


ftadel:
Thanks!
Hopefully I didn't break anything else...

PR #579, opened 2022-09-17
cs_convert bugfix

Edouard2laire:
Define Porig after checking P exists and has the correct orientation


Edouard2laire:
This PR fix issue when importing from FreeSurfer; but i am not 100% sure if the backup of P to Porig should be done before or after transposing P (line 74 - ```    P = [P'; ones(1,size(P,1))];  ``` )

ftadel:
Ooooops.... 6 days with a broken super basic function.... 

Fixed here: https://github.com/brainstorm-tools/brainstorm3/commit/6aa4dce73c33e606f4666265d256517fa2bd9b96

Thanks for spotting, and insisting to put it again at the top of my todo list!

PR #580, opened 2022-09-19
Fibers issues

rcassani:
* Allow non-interactive call to `import_fibers()`
* Skip **fibers** and **fem** files in `bst_warp()`


ftadel:
It's actually easy to warp FEM meshes and fibers.

Please double-check that I didn't break anything with these two commits.

PR #581, opened 2022-09-19
5s Timeout in `urlread()` with Matlab > 2014a

rcassani:
Without this check `urlread` takes more than 120 s to return with error.
For older Matlab versions, there is not Timeout option


ftadel:
If the goal is only to add a timeout, let's do it this way:
https://github.com/brainstorm-tools/brainstorm3/commit/3bc610f46579839e5d506c1eeb5c38fef47e8e8b

The order of the two functions has some importance. They use different libraries, and are more or less robust to invalid certificates, versions of root certificates, etc. It's been a nightmare in the past, and now it's quite stable this way, so let's not play with it :)

PR #584, opened 2022-10-04
Bugfix: Get number of children after removing Intra (Bug reported in Forum)

rcassani:
Fix bug reported in: https://neuroimage.usc.edu/forums/t/36049
`nbChild` should be counted after removing `@intra`

Replicate bug:
1. Protocol with 1 Subject with only the `@intra` node
2. In condition view, root node will have only global `@default_study` and `@intra`


ftadel:
Thanks for debugging this!

PR #595, opened 2022-12-15
Allow to export matrix as EDF file

rcassani:
The idea would be to add an option to export a matrix (including its events) to an EDF file.
Any thought @ftadel?


ftadel:
I haven't tested it, but from overlooking the code, it looks good to me!
What are your doubts about it?

rcassani:
>  What are your doubts about it?
No doubts, just a second pair of eyes.

This PR is ready to merge!

ftadel:
When testing this on a few files, I found out that there were additional bugs in the export... sometimes the exponential notation of the physical_min and physical_max were truncated to 8 chars, right after the "e", making the fields useless.

I added some more fixes in this commit: https://github.com/brainstorm-tools/brainstorm3/commit/c2043436b871c63f6e9822c31afed47a4f87ff8c

In general, we should just discourage people to use this horrible format...

PR #596, opened 2023-01-06
Connect image: Checkbox menu item to zero diagonal values: Part 2

rcassani:
@ftadel, This PR improves the behaviour of the tool introduced in PR #593

The idea is to check that the connectivity matrix comes from a NxN process from Process 1 tab, or from an AxB process from Process 2 tab.

[Comments in previous PR](https://github.com/brainstorm-tools/brainstorm3/pull/593#issuecomment-1372531235): 

> [...] cases where this does not work as expected:
> * If the matrix is AxB non-square: most likely wrong
> * If the matrix is 1xN and the selected signal is not the first one

Option of zero diagonal values is Off and not shown for non-square connectivity matrices

> [...] cases where this does not work as expected:
> * If the matrix is square but with a different lists of regions in X and Y (you can obtain this with Process2: Connect AxB)

Addressed by comparing row names A and row names B. It was to allow the zeroing of diagonal values and to display matrix as graph.


rcassani:
@ftadel
There is still an issue with the solution in this PR:

Checking for same row names in A and B **is not enough** as an square matrix with same row names but different data (i.e., filesA != filesB) can be can be obtained with AxB with different files but using the same targets. To solve this we could:

1. In bst_connectivity, if FilesA and FilesB are same, and also TargetA and TargetB are the same, indicate that the process called was `connectivity_1n` instead of `connectivity_2`. Then zero diagonal values would be enable IFF the matrix was computed with a `connectivity_1n` process.

2. A bigger change. Update the filename prefixes in connectivity files to: 
  * `connectnn` (for P1 NxN results, OR P2 AxB with **same files** and same targets)
  * `connect1n` (for P1 1xN results, OR P2 AxB with **same files**, but targets A=1 and B=N)
  * `connectab` (for P2 AxB with **different files**, any targets)

  Then:
  * On/off zero auto-connectivity would enable for of `connectnn` files, and for `connect1n` files if the targetA is in targetsB.
  * Graph display would be enable only for `connectnn` files
  * Zero auto-connectivity would be off for `connectab`, and not graph display regardless the result may be a square matrix

Some thoughts on this?

ftadel:
Or... we simply consider that would be a very unusual case, and simply ignore it? :)

rcassani:
> Or... we simply consider that it would be a very unusual case, and simply ignore this problem? :)

+1 We can also do this. 

With such approach, we can detect auto-connectivity by the name of signals. Then, the improved behaviour is: show/hide all the auto-connect values in images from NxN and 1xN even if they are not in the diagonal. 

Example:
wi_val

wo_val

This PR is ready to be merged ;)

ftadel:
What about "self-connectivity" instead of "auto-connectivity"?
I guess this is by analogy with "auto-correlation", but "auto-connectivity" sounds weird, doesn't it?

And maybe the label of the popup menu could be 'Hide self-connectivity' instead of 'Zero auto-connectivity values'?

@sbaillet may have a better feeling for what is the most appropriate wording.

rcassani:
@ftadel, the discussed changes are done. 
Ready to merge 😉

ftadel:
I updated the part of the tutorial where this feature is introduced: 
https://neuroimage.usc.edu/brainstorm/Tutorials/Connectivity#Result_visualization

Other images in this tutorial and the connectivity graph tutorials should be updated as well:
- https://neuroimage.usc.edu/brainstorm/Tutorials/Connectivity
- https://neuroimage.usc.edu/brainstorm/Tutorials/ConnectivityGraph#Loading_connectivity_graphs

rcassani:
I'll handle those 😉

PR #597, opened 2023-01-09
Add Modality validation

rcassani:
Add validation for Modality in `view_mri`.
The replaced code on Modality was introduced in 2d451cbd74f7efa13f5b542e10167de6ed24050a, and produced this warning in the Command Window:

image


ftadel:
Thanks for finishing what I surprisingly stopped half-way :)

PR #599, opened 2023-01-31
Matrix options

rcassani:
* Context menu to review matrix node as raw
* Matrix node can be exported as Brainstorm binary file
* ~~Remove exporting Matrix node as EDF+ file (as it's not recommended.~~ 
  ~~(It can be still done with two steps: 1. Review matrix as raw, 2. Export raw file as as EDF+)~~
* Update Event processes not related to artifact detection nor requiring ChannelFile to work with Matrix inputs


ftadel (approved):
Looks good to me!

Sorry for guiding you towards removing the EDF+ export. 
Even though not recommended in most cases, it might still have meaningful use cases. And since it is already in there, let's keep it. 
=> I added it back

rcassani:
@ftadel, it looks good. 
Can you merge it?

PR #600, opened 2023-02-13
Process_evt_timeoffset : Add suffix for creating new events and load relative/variable timeoffset

annesodub:
Modify process_evt_timeoffset to allow the use to load variable event file for time offset

This modification includes :

1. with previous "fixed" time offset : the possibility to create a new event name with a suffix (so that the user can keep previous event without time offset) 
3. with a new input option (Event file) : the possibility to load an event file to apply a variable time offset relative to existing event(s) (e.g. reaction time)


ftadel:
The only file included in this PR is a `.DS_Store`...

annesodub:
ooops... sorry, I am closing this PR and pushing again (with the modified file :-) )

PR #601, opened 2023-02-14
Process_evt_timeoffset : Add suffix for creating new events and load relative/variable timeoffset

annesodub:
Modify process_evt_timeoffset to allow the use to load variable event file for time offset

This modification includes :

- with previous "fixed" time offset : the possibility to create a new event name with a suffix (so that the user can keep previous event without time offset)
- with a new input option (Event file) : the possibility to load an event file to apply a variable time offset relative to existing event(s) (e.g. reaction time)


rcassani:
> * with previous "fixed" time offset : the possibility to create a new event name with a suffix (so that the user can keep previous event without time offset)

I think, it may be better to add the suffix to the original event, resulting in something like `eventName-original`, and keep the `eventName` for the shifted.

ftadel:
> I think, it may be better to add the suffix to the original event, resulting in something like `eventName-original`, and keep the `eventName` for the shifted.

It feels more logical to me that a process that creates a new event uses a new name (1 modification), rather than doing 2 modifications simultaneously: renaming the older + creating a new event with the previous name (or: creating a backup + modifying the existing).
Or am I missing a more obvious understanding for this feature?

Anyway, in most cases this field would remain empty and the process would exactly as it was before.

rcassani:
I suggested that to have the same behaviour of "Add a time offset to all the events of a given category" regardless of the suffix field. Indeed, it can be done with a checkbox (as in the merge process):
`[ ] Make a copy of original event before adding time offset `

Either the suffix text field or checkbox are fine.

ftadel:
I fixed a few things:
- Restricted input files to arrays of values and read it without all the complexity of `import_events` 
- Modified the labels and organization of the panel
- Simplified the code
- Made it EITHER fixed OR variable (not both)

Does this work for both of you?

image

rcassani (approved):

rcassani:
Looks good to me ;)

ftadel:
@annesodub The PR is ready for your review.

annesodub:
Hi! Thank you both for improving the design and code. I just tested the creation of an event with relative time offset and experienced (what I believe to be) a bug. Here are the steps I followed :

From the dataset used at PracticalMEEG ([Wakeman and Hanson](https://pubmed.ncbi.nlm.nih.gov/25977808/))

1. I select the "Famous" events  
2. Menu File -> Export Selected Event -> File format Array of Time (.txt) -> saved the file xxx.txt
3. Then Menu Events -> Add time offset : 
Event Name : Famous 
Fixed time offset : 0 ms
Relative time offset : xxx.txt 
Backup suffix : orig

So here the idea (just for the test) is to obtain a new Famous event which would results in being just Famous + Famous. Howerver,I do not end up with this, but rather with an extended event (see snapshot below).

image

Do you know what have gone wrong?

ftadel:
Fixed. Please try again.

ftadel:
@annesodub  Do you want to test this before we merge it?

annesodub:
Yes! Sorry about the late reply. I tested it with no sucess for the latest modification (the event is still extended..). I can provide more details tomorrow.

ftadel:
> I tested it with no sucess for the latest modification (the event is still extended..). I can provide more details tomorrow.

I re-tested again your example above, and it does not produce anymore extended events on my computer.
Please check that you tested the last version I committed to this branch:
- run `edit process_evt_timeoffset.m` (to open the plugin function that is actually used by Brainstorm)
- check the authors information, it should show `Authors: Francois Tadel, 2013-2023`   (instead of 2015)

annesodub:
Hi, thank you Francois. You are right, it works fine (no more extended events). 

I just have one more concern: when using the field "Relative time offset (file)", if the user enter a value into the field  "Fixed time offset", it does not have any impact (if I use 0 or 50ms, it gives the same results). 

I think we should either not allow the user to use the field "Fixed time offset" when using the field "Relative time offset (file)" or, alternatively, (my preference) apply the "Fixed time offset" to the resulting event (in addition to the relative time offset). Do you see what I mean? 

Best,

AnneSo

ftadel:
It is not possible to disable the "Fixed time offset" option when there is a file that gets selected.
One option would be to add an extra radio button with the choice "fixed"/"relative", and each one enables one of the lines below. But it makes the process incompatible with the previous versions.
Technically, it would be easier to simply add more information to make it more explicit that the two options are mutually exclusive.

Is it better with this extra line of comment?

image

annesodub:
Why should we not allow the combination of both? (fixed + relative) 

If we allow this could we send a warning when both are used AND that Fixed time offset is different from 0 ms (just to make sure the user is aware that he/she is using both)

Otherwise, yes the comment is clear, I am just afraid some users won't read it... grrr

ftadel:
> Why should we not allow the combination of both? (fixed + relative)

I'm not sure I understand what the use case is for these combined options.
It corresponds to a weird mix of: 1) compensating for hardware delays and 2) creating response events from (stim latency + response time).
One process call for doing both doesn't seem readable to me, if you think in terms of a processing script that an uninformed person reads. If both are needed, then having two separate calls could be clearer.

But this is not going to be used much anyway, so whatever works...

annesodub:
Yes, you are right, the combination does not make much sense on the ground.
This looks good to me now. Could you please merge? 

For the release notes, I have a suggestion : adding the keywords "e.g. reaction times" would make it more explicit to the researchers in the field of cognitive neuroscience and psychology (on what they could use this process for, rather than just "relative time offset", which may sound more obscure to them) 

Thank you very much!

ftadel:
I added a longer description, and renamed "relative" into "variable".
(fixed/variable or absolute/relative)

image

Hopefully this is now clear enough.

PR #603, opened 2023-02-24
Feature: sort ICs (from ICA) by variance is not reference channel

rcassani:
Sort ICs (from ICA) by variance is not reference channel.
Previous behaviour: If reference channel string is empty, ICs are not sorted.

IC selection in tutorials: tutorial_epilepsy and tutorial_yokogawa is not affected

### DIFF
--- a/toolbox/process/functions/process_ssp2.m
+++ b/toolbox/process/functions/process_ssp2.m
@@ -772,7 +772,11 @@ function OutputFiles = Run(sProcess, sInputsA, sInputsB)
             [corrs, iSort] = sort(max(abs(C),[],1), 'descend');
         else
             % By explained variance
-            M = inv(W);
+            if diff(size(W)) == 0
+                M = inv(W);
+            else
+                M = pinv(W);
+            end
             var = sum(M.^2, 1) .* sum(Y.^2, 2)';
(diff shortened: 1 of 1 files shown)


rcassani:
>We should handle runica separately in order to avoid doing it twice and avoid changing the sorting performed by runica.

Sure, no need for double sorting. Although, it shouldn't change as it's the same criterion.

> Did you do this initially based on tests using Picard?

Yes, I realized the ICs from Picard were not sorted (with empty `icaSort` and `nIcaComp` == 0).

Looking into the runica code, there is also the need to code the case for the case when the number of IC is smaller than the number of channels.

ftadel:
> Sure, no need for double sorting. Although, it shouldn't change as it's the same criterion.
> Looking into the runica code, there is also the need to code the case for the case when the number of IC is smaller than the number of channels.

@rcassani Are you planning to work on these two aspects?

rcassani:
> Are you planning to work on these two aspects?

Yes, I'll let you know when it's ready

rcassani:
@ftadel, 
1. The case for nICs < nChannels is addressed.
2. IC needed to be sorted by the variance they explain only for Picard and FastICA methods.
I've tested the code, and this PR seems ready to be merged.

PR #604, opened 2023-03-09
Allow contralateral projection of volume scouts

rcassani:
Allow contralateral projection of volume scouts
- Projection is done in the MNI space (changing x-axis sign)
- For polyhedron scouts (>= 4 vertices), the boundary is projected, then this is used to find the projected scout
- For "open" scouts (< 4), the closest grid points to the projected vertices are used for the projected scout

Examples: In the two images below, all the scouts in the right hemisphere are projections from scouts in the left hemisphere

Screenshot from 2023-03-09 14-31-36

image

### DIFF
--- a/toolbox/gui/panel_scout.m
+++ b/toolbox/gui/panel_scout.m
@@ -1155,8 +1155,9 @@ function iScouts = SetScouts(SurfaceFile, iScouts, sScouts)
     end
     % Save the previous scouts configuration
     sScoutsOld = GlobalData.Surface(iSurf).Atlas(sSurf.iAtlas).Scouts;
-    % Detect region if not defined yet (only for new scouts)
-    if isAdd
+    isVolumeAtlas = ParseVolumeAtlas(GlobalData.Surface(iSurf).Atlas(sSurf.iAtlas).Name);
+    % Detect region if not defined yet (only for new surface scouts)
+    if isAdd && ~isVolumeAtlas
         for i = 1:length(sScouts)
             if ~isempty(sScouts(i).Seed) && (isempty(sScouts(i).Region) || strcmpi(sScouts(i).Region, 'UU'))
                 sScouts(i) = SetRegionAuto(sSurf, sScouts(i));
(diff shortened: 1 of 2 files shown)


ftadel:
I tested this procedure in various ways, and I am concerned with the outputs that we get from this solution, even with non-linear MNI registration.
Example: I took the MNI atlas AAL3, kept only the "precentral" scouts, projected the left one, and observed the right ones (original in purple + projected in red). They don't overlap much. As such, I don't think we want to recommend the use of this solution.

Is the problem the growth of the scout after calling inpolyhedron? It's making it more round instead of growing in all the directions, closer to the original shape.
Maybe the boundaries should be inflated in all the directions before calling inpolyhedron.

image

Moo-Marc:
Although this may be a real issue to fix, perhaps the better test is to project left, then back right (or vice-versa) and compare to the original scout.  I don't think we should expect the projection to find the "real" contralateral region if their shapes differ.  

I believe this is a request related to epilepsy delineated zones, to use the other hemisphere as a control.  I guess we could use an atlas to try to only include vertices within the same subset of regions after projection, but they probably also want an equivalent volume, in which case it's not straightforward.

rcassani:
@ftadel 

> Is the problem the growth of the scout after calling inpolyhedron? It's making it more round instead of growing in all the directions, closer to the original shape. Maybe the boundaries should be inflated in all the directions before calling inpolyhedron.

In the last set of updates, the boundary is not computed as the convex hull (`shrink factor = 0`), but as the compact surface (`shrink factor = 1`), to conserve the original scout geometry. Also the boundary is computed after projecting the original scout points to the GridLoc in the other hemisphere.

@Moo-Marc 

> Although this may be a real issue to fix, perhaps the better test is to project left, then back right (or vice-versa) and compare to the original scout. I don't think we should expect the projection to find the "real" contralateral region if their shapes differ.

Agree, indeed these were my ground-of-truth tests

--- 
In addition, now `inpolyhd()` is used, as it is not sensitive to the orientation of the triangle normals (as `inpolyhedron()` is). 

The growing algorithm, now grows in all the scout directions. However, if shrinking is needed it will trim from the boundary vertices, starting the farther points to the scout center.

Results: 
* Original scout (red): 183 vertices, 18.56 cm3
* Red scout projected to right hemisphere (right green): 183 vertices, 19.67 cm3
* Right green scout projected to left hemisphere (left green): 183 vertices, 19.29 cm3

image

ftadel:
It works better, thanks.

Now does it make sense physiologically? 
The surface version is based on a registration that aligns similar structures left-right. There is no such thing done in volume, and the MNI template used for the MNI registration in SPM is NOT symmetrical left-right.
Therefore the method is by construction inaccurate, and should not be used for precise anatomical processing.

However, when used in the context of volume-based source estimation, the spatial resolution is already so low that it probably doesn't matter much to be off by 1cm in the projection of a scout.

PR #608, opened 2023-04-01
improve display of plugin list

Edouard2laire:
Hello, 

This is a very small change on the display of the plugin list. 

Before: 
Capture d’écran 2023-03-31 à 21 13 54

After:
Capture d’écran 2023-03-31 à 22 24 42


ftadel:
The problem is related with MacOS not being capable of using a standard monospace font correctly.
On Windows and Linux, it looks like this:
image

I would prefer to keep this as plain text, as it allows for quick copy-paste of the entire thing at once, which a table would not offer.

Could you try to find a font replacement for "Courier New" that is available on all the recent MacOS installs?
And add a switch in view_text, line 72, to use it when on a mac? 
`if strncmp(computer,'MAC',3)`

Please also keep the lower case for the plugin names, as it is the way they are used in the code and interface.

Thanks

Edouard2laire:
Ok. That make sense. I will give a try this weekend but it doesn't seem to be an easy topic :)

ftadel:
If you can't find any solution, there is also the option to use the table only for MacOS, as it would be better than the currently unreadable thing that you showed :)

In that case, could you please embed back the ViewTable function into "view_test", in order to limit the number of functions?

rcassani:
@ftadel 
> The problem is related with MacOS not being capable of using a standard monospace font correctly. On Windows and Linux. 

In Linux, it is not working either :'(

I believe the issue is that `Courier New` font (which I have in a MacMini with Ventura) does not get cached here: https://github.com/brainstorm-tools/brainstorm3/blob/abcd279aed925e8c9b4c627dfeadecb1c12e2525/toolbox/core/bst_get.m#L3909

Replaced: 
```
end
    fontTypes{end + 1} = 'Arial';  % Default font
    fontTypes{end + 1} = 'Liberation Sans';  % Free Arial substitute
```
with
```
else
    fontTypes{end + 1} = 'Arial';  % Default font
    fontTypes{end + 1} = 'Liberation Sans';  % Free Arial substitute
end
```

This seems to do the trick. 

There was a change around these in 9b24c4a, at fixing font for "up" and "down" buttons (not sure which buttons the commit refers to)

* To avoid dealing with the `Courier New` font, we could use `Monospaced` which will point to the system's monospace font.

Edouard2laire:
Indeed, adding changing the lines to "
        fontTypes{end + 1} = 'Courier New';
        fontTypes{end + 1} = 'Arial';  % Default font
        fontTypes{end + 1} = 'Liberation Sans';  % Free Arial substitute" 

do the trick : 
Capture d’écran 2023-06-21 à 09 21 27

But is changing the font of the entire software. is there a way to change the font only for the table display ?

rcassani:
Addressed at [2163f7e](https://github.com/brainstorm-tools/brainstorm3/commit/2163f7ed20461aebc10a4d267aecc1e26570fc79)

PR #610, opened 2023-04-03
Bugfix cnx display units

rcassani:
Handle `DisplayUnits` for connectivity TimeFreq files
Fixes issue reported in: https://neuroimage.usc.edu/forums/t/39543

### DIFF
--- a/toolbox/connectivity/bst_connectivity.m
+++ b/toolbox/connectivity/bst_connectivity.m
@@ -346,6 +346,7 @@ for iFile = 1:length(FilesA)
     switch (OPTIONS.Method)
         % === CORRELATION ===
         case 'corr'
+            DisplayUnits = 'Correlation';
             bst_progress('text', sprintf('Calculating: Correlation [%dx%d]...', size(sInputA.Data,1), size(sInputB.Data,1)));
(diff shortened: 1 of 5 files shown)


ftadel:
> We could use [yyaxis](https://www.mathworks.com/help/matlab/ref/yyaxis.html) to put the label on the left side of the colorbar, but this function is for Matlab16b or newer to have something like below. What do you think?

The vertical option is also cut, so this is not a correct solution.
Maybe some units names could be shortened specifically when being displayed as a colorbar legends?
"Magnitude-squared coherence" => "Coh"

ftadel:
Other than that, it looks great!
Thanks

rcassani:
> Maybe some units names could be shortened specifically when being displayed as a colorbar legends?

Sound good! I'll use their short names and be sure they are in the GUI and connectivity page

rcassani:
Since connectivity images and graphs are two representations of the same information, it seems logic to use the same colorbar (position and size). Shortening some unit labels may introduce some confusion; now if the string is longer than the available space, it is vertically shown (without `yyaxis` thus no back-compatibility issue). This is the result:
image

PR #611, opened 2023-04-06
Bugfix tPAC

rcassani:
Bugfix for tPAC process (by @sbaillet)
- Use the 2019 version of the filters
- Fix the definition of the bandpass which was causing problems, as also identified by @Moo-Marc 
- Add a refinement procedure to optimize the estimation of fP (frequency for phase), as also suggested by Marc 
- Minor fixes
- Add a new `Options.fAResolution` parameter, for users to specify the frequency resolution for exploring possible values of A (frequency for amplitude) within the specified range.


toolbox/process/functions/process_pac_dynamic.m
```diff
@@ -760,49 +771,56 @@
...
+                Xnesting = bst_bandpass_hfilter(X, sRate,bandNesting(1), bandNesting(2), isMirror, isRelax, [], fProlloff, Method);    % Filtering
+                Xnesting = Xnesting(:, nMargin-nHilMar+1:fix((margin+winLen)*sRate)+nHilMar);              % Removing part of the margin
+                % Hilbert transform
+                Z = hilbert_fcn(Xnesting')';
+                % Phase detection
+                nestingPh = angle(Z-repmat(mean(Z,2),1,size(Z,2)));    % Phase of nesting frequency
+                nestingPh = nestingPh(:,nHilMar:fix(winLen*sRate)+nHilMar-1);              % Removing the margin
+
+                %%% Refine PAC estimation
+                refine_bandNested = [nestedCenters(ifreq)-1.1*bandNesting(2),nestedCenters(ifreq)+1.1*bandNesting(2)];
```
Moo-Marc, line 795: Doing it in 2 steps like this were only the selected fp is recomputed is efficient and would remove a potential bias in the pac values at the selected frequencies, but I'm still worried there may be some bias in the first step of selecting the fa frequency of the pac peak.

toolbox/process/functions/process_pac_dynamic.m
```diff
@@ -637,17 +643,18 @@
     else
         nestedCenters = linspace(faBand(1),faBand(end),nHighFreqs);
     end
-    Fstep = diff(nestedCenters)/2;  % the range of frequency around each nested center
+    Fstep = diff(nestedCenters);  % the range of frequency around each nested center
     Fstep = [Fstep(1),Fstep,Fstep(end)];
-    Fstep = max(Fstep, fpBand(2)/2);  % Minimum band width is defined to cover the whole interval between consecutive centre frequencies and at the same time consider all coupled frequencies to it in the range of interest.        
-    fArolloff = [];
+    Fstep = max(Fstep, fpBand(2));  % Minimum band width is defined to cover the whole interval between consecutive centre frequencies
```
Moo-Marc, line 648: This is the bandwidth needed for our example, but are we convinced it's sufficient in general?  I presume here the amplitude simply varies sinusoidally at fp, but maybe we can construct more complex examples of coupling "shapes" and look at the spectra.
Also, do we know that it won't artificially increase pac at max(fp) because we cut off the possible amplitude variations exactly at that frequency?  I'd recommend running tests on a few different types of noise, and examples with low noise, and looking at the stats.  We should ensure there's no bias in fp due to this.

toolbox/process/functions/process_pac_dynamic.m
Moo-Marc, line 1: The test at line 577 should also be fixed: fpBand(2) > faBand(1)/2
And I would recommend changing the fa(1) instead of fp(2) as the latter is not immediately obvious in the results.  In our example, it would cause fp to stop at 5 Hz and miss the 9 Hz coupling.  I think there was also a bug where this change of bands is not saved in the output file: it keeps the inputs, which is misleading.

My other comments below are more thoughts about how we should test and validate this.

rcassani: Check fixed, and it stops execution showing error
Commit [6b49027](https://github.com/brainstorm-tools/brainstorm3/pull/611/commits/6b490278379f10e1de9680e1c49aa32f8650aeaa)

PR #612, opened 2023-04-12
Add 'MSC' option to `henv` envelope correlation processes

rcassani:
- Add magnitude-squared coherence option to `henv` processes
- Add equation description for coherence options in `henv` processes


Moo-Marc (approved):
This is mainly to clarify that the option "Coherence" in this envelope correlation process was **not** squared, vs MSC in the actual coherence process.  But we will soon discuss removing these options from this process, and rather keep all desired coherence options in the coherence processes only.

PR #613, opened 2023-04-13
Nirs projection

Edouard2laire:
Hello, 

This made the projection of sensors also works for NIRS. Also, fix a bug where there would be no warning if the MNI transformation was not computed for the subjects. 

Edouard

### DIFF
--- a/toolbox/math/bst_project_channel.m
+++ b/toolbox/math/bst_project_channel.m
@@ -121,7 +121,7 @@ ChannelMatDest.Channel  = ChannelMatSrc.Channel;
 for i = 1:length(ChannelMatSrc.Channel)
     if ~isempty(ChannelMatSrc.Channel(i).Loc)
         if size(ChannelMatSrc.Channel(i).Loc,2) == 2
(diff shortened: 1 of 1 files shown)


toolbox/math/bst_project_channel.m
```diff
@@ -87,16 +88,16 @@
 
 % Load source MRI
 sMriSrc = in_mri_bst(sSubjectSrc.Anatomy(sSubjectSrc.iAnatomy).FileName);
-if cs_convert(sMriSrc, 'scs', 'mni', [0, 0, 0])
+if isempty(cs_convert(sMriSrc, 'scs', 'mni', [0, 0, 0]))
     errMsg = ['Compute MNI normalization for subject "' sSubjectSrc.Name '" first.'];
```
Edouard2laire, line 92: Here, we might want to do something like errMsg = [errMsg , '\n .... new message' ] to concatenate with previous error message ?

rcassani: The script should terminate if any of the previous conditions happen. Addressed in: 609ea50

rcassani:
@Edouard2laire, Thanks for the adding the functionality for NIRS sensors and catching the strange behaviour in error handling 😉

PR #615, opened 2023-04-21
Bugfix BrainSuite MacOs

Edouard2laire:
Hello, 

It seems we need to call bdp.sh on Mac, instead of bdp. For windows, according to the BrainSuite website, we might need to use bdp.exe but I am not sure. 

Edouard


rcassani:
Nice catch, thanks @Edouard2laire!

After checking it in Linux and Windows:
- Linux also requires the `.sh`
- Windows it's the same with or without the `.exe` extension

PR #617, opened 2023-05-01
pac and tpac averaging bug fixes

Moo-Marc:
Fixes several PAC averaging bugs, some recently reported on the forums: [1](https://neuroimage.usc.edu/forums/t/error-in-averaging-pac-comodulogram-from-different-subjects/39997), [2](https://neuroimage.usc.edu/forums/t/problem-with-subject-averaging-in-time-resolved-pac/39967)

Also: hide the tPAC option for 2-seconds buffer (as discussed with @sbaillet), fix GUI typo for seconds, and added checks and error messages for consistency between files when averaging (frequency lists and array sizes).

While this fixes bugs in the averaging process, this is not the recommended way to aggregate tPAC across trials since the saved values are already "peak" values in fP.  It remains to discuss if the "average" option of the tPAC process should be improved.  It currently concatenates the "full" arrays (DynamicPAC field) - currently inconsistent with visualization (only displays first trial), while averaging the max values (TF field).

### DIFF
--- a/toolbox/connectivity/bst_pac.m
+++ b/toolbox/connectivity/bst_pac.m
@@ -1,7 +1,7 @@
 function [DirectPAC, LowFreqs, HighFreqs] = bst_pac(F, sRate, bandNesting, bandNested, isUseParallel, isUseMex, numfreqs)
 % BST_PAC: Calculate the DirectPAC metric for all the input time series.
 %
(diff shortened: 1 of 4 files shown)


Moo-Marc:
Also fixed recently introduced bug for multi-source files, where the peak fP of the first source was kept for others.  Example with 4 sources with fP=6,7,8,9 respectively.  Before: 
TF_Test_PAC_multi_tPAC_4_signals_02
Now fixed:
TF_Test_PAC_multi_tPAC_4_signals_fixed

PR #622, opened 2023-05-18
Export Nirs channels coordinates

Edouard2laire:
Hello, 
This PR extends the channel export functions to NIRS.

This PR also changes out_channel_nirs_brainsight.m to make it more similar to the other output functions :)

Edouard


Edouard2laire:
@ftadel, in BIDS import function, one of the file is *_channels.tsv. Is there a place in brainstorm to export that file ? i haven't found it.

rcassani:
> @ftadel, in BIDS import function, one of the file is *_channels.tsv. Is there a place in brainstorm to export that file ? i haven't found it. Here export channel is only exporting the electode file, i guess.

That's correct. Brainstorm does not create `*_channels.tsv` files, just the `*_electrodes.tsv`

rcassani:
@Edouard2laire, the changes in this PR look good

For `out_channel_bids.m`, I think it would be better to add an extra input argument, `isNIRS`:
* If `isNIRS` is not set, the function behaves as today. 
* If `isNIRS` is set, the function exports only NIRS sensors in the [`*_optodes.tsv` format](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/11-near-infrared-spectroscopy.html#optode-description-_optodestsv).
Lastly, the NIRS option would have its file format in the GUI: **'NIRS: BIDS (*_optodes)'** which simply calls `out_channel_bids.m` with the isNIRS flag.

What do you think about this approach for exporting NIRS channels in BIDS?

Edouard2laire:
> > @ftadel, in BIDS import function, one of the file is *_channels.tsv. Is there a place in brainstorm to export that file ? i haven't found it. Here export channel is only exporting the electode file, i guess.
> 
> That's correct. Brainstorm does not create `*_channels.tsv` files, just the `*_electrodes.tsv`

To be able to export it; should I create a new export format, or can I add an option to out_channel_bids asking if we should also export *_channels.tsv?

Edouard2laire:
> @Edouard2laire, the changes in this PR look good
> 
> For `out_channel_bids.m`, I think it would be better to add an extra input argument, `isNIRS`:
> 
> * If `isNIRS` is not set, the function behaves as today.
> * If `isNIRS` is set, the function exports only NIRS sensors in the [`*_optodes.tsv` format](https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/11-near-infrared-spectroscopy.html#optode-description-_optodestsv).
>   Lastly, the NIRS option would have its file format in the GUI: **'NIRS: BIDS (*_optodes)'** which simply calls `out_channel_bids.m` with the isNIRS flag.
> 
> What do you think about this approach for exporting NIRS channels in BIDS?

something like that ? or did I misunderstood?

Edouard2laire:
oh ok. That makes sense.  However, I am afraid it might be confusing that all the variants of EEG: ASCII: XYZ (*.txt), using out_channel_ascii can be used for EEG and NIRS but the bids are exclusive for only NIRS or EEG. 

With the actual version, it seems that we cannot export to ASCII with nirs. So I guess we should also duplicate the file format for NIRS.  But it might make the menu very busy. Maybe we could only display part of the menu depending on the file that is being exported: If EEG/MEG, we don't show the option about NIRS and vise-versa?

rcassani:
> With the actual version, it seems that we cannot export to ASCII with nirs. So I guess we should also duplicate the file format for NIRS. But it might make the menu very busy.

Yeah, that would be misleading. And will make so long the menu

> Maybe we could only display part of the menu depending on the file that is being exported: If EEG/MEG, we don't show the option about NIRS and vise-versa?

Aren't recordings that can have NIRS and EEG sensors? 

Since it's the text in the `File of Type`, another way could be to edit the text to be `EEG/NIRS: ...`, or using other text to indicate that it would work with any modality

Edouard2laire:
Thanks for the merge :)

PR #623, opened 2023-05-24
Label and export NIRS channels

Edouard2laire:
Hello, 

This PR is adapting the anatomical labeling process of SEEG (https://neuroimage.usc.edu/brainstorm/Tutorials/ECoG#Anatomical_labelling) for fNIRS. 

The main difference is that the contact point of the SEEG electrode is replaced by the point on the cortex to which the nirs channel is the most sensitive to. 

The code also exports the channel length and that sensitivity. The rest of the code is the same. 

Here is an example output: 

Capture d’écran 2023-05-23 à 22 40 33

I just want to confirm that sCortex.Vertices(iVertices,:) are in SCS coordinate ? 

Edouard

### DIFF
--- a/toolbox/io/export_channel_nirs_atlas.m
+++ b/toolbox/io/export_channel_nirs_atlas.m
@@ -1,0 +1,517 @@
+function TsvFile = export_channel_nirs_atlas(ChannelFile, TsvFile, Radius, isProba, isInteractive)
+% EXPORT_CHANNEL_NIRS_ATLAS: Compute anatomical labels for NIRS channels from volume and surface parcellations
+%
+% USAGE:  TsvFile = export_channel_atlas(ChannelFile,  TsvFile=[ask], Radius=[ask], isProba=[ask], isInteractive=1)
+%
+% INPUT: 
+%     - ChannelFile   : Path to Brainstorm channel file to be processed
+%     - TsvFile       : Output text file (tab-separated values)
+%     - Radius        : Size in millimeters of the neighborhood to consider around each contact
+%     - IsInteractive : If 1, display the output table at the end of the process
+%     - iChannels     : Limit export to a subset of channel indices
+% 
+% @=============================================================================
+% This function is part of the Brainstorm software:
+% https://neuroimage.usc.edu/brainstorm
+% 
+% Copyright (c) University of Southern California & McGill University
+% This software is distributed under the terms of the GNU General Public License
+% as published by the Free Software Foundation. Further details on the GPLv3
+% license can be found at http://www.gnu.org/copyleft/gpl.html.
+% 
+% FOR RESEARCH PURPOSES ONLY. THE SOFTWARE IS PROVIDED "AS IS," AND THE
+% UNIVERSITY OF SOUTHERN CALIFORNIA AND ITS COLLABORATORS DO NOT MAKE ANY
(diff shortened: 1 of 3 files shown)


toolbox/io/export_channel_nirs_atlas.m
```diff
@@ -0,0 +1,536 @@
...
+% Select high-resolution white surface if available, otherwise any high-resolution surface
+if ~isempty(iCortex)
+    if any(isWhite)
+        isSelect = [isSelect, isWhite & (nVertices == max(nVertices))];
+    else
+        isSelect = [isSelect, nVertices == max(nVertices)];
+    end
+end
+% Checkboxes
+isSelect = java_dialog('checkbox', 'Select information to export:', 'Compute contact labels', [], Columns(:,2), isSelect);
```
rcassani, line 189: Update window title

rcassani (changes requested):

rcassani:
@Edouard2laire, are there recommendations regarding the Radius to be used for NIRS channel labelling?

Edouard2laire:
ok. i applied your comments. Let me know if that looks ok for you :)




> @Edouard2laire, are there recommendations regarding the Radius to be used for NIRS channel labelling?

Unfortunately, up to my knowledge there is no recommendation. but I will try to check a bit more. From my tests, 3mm seems to make the more sense

toolbox/io/export_channel_nirs_atlas.m
```diff
@@ -0,0 +1,538 @@
...
+% LIABILITY OR RESPONSIBILITY FOR THE USE OF THIS SOFTWARE.
+%
+% For more information type "brainstorm license" at command prompt.
+% =============================================================================@
+%
+% Authors: Francois Tadel, 2021; Edouard Delaire, 2023
+
+
+% ===== PASRSE INPUTS =====
+if (nargin < 5) || ismpety(isInteractive)
```
Edouard2laire, line 35: isempty instead of ismpety

toolbox/io/export_channel_nirs_atlas.m
```diff
@@ -0,0 +1,557 @@
...
+% Get channel indices
+iChannelMod = channel_find(ChannelMat.Channel, 'NIRS');
+
+if isempty(iChannelMod)
+    error('No available NIRS channels.');
+end
+
+if isempty(sStudy.HeadModel) || isempty(sStudy.iHeadModel)
+    error('No available head model.');
+end
```
rcassani, line 104: This check should be just after getting the study, to avoid asking the user for a file that will not be created

toolbox/io/export_channel_nirs_atlas.m
```diff
@@ -0,0 +1,557 @@
...
+    else
+        isSelect = [isSelect, nVertices == max(nVertices)];
+    end
+end
+
+% Checkboxes
+isChanSelected = java_dialog('checkbox', 'Select information to export:', 'Compute channels labels', [], ChanNames, isChanSelected);
+if ~any(isSelect)
+    return;
+end
```
rcassani, line 200: Should not be the check for `isChanSelected`?

rcassani (changes requested):
Minimal changes, after that it should be ready to merge :)

Edouard2laire:
Done. Thanks a lot for your comments :)

Edouard2laire:
hello.
do you think you would have time to merge this next week ? :)

Issue #614, opened 2023-04-18
Multiple progress bar


Edouard2laire:
Hello, 

it would be great to have brainstorm handle multiple progress bar, like this: 
![1596470707](https://user-images.githubusercontent.com/24530402/232644319-1e6cd0d8-4e74-4c8f-9046-f34d14b3b95c.png)

## Use case

When we lunch a pipleline on a set of files, it would be great to have a progress bar indicating the how many subject have been process while still allowing the executed process to have their own progress bar (case for some process being long to execute such as MEM).  

## Toy example:

```matlab
p = bst_progress('start', 'External Loop', '', 0, 10);
for iSubject = 1:10
    
    internal_loop();
    bst_progress('inc', 1);
end
bst_progress('stop');


function internal_loop()

    N = 20;    
    bst_progress('start', 'Internal Loop', '', 0, N);
    for i = 1:N
        pause(1)
       bst_progress('inc', 1);
    end

    bst_progress('stop');

end
```

## Suggestions: 

Have bst_progress always create  a new progress bar when using bst_progress('start', ...) and have it pass as first argument of the following functions. It's working however, there is one new window per progress bar. 

```matlab
p = bst_progress('start', 'External Loop', '', 0, 10);
for iSubject = 1:10
    
    internal_loop();
    bst_progress(p,'inc', 1);
end
bst_progress(p,'stop');
```

HS :  in case, that's helfull for anyone; here is how the current progress bar can be using with parfor: 

```Matlab
    N = 20;
    q = parallel.pool.DataQueue;
    afterEach(q, @(x) bst_progress('inc', 1));
    
    bst_progress('start', 'Internal Loop', '', 0, N);
    parfor i = 1:N
        pause(1+1.2*abs(randn()))
        send(q, i); 
    end

    bst_progress('stop');
```

PR #626, opened 2023-06-13
pca fix progress bar

Moo-Marc:
Improved progress bar "nesting" for recent modifications (PCA, flattening, scouts)

### DIFF
--- a/toolbox/math/bst_pca.m
+++ b/toolbox/math/bst_pca.m
@@ -195,6 +195,10 @@ SurfaceFile = sResults.SurfaceFile;
 DisplayUnits = sResults.DisplayUnits;
 
 % Progress bar
+isProgressBar = bst_progress('isVisible');
+if isProgressBar
+    PrevProgress = bst_progress('get');
+end
 if isScout
     bst_progress('start', 'Extract scouts with PCA', sprintf('Extract scouts for %d files', nInputs), 0, 2*nInputs);
 else
@@ -661,6 +665,14 @@ switch PcaOptions.Method
 
             bst_progress('inc', 1);
         end
+end
+
+if isProgressBar
+    % Reset existing progress bar as best we can; no way to know previous title or text.
+    bst_progress('start', '', '', PrevProgress, 100);
+else
+    % Hide.
+    bst_progress('stop');
 end
 
 % Don't update the tree here since the files may be temporary (e.g. flattening from process_extract_scout).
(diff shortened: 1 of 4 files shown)


toolbox/math/bst_pca.m
```diff
@@ -663,6 +667,14 @@
         end
 end
 
+if isProgressBar
+    % Reset existing progress bar as best we can; no way to know previous title or text.
```
rcassani, line 671: Previous Title and Text could be obtained from the swing objects:
```
oldValue = bst_progress('get');
% Set is used to retrieve the bar structure
oldBar   = bst_progress('set', oldValue);            
oldTitle = oldBar.jWindow.getTitle().toCharArray';
oldText  = oldBar.jLabel.getTitle().toCharArray';
```

Edouard2laire:
This is interesting. I am wondering if instead of 'nesting' progress bar, if we should not instead allow to have multiple progress bars (#614)

Maybe something like this (https://www.mathworks.com/matlabcentral/fileexchange/26589-multiwaitbar):

Moo-Marc:
Thanks @rcassani, I'll make that a new bst_progress command: pBar = bst_progress('reset', pBar)
Would you have another suggestion instead of calling it 'reset'?  Maybe 'restart', or even overload 'start'?

Hi @Edouard2laire, indeed I saw your post at the time and like it, but I doubt we'll have resources for such improvements soon, with François being away.  This PR is just a quick fix to follow Bst "nesting conventions" - my recent changes had made some processes stick at 100% (e.g. connectivity with PCA scouts).

rcassani:
Thanks @Edouard2laire, indeed that PR is in the queue of things to do.

> Would you have another suggestion instead of calling it 'reset'? Maybe 'restart', or even overload 'start'?

I think we can add two methods 'GetBar' and 'SetBar'. 
Will do

```
oldBar = bst_progress('GetBar');
bst_progress('SetBar', oldBar);

```

toolbox/math/bst_pca.m
rcassani: Added in [05021b2](https://github.com/brainstorm-tools/brainstorm3/pull/626/commits/05021b2c73a2f9f66b77667f1879231ceb7dfc81)

rcassani:
@Moo-Marc, last two commits should do the trick. Can you test that it behaves as expected?

Moo-Marc:
@rcassani  Few additional tweaks.  Works ok now!

PR #627, opened 2023-06-21
fixed font

Edouard2laire:
Without this fixed, the font would always be arial (first font available in the list)

### DIFF
--- a/toolbox/core/bst_get.m
+++ b/toolbox/core/bst_get.m
@@ -3913,10 +3913,11 @@ switch contextName
         fontTypes = {};
         if (nargin >= 3)
             fontTypes{end + 1} = varargin{3};
+        else
+            fontTypes{end + 1} = 'Arial';  % Default font
+            fontTypes{end + 1} = 'Liberation Sans';  % Free Arial substitute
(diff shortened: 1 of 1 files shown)


rcassani:
The `Courier New` font should not be added there, only `Arial` and `Liberation Sans` as in the [proposed solution](https://github.com/brainstorm-tools/brainstorm3/pull/608#issuecomment-1515291992).  

For text "tables", as the plugin list, the font is set in [view_text@L72](https://github.com/brainstorm-tools/brainstorm3/blob/59a2e57553b15b0f18fe32643ef7b6094fca3c07/toolbox/gui/view_text.m#L72)

I'll test the different GUIs in different OSs and merge directly to master if nothing awfully breaks 🐞

rcassani:
>  Edit2: we are actually proposing the same thing. The addition of the else condition haha

Yup, I was a bit confused 😃

rcassani:
Addressed at [2163f7e](https://github.com/brainstorm-tools/brainstorm3/commit/2163f7ed20461aebc10a4d267aecc1e26570fc79)

Issue #630, opened 2023-06-24
Process -  working with continuous recording


Edouard2laire:
Hello, 

A few months ago I worked on a process to synchronize two files https://github.com/brainstorm-tools/brainstorm3/pull/397 and it was rejected because it didn't handle continuous files. I was therefore trying to make the code handle those files. It seems I am able to read continuous files: https://gist.github.com/Edouard2laire/d96006297a4a854b96d42703b0e3b715#file-process_synchronise_multiples_signals-m-L76-L82 but i am very confused about how to save new continuous files recording into the database. 

Therefore my question is, how to save continuous recordings in the database? 

A simple example is process_nst_mbll in Nirstorm. It can take both continuous recording and imported data but always create imported data:  https://github.com/Nirstorm/nirstorm/blob/master/bst_plugin/mbll/process_nst_mbll.m#L135-L153

How could I change this block of code that saves the output of the process and create a continuous recording if the input of the process was itself a continuous recording? 

Thanks a lot, 
Edouard

rcassani:
@Edouard2laire, can we consider this issue closed?

PR #634, opened 2023-07-11
Export ROI as Freesurfer annotation

Edouard2laire:
This PR is doing three things : 
- Give a better error when trying to export multiple ROI using FreeSurfer label format (that only support on ROI per file)
https://surfer.nmr.mgh.harvard.edu/fswiki/LabelsClutsAnnotationFiles

- Export multiple ROI as annotation files   https://surfer.nmr.mgh.harvard.edu/fswiki/AnnotFiles 
- Fix a small bug in the reading of annotation file  https://surfer.nmr.mgh.harvard.edu/fswiki/LabelsClutsAnnotationFiles#Annotationfiledesignsurprise

### DIFF
--- a/external/freesurfer/write_annotation.m
+++ b/external/freesurfer/write_annotation.m
@@ -1,0 +1,150 @@
+% Contact ythomas@csail.mit.edu or msabuncu@csail.mit.edu for bugs or questions 
+%
+%=========================================================================
+%
+%  Copyright (c) 2008 Thomas Yeo and Mert Sabuncu
+%  All rights reserved.
+%
+%Redistribution and use in source and binary forms, with or without
+%modification, are permitted provided that the following conditions are met:
+%
+%    * Redistributions of source code must retain the above copyright notice,
+%      this list of conditions and the following disclaimer.
+%
+%    * Redistributions in binary form must reproduce the above copyright notice,
+%      this list of conditions and the following disclaimer in the documentation
+%      and/or other materials provided with the distribution.
+%
+%    * Neither the names of the copyright holders nor the names of future
+%      contributors may be used to endorse or promote products derived from this
+%      software without specific prior written permission.
+%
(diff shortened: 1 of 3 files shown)


rcassani:
> * Export multiple ROI as annotation file

As test, you could try to import the created `.annot` file with **Scout tab > Atlas > Load atlas**

Edouard2laire:
Hello, 

This should be ready. The function write_annotation is from FreeSurfer: https://github.com/freesurfer/freesurfer/blob/dev/matlab/write_annotation.m


There is only one limitation that I am not sure how to address. In the annotation format, each vertex is not associated with a label associated with a name and a color. But each vertex is associated with a color and then the color is associated with the region's name.

 So two scouts cannot be described in an annotation file if they have the same color.  This is for example the case for atlas where the left/right scout have the same color. How should we handle it ?

rcassani:
> There is only one limitation that I am not sure how to address. In the annotation format, each vertex is not associated with a label associated with a name and a color. But each vertex is associated with a color and then the color is associated with the region's name.

The plan would be to set the repeated colors to a small variations of the original. 
I will work on this and add it to the PR

rcassani:
Duplicate colors addressed! 
Can you test the exported `.annot` in FreeView?

Edouard2laire:
it's working well (left freeview, right: Brainstorm) 
Thank you :) 

Screenshot from 2023-07-13 10-12-01


On another, topic, I know you are busy but do you think you would have time to have a look at https://github.com/brainstorm-tools/brainstorm3/issues/630 . I am giving a presentation on Nirstorm on the 20 of July and I would like to try to have this implemented before the presentation  :)

Edouard2laire:
I just thought about something this morning: is it possible to call this code from a script to be able to export the ROI from a script?

rcassani:
It is not possible as it is, as it requires user interaction to set the output file and format.

The non-interactive call for this function, and other functions with Scouts would be a nice-to-have feature

PR #635, opened 2023-07-17
fix topography (NIRS)

Edouard2laire:
Hi :)
When cloning a topography figure, I had the following error : 

```
Error using cell/ismember (line 34)
Input A of class double and input B of class cell must be cell arrays of character vectors, unless one is a character
vector.

Error in bst_getunits (line 46)
    if ismember(lower(DataType), {'results', 'sources', 'source'})

Error in bst_colormaps>ConfigureColorbar (line 1470)
                    [valScaled, fFactor, DisplayUnits] = bst_getunits( fmax, DataType, 'nirs', DisplayUnits);

Error in bst_colormaps (line 74)
eval(macro_method);

Error in panel_surface>UpdateSurfaceColormap (line 2053)
        bst_colormaps('ConfigureColorbar', hFig, ColormapInfo.Type, DataType, ColormapInfo.DisplayUnits);

Error in panel_surface (line 40)
eval(macro_method);

Error in figure_3d>ColormapChangedCallback (line 167)
    panel_surface('UpdateSurfaceColormap', hFig);
```

DataType is empty when cloning a topography. I guess it's because there is no surface associated to the topography since it's at the channel level. 

This fix is not perfect as after cloning the figure, the unit is written as 'invalid scale' until a new montage is selected and the color bar is updated. 

Edouard

### DIFF
--- a/toolbox/math/bst_getunits.m
+++ b/toolbox/math/bst_getunits.m
@@ -40,6 +40,12 @@ if (nargin < 4) || isempty(DisplayUnits)
     DisplayUnits = [];
 end
 
+% If no modality (ex: surface mask, mri values...)
+if (nargin < 3) || isempty(DataType)
+    DataType = 'none';
+end
+
+
 % Check if there is something special in the filename
 if (nargin >= 3) && ~isempty(FileName)
     % Source files
@@ -56,10 +62,7 @@ end
 
 % Consider input data in absolute value
 val = abs(val);
-% If no modality (ex: surface mask, mri values...)
-if isempty(DataType)
-    DataType = 'none';
-end
+
 
 % If the display unit is already defined
 if ~isempty(DisplayUnits)
--- a/toolbox/math/bst_getunits.m
+++ b/toolbox/math/bst_getunits.m
@@ -40,12 +40,6 @@ if (nargin < 4) || isempty(DisplayUnits)
     DisplayUnits = [];
 end
 
-% If no modality (ex: surface mask, mri values...)
-if (nargin < 3) || isempty(DataType)
-    DataType = 'none';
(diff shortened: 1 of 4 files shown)


toolbox/math/bst_getunits.m
```diff
@@ -40,6 +40,12 @@
     DisplayUnits = [];
 end
 
+% If no modality (ex: surface mask, mri values...)
+if (nargin < 3) || isempty(DataType)
+    DataType = 'none';
+end
```
rcassani: This introduced troubles in the call `bst_getunits(val, 'ValidDataType')`

rcassani:
- Bugfix on cloning NIRS topography
- Bugfix on reloading montage for cloned NIRS topography
- Bonus bugfix: 2DDisk topo should have white background

rcassani:
> [...] the unit is written as 'invalid scale' until a new montage is selected and the color bar is updated.

@Edouard2laire, I could not replicate the problem this problem. 
Can you check if it is fixed with the last commit?

Edouard2laire:
yes it is working ! thanks a lot. 

One small issue related is if you duplicate a tomography then use the surface panel to add the cortical surface, then it's only adding the cortical surface to the first topography and it's not possible to add it to the duplicated figure.

rcassani:
> One small issue related is if you duplicate a tomography then use the surface panel to add the cortical surface

Solved at [e178fed](https://github.com/brainstorm-tools/brainstorm3/pull/635/commits/e178fedb4851c7f64dc9b594ff58ef8eb94a32bf) 🎉

Edouard2laire:
it's working ! thanks a lot !!

PR #637, opened 2023-08-18
update the MEG sensors display, this PR allow the display the correct MEG sensor location

tmedani:
update the MEG sensors display, this PR allows the display of the correct MEG sensor location when displaying the LF vectors


rcassani:
Just removed the old code and added comments

tmedani:
Thanks,

PR #646, opened 2023-09-23
Distance computation (euclidean / geodesic)

Edouard2laire:
Hello. 

This PR adds a small function to compute the distance between two points using either the geodesic distance (as the number of edges separating the two points, or the actual distance) or the Euclidean distance.  When calculating the distance to a region (verticesA), return the minimum distance to that region.  Units are either mm or number of edge.

This was asked a few times on the forum: 
https://neuroimage.usc.edu/forums/t/geodesic-distance/1483/6

Here is an example showing the distance to the green ROI using the Euclidean distance, geodesic ( number of edge) or geodesic (path length): 

Screenshot 2023-09-23 at 17 40 32

This needs a few more test, but I think it's ready to be reviewed :) 

HS: the last commit might help with the smoothing function.  
``` Dist = bst_tess_distance(SurfaceMat, 1:n, 1:n, 'geodesic_length', 1);``` returns a matrix where Dist(I,j) contains the geodesic distance between the vertex i, and j needed in https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/anatomy/tess_smooth_sources.m#L117-L168

Edit: done :) I guess we should only keep the function using the actual geodesic since it's not that slow (a few seconds)
Maybe we can merge this with the other smoothing process from #645 

Useful link: https://www.mathworks.com/help/matlab/ref/graph.distances.html

### DIFF
--- a/toolbox/anatomy/tess_smooth_sources.m
+++ b/toolbox/anatomy/tess_smooth_sources.m
@@ -58,7 +58,7 @@ nv = size(Vertices,1);
 Sigma = FWHM / (2 * sqrt(2*log2(2)));
 
 [vi,vj] = find(VertConn);
-meanDist = mean(sqrt((Vertices(vi,1) - Vertices(vj,1)).^2 + (Vertices(vi,2) - Vertices(vj,2)).^2 + (Vertices(vi,3) - Vertices(vj,3)).^2));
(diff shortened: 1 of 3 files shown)


tmedani:
Hi @Edouard2laire 

Thanks for this PR  & contribution. 
We will review it ASAP.

PR #647, opened 2023-09-27
Coregistration and exporting it to BIDS

Moo-Marc:
This is still a work in progress, a revised version of this previous closed PR: https://github.com/brainstorm-tools/brainstorm3/pull/511
I'm putting it here now to share it in its current state with users that also needed this functionality.
The main features are:

- The ability to replace the MRI anatomical fiducials with the digitized ones, meant to be done after automatic or manual coregistration.  Currently added to the process: Import > Channel File > Adjust Coordinate System > "Replace MRI nasion and ear points with digitized landmarks (cannot undo)"
- A script (not yet accessible in the GUI) to export the coregistration in the original BIDS metadata files, based on the MRI anat fids.  _bst_save_coregistration.m_

Note: This relatively old branch of mine contains other related changes, e.g. changes to generating the head surface and the automatic registration with head points, which would likely be removed from this PR prior to being ready for review.


toolbox/anatomy/tess_isohead.m
```diff
@@ -85,7 +91,7 @@
 %% ===== ASK PARAMETERS =====
 % Ask user to set the parameters if they are not set
 if (nargin < 4) || isempty(erodeFactor) || isempty(nVertices)
-    res = java_dialog('input', {'Number of vertices [integer]:', 'Erode factor [0,1,2,3]:', 'Fill holes factor [0,1,2,3]:', '<HTML>Background threshold:<BR>(guessed from MRI histogram)'}, 'Generate head surface', [], {'10000', '0', '2', num2str(sMri.Histogram.bgLevel)});
+    res = java_dialog('input', {'Number of vertices [integer]:', 'Erode factor [0,1,2,3]:', 'Fill holes factor [0,1,2,3]:', '<HTML>Background threshold:<BR>(guessed from MRI histogram)'}, 'Generate head surface', [], {'15000', '0', '0', num2str(sMri.Histogram.bgLevel)});
```
Edouard2laire, line 94: what is the rationale for changing the default value? Unless there is a very good reason, i think we should not change the default value as it might be very confusing for most user

PR #654, opened 2023-11-08
Add ct2mrireg plugin

chinmaychinara91:
Link for ct2mrireg: https://github.com/ajoshiusc/USCCleveland/tree/master/ct2mrireg

### DIFF
--- a/toolbox/io/import_mri.m
+++ b/toolbox/io/import_mri.m
@@ -89,11 +89,11 @@ if isempty(MriFile)
 
     % Get MRI/CT file
     [MriFile, FileFormat, FileFilter] = java_getfile( 'open', ...
-    ['Import ' volType '...'], ...   % Window title
-    LastUsedDirs.ImportAnat, ...      % Default directory
-    'multiple', 'files_and_dirs', ... % Selection mode
-    bst_get('FileFilters', 'mri'), ...
-    DefaultFormats.MriIn);
+        ['Import ' volType '...'], ...   % Window title
+        LastUsedDirs.ImportAnat, ...      % Default directory
+        'multiple', 'files_and_dirs', ... % Selection mode
+        bst_get('FileFilters', 'mri'), ...
+        DefaultFormats.MriIn);
     % If no file was selected: exit
     if isempty(MriFile)
         return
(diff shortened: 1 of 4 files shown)


chinmaychinara91:
Users can now import CT and do co-registration with the default reference MRI using [CT2MRIREG](https://github.com/ajoshiusc/USCCleveland/tree/master/ct2mrireg) plugin.

**IMPORT CT option:**

import_ct

**CT2MRIreg option:**

ct2mri

rcassani (commented):
Looks good!
Added few commits to cleanup the code

toolbox/io/import_mri.m
```diff
@@ -379,8 +352,8 @@
 
 
 %% ===== SAVE MRI IN BRAINSTORM FORMAT =====
-% Add a Comment field in MRI structure, if it does not exist yet and if not importing a CT
-if ~isempty(Comment) && ~strcmpi(Comment, 'Import CT')
+% Add a Comment field in MRI structure, if it does not exist yet
+if ~isempty(Comment)
```
chinmaychinara91: @rcassani we need to add a condition so that the correct file name is displayed:

Currently it just shows this after processing:
Screenshot 2023-11-09 094704

It should display:
Screenshot 2023-11-09 094800

if we do this as under it solves the issue:

`if ~isempty(Comment) && ~strcmpi(volType, 'CT')`

toolbox/io/import_mri.m
```diff
@@ -269,28 +256,22 @@
             strOptions = '<HTML>How to register the new volume with the reference image?<BR>';
             cellOptions = {};
 
-            if strcmpi(Comment, 'Import CT')
+            % Register with the SPM
```
chinmaychinara91: @rcassani Regarding line 259-264 they should be under 'else' statement as when we choose CT as per my understanding we do not want SPM and MNI buttons to show up rather just CT2MRI, correct me if I am wrong.

chinmaychinara91: Discussion with @tmedani , @rcassani:

The reason why this CT2MRI was written in the first place was because the current SPM and MNI coreg methods work for most CT-MRI coreg but there is no proper validation reference. Hence, for now we keep SPM and MNI as optional methods while doing CT-MRI coreg but CT2MRI is the best validated one.

toolbox/io/import_mri.m
rcassani: Solved at ca09b1a

chinmaychinara91: Looks good ! Thanks @rcassani

chinmaychinara91:
I guess it can be merged now.

PR #663, opened 2023-12-05
Brainstorm: Apple silicon support

Edouard2laire:
This PR is here to use the maint branch of SPM when using an ARM Mac (which requires new Mex files) and fix a small bug

Note: this cannot be merged as is as brainstorm doesn't really like that SPM is in the folder called spm12-maint. But this works after adding SPM manually to the path: addpath('/Users/edelaire1/.brainstorm/plugins/spm12/spm12-maint')

Edit: actually would that be ok with you if we directly pull from the main branch of SPM: https://github.com/spm/spm ?


rcassani:
Since the only difference between the `main` (the download link) and `maint` branches are the `*.mexmaca64` files. It would make more sense to change the plugin to point to `maint` It should work by changing `PlugDesc(end).Version` from `latest` to `github-maint` and using the `maint` archive ULR.

- https://github.com/spm/spm12/commit/c28485d46371138ccc93a4bd46cf151113e7f43b
- https://neuroimage.usc.edu/forums/t/should-use-the-latest-version-of-spm12-for-better-compatible-for-apple-silicon/42703

Edouard2laire:
Like that or would you change the version for everyone? I guess most users don't need the new mex.

rcassani:
The switch option should be ok. Not only saves the 6 MB for the Apple silicon binaries, but also saves the user of downloading from GitHub which can be slow, SPM is around 200 MB. I hope the zip in their page get updated at some point.

Regarding the duplicated binaries, the opposite approach will be done (later push by me also in this PR): 
1. Add the to external also the Apple silicon ones
2. Remove them from the installed SPM as the binaries in external are the ones that add the _In-Brainstorm_ MNI normalization (maff) method 😉

P.S. Adding the small bugfix for not existent plugin process directly in `master`

Edouard2laire:
Thanks. let me know if i can help. 

Maybe we should identify the toolbox that doesn't have new mex yet like iso2mesh, Mcxlab, and put a warning or something indicating that they cannot be used on ARM mac with matlab2023b.

rcassani:
These things are needed:

* A warning in the startup to indicate that Brainstorm is running on Apple silicon and some stuff may not work as expected, and recommend to use the Intel version. 
* Add the Apple silicon binaries for external SPM
* Add warnings in the plugins that do not have support yet

Edouard2laire:
> Edit: It is quite a lot what is not supported yet:
> https://www.mathworks.com/support/requirements/apple-silicon.html

it might seem like a lot but this is actually not really impacting us as its mainly packages for  Simulink or specific decide like Arduino boards.

rcassani:
This is what I have check so far 
(11-dec-2023)

  ❓ 'adi-sdk'
  ✅ 'axion'
  ❓ 'bci2000'
  ✅ 'blackrock'
  ❌ 'brain2mesh' (iso2mesh)
  ❌ 'brainentropy'
  ✅ 'cat12'
  ❌ 'ct2mrireg'
  ✅ 'derivelfp'
  ❌ 'duneuro'
  ✅ 'easyh5'
  ✅ 'fastica'
  ❌ 'fieldtrip'
  ❌ 'iso2mesh'
  ✅ 'jsnirfy'
  ❓ 'kilosort'
  ❓ 'kilosort-wrapper'
  ❓ 'libsvm'
  ❓ 'mcxlab-cl'
  ❓ 'mcxlab-cuda'
  ❓ 'mff'
  ✅ 'mia'
  ✅ 'neuroelectrics'
  ✅ 'neuromaps'
  ❌ 'nirstorm' (brainentropy)
  ✅ 'npy-matlab'
  ❓ 'nwb'
  ❌ 'openmeeg'
  ❓ 'phy' (?)
  ✅ 'picard'
  ❌ 'plexon'
  ❓ 'plotly'
  ❌ 'roast' (iso2mesh)
  ❌ 'simmeeg' (fieldtrip)
  ✅ 'spm12'
  ❓ 'tdt-sdk'
  ✅ 'ultramegasort2000'
  ❌ 'waveclus'
  ❓ 'xdf'

Edouard2laire:
For brainentropy, if fminunc is available, then its already working as part of the optimisation toolbox

rcassani:
Added an error message for the Plugins that have been already tested.
 
As testing exhaustively all plugins may take a long time, due to lack of HW to test on my side. I think we should merge this PR for the moment and update the list (in [f72d0a6](https://github.com/brainstorm-tools/brainstorm3/pull/663/commits/f72d0a607ef49fc5f81906ad8ce07cf11792a47e)) with future tests and reports from users. What do you think?

Edouard2laire:
Also added an error when loading an already downloaded plugin. I'll try to compile them slowly first with iso2mesh and mxclab.

Edouard2laire:
update. I was able to compile iso2mesh today: https://github.com/fangq/iso2mesh/pull/75

Some tests are needed but when this is merged; we should be able to check that from our todo list :)

PR #669, opened 2023-12-14
Expose update plugDesc function

Edouard2laire:
Small update to expose a function to update the plugin.mat file of a plugin.

This is quite useful at developing plugins as it allows adding/removing processes without having to manually change the plugin.mat file.
 
For example, the two following lines can now be used to update the mat file for NIRSTORM: 

```matlab
PlugDesc = bst_plugin('GetInstalled', 'NIRSTORM');
% OR
% [isOk, errMsg, PlugDesc] = bst_plugin('Load', 'NIRSTORM');

[isOk, errMsg, PlugDesc] = bst_plugin('UpdateDescription', PlugDesc);
```

### DIFF
--- a/toolbox/core/bst_plugin.m
+++ b/toolbox/core/bst_plugin.m
@@ -1517,72 +1517,9 @@ function [isOk, errMsg, PlugDesc] = Install(PlugName, isInteractive, minVersion)
         return;
     end
     
-    % === DELETE UNWANTED FILES ===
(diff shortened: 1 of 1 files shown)


rcassani:
* Changed default behaviour to explicit removal
* Added extra check that the plugin to modify need to be installed (it can be unloaded though)

rcassani:
Such a mess! 🥲 
The missing commit is in [c67d110](https://github.com/brainstorm-tools/brainstorm3/commit/c67d110db14955521bbf53733eb53e04757c41a0)

PR #682, opened 2024-02-14
New Digitize panel

Moo-Marc:
This adds a new version of the Digitize panel, in parallel to the current version.
Main changes:
- Added settings: 
  - set of fiducials to digitize, allowing e.g. a 4th MEG head coil.
  - Polhemus device configuration commands, e.g. "hemisphere of operation"
  - Distance threshold for warning message
- Large display of next point to digitize, to be visible away from the desk
- Added "beep" and more visible error message when repeated measures > 5mm (default distance threshold)
- New button to add a set of fiducials to be digitized at any moment, e.g. at end to confirm receiver didn't move and points are good
- Uses serialport instead of serial (no more warnings), but requires R2019 or newer.
- Help menu linking to Digitize tutorial
- New menu item to switch versions of the panel: original or 2024.  The panel closes and must be reopened after switching.

digitize_2024
digitize_2024_settings

The tutorial is not yet updated. 
This version is now installed in our lab to use for a while before merging.  Others may want to test it also. (@tmedani?)


Moo-Marc:
For Brainstorm "news", we could describe the main advantages as:
New Digitize version: can recheck fiducials, supports more fids, customizable device config, improved GUI

tmedani:
> This adds a new version of the Digitize panel, in parallel to the current version. Main changes:

> 
> The tutorial is not yet updated. This version is now installed in our lab to use for a while before merging. Others may want to test it also. (@tmedani?)

Hey @Moo-Marc 
Sorry for the very late reply, 
I saw this PR, and I didn't reply at that time, and then I forgot to follow up.

Our digitizer was down for a few months and we sent it for reparation, it is back for few day, 
I'll try to plug it in and test it as soon as I can. 

our experimental team is now using the 3D Scanner instead [https://neuroimage.usc.edu/brainstorm/Tutorials/TutDigitize3dScanner]
Much easier to use and less errors. 

I'll check this PR asap, and back with feedback

rcassani (approved):

PR #684, opened 2024-02-19
CT integration and Generation of thresholded mesh from CT

chinmaychinara91:
# CT integration

This PR adds: 
* New node type `'volct'` for CT data, icon and tooltip (It requires `brainstorm.jar` from [here](https://github.com/rcassani/bst-java/blob/add-ct/brainstorm/dist/brainstorm.jar))

image

* Updated popup menu in the database tree for `'volct'`
    * MRI-exclusive actions are removed
    * Do not allow to be Subject's default MRI
    * Do not allow editing MRI form CT

image

* Example of updating a process to handle MRI and CT: `process_mri_deface.m`
image

# Generation of thresholded mesh from CT
**[THESE ARE USED AS TEST DATA BELOW](https://drive.google.com/drive/folders/1TI8wANAb09erddYP52bXyXo5xuajLCNf?usp=sharing)**

**NOTE: This is just a CT specific feature**

1. Assuming we have a CT by following [Import CT](https://neuroimage.usc.edu/brainstorm/Tutorials/ImportAnatomy#Import_CT)

2. Right-click on the **CT → MRI segmentation → Generate mesh from CT** . 

3. This will bring the **Generate isosurface** window. This window shows 3 things in the Hounsfield Unit, the Background level, the White level and the Set isoValue fields. The first and second values are calculated automatically from the histogram of the CT. This is to help the user to decide on what to set the isoValue for getting a good thresholded mesh. Another way is to refer to this https://en.wikipedia.org/wiki/Hounsfield_scale. The best way, however is to just open the CT in the MRI viewer browse through the electrode intensities and do trial and error to get the best value for a clean mesh. In this case, setting the value to **2400** gave a good result. Set the **isoValue** to **2400** and press **OK.** 

4. This will generate a isosurface which looks something like under. **THIS WILL HELP IN DOING SEEG ELECTRODE CONTACT LOCALIZATION**.


toolbox/anatomy/tess_isosurface.m
```diff
@@ -0,0 +1,156 @@
...
+
+%% ===== ASK PARAMETERS =====
+% Ask user to set the parameters if they are not set
+if (nargin < 2) || isempty(isoValue)
+    res = java_dialog('input', {'<HTML>Background level (HU):<BR>(guessed from MRI histogram)', ...
+                                '<HTML>White level (HU):<BR>(guessed from MRI histogram)', ...
+                                '<HTML>Set isoValue (HU):'}, ...
+                                'Generate isosurface', ...
+                                [], ...
+                                {num2str(sMri.Histogram.bgLevel), num2str(sMri.Histogram.whiteLevel), num2str(1900)});
```
rcassani, line 96: The first and second inputs are not used for now, right?
If they are not going to be used in the future, we should just print them instead of being a editable fields.

rcassani, line 96: I'm not familiar with CT data. 
Will this guess number (1900) change from CT to CT?
If so, we can we propose a better guess based on the CT data?

toolbox/anatomy/tess_isosurface.m
```diff
@@ -0,0 +1,156 @@
...
+    % If user cancelled: return
+    if isempty(res)
+        return
+    end
+    % Get new value isoValue
+    isoValue = round(str2double(res{3}));
+else
+    isoValue = sMri.Histogram.whiteLevel;
```
rcassani, line 104: Potential error, this overrides the input argument given to the function

rcassani (changes requested):
Overall the code looks good to me.
Few things to address.

rcassani:
@chinmaychinara91, check the review above.

There are few commits that I want to add on integration of CT now in Brainstorm, and I think this PR is a good place to put them so we take it as common place for further development in the sEEG contact location. What do you think?  

The two small modifications on `mri_coregister` added to this branch, as part of those modifications.

toolbox/anatomy/tess_isosurface.m
chinmaychinara91: That is correct. Let me make these changes.

chinmaychinara91: That was my thought as well. But from all the CT data I have encountered till now that seems to be a good number based on manual inspection as mentioned in the PR description above. For now let's keep it to 1900. But if we find any new ways to better initialize this value in the future will update it.

toolbox/anatomy/tess_isosurface.m
chinmaychinara91: Yes thanks for catching this.

chinmaychinara91:
@rcassani:

The `mri_coregister` changes look good. Others I will update.

chinmaychinara91:
@rcassani updated with the requested changes.

chinmaychinara91:
just found another issue. let me update it.

toolbox/anatomy/tess_isosurface.m
chinmaychinara91: Check https://github.com/brainstorm-tools/brainstorm3/pull/684/commits/ea8777bffd01536a3559a3860c444237a1433900. Found a better way using `sMRI.Histogram.maxIntensity`.  Added a parameter check for that as well since anything above `sMRI.Histogram.maxIntensity` retunrs empty and there is nothing to compute.

chinmaychinara91:
@rcassani done from my end.

rcassani:
I have updated the integration of CT data into Brainstorm: 
 * By creating its own node type we can handle better its behaviour, display and menus
 * The MRI menus that do not apply to CT have been removed
     * @tmedani, What do you think of the updated menu for CT? 
       Do we want to keep the 'MNI normalization' option for CT data? 
     
@chinmaychinara91, please test that I did not break anything with these changes.
To test the new node type the file:
1. The file `brainstorm3\java\brainstorm.jar` must be replaced with the one in this one: https://github.com/rcassani/bst-java/blob/add-ct/brainstorm/dist/brainstorm.jar
2. Matlab needs to be restarted as well

---
The title and description of this PR have been updated to match the contents

tmedani:
> I have updated the integration of CT data into Brainstorm:
> 
> * By creating its own node type we can handle better its behaviour, display and menus
> * The MRI menus that do not apply to CT have been removed
>   
>   * @tmedani, What do you think of the updated menu for CT?

I need to check the menu first for a complete feedback, 

>     Do we want to keep the 'MNI normalization' option for CT data?

I think we can keep it since it can be used in some other scenario, and it is always useful.

> The MRI 3D slices or,

Let's do it on the 3D MRI viewer, it has more significance.

toolbox/anatomy/tess_isosurface.m
```diff
@@ -0,0 +1,155 @@
+function [MeshFile, iSurface] = tess_isosurface(iSubject, isoValue, Comment)
+% TESS_ISOSURFACE: Reconstruct a thresholded surface mesh from a CT
+%
+% USAGE:  [MeshFile, iSurface] = tess_isosurface(iSubject, isoValue=1900, Comment)
+%         [MeshFile, iSurface] = tess_isosurface(MriFile,  isoValue=1900, Comment)
+%         [Vertices, Faces]    = tess_isosurface(sMri,     isoValue=1900)
```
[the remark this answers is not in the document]
chinmaychinara91: doing it now

chinmaychinara91:
Also we both reviewed the changes they look great and work perfect.

toolbox/tree/tree_callbacks.m
```diff
@@ -176,6 +176,19 @@
                 else
                     view_mri(filenameRelative);
                 end
+
+            % ===== VOLUME CT =====
+            case 'volct'
```
chinmaychinara91, line 181: can we maybe merge `case 'volct'` with `case 'volatlas'` to `case {'volatlas', 'volct'}` as they are doing the same thing essentially ?

rcassani: Sounds good. Done at [8810927](https://github.com/brainstorm-tools/brainstorm3/pull/684/commits/8810927dcdea4f047b1f5c49ee4e7cf46e2d664a)

chinmaychinara91:
all things look good @rcassani. Just wrt https://github.com/brainstorm-tools/brainstorm3/pull/684#discussion_r1501248906, as you suggested and as @tmedani agreed, we can add loading of just the MRI 3D slices along with the new mesh so that we get a good reference for the contacts.

chinmaychinara91:
> all things look good @rcassani. Just wrt [#684 (comment)](https://github.com/brainstorm-tools/brainstorm3/pull/684#discussion_r1501248906), as you suggested and as @tmedani agreed, we can add loading of just the MRI 3D slices along with the new mesh so that we get a good reference for the contacts.

Updated and added this.

rcassani:
Great, I'm merging this PR!
@chinmaychinara91, please update the "Generation of thresholded mesh from CT" in the PR description to match the final version

rcassani (approved):

PR #685, opened 2024-02-19
Update projector structure. And add ICA explained variance

rcassani:
This PR adds code to compute the variance explained by the IC when using ICA decomposition.

The reasonable place to store these values is the `SingVal` field in the `projector` structure. However, this changes for back compatibility are needed as this field is currently either the PCA explained variance, or a string to identify other projectors `ICA` or `REF` (montage).

https://github.com/brainstorm-tools/brainstorm3/blob/e532c6d83badc93c1476004cc493f3df08a2a369/toolbox/process/functions/process_ssp2.m#L947-L952

~~Tentative solution:~~ 
~~1. Use the Comment to identify the type of projector~~
~~2. Always require the projector Comment to have the type of projector on it~~

Implemented solution:
- [x] Add field 'Method' to 'Projector' structure to identify Projector
- [x] Support projectors without 'Method'. Get projector method from its content
- [x] Store ICA explained variance in `SingVal`

References:
https://github.com/mne-tools/mne-python/pull/11141

---
Being `F` the original data, the independent components `Y` can be obtained by multiplying the unmixing matrix `W` with `F`. Then we can mix those ICs `Y` with the mixing matrix `M` (which is the inverse of `W`) to recover the observed data `F` this is to say `F'`

`Y = W * F`
`F' = M * Y`

Then we can compute the variance explained by each IC as: 

`varIcs = sum(M.^2, 1) .* transpose(sum(Y.^2, 2))`
`varIcs = varIcs / sum(varIcs)`

However this is the explained variance of `F'`, so these explained variances need to be scaled with ratio of variance between `F'` and `F`.

`Fdiff = (F - M * Y)`
`rVarExp = 1 - (sum(sum(Fdiff.^2, 2)) ./ sum(sum(F.^2, 2)))`

Variance in original data by each component
`varIcs = rVarExp * varIcs`


toolbox/gui/panel_ssp_selection.m
```diff
@@ -448,35 +448,42 @@ function PlotComponents(UseSmoothing, isPlotTopo, isPlotTs)
...
-            % Display name
-            strDisplay = 'SSP';
+        componentType = sCat.Method(1:3);
+        switch lower(componentType)
+            % ICA: Get the topography to display
+            case 'ica'
+                % Explained variance
+                Singular = [];
+                if ~isempty(sCat.SingVal)
+                    Singular = sCat.SingVal ./ sum(sCat.SingVal);
```
rcassani, line 458: They do not necessarily sum 1
For ICA, `Singular` contains the fraction of explained original variance

toolbox/gui/panel_ssp_selection.m
```diff
@@ -707,24 +710,31 @@ function UpdateComp()
...
-                iDispComp = intersect(iDispComp, 1:20);
-                % Always show at least the 10 first components
-                iDispComp = union(iDispComp, 1:min(10,length(Singular)));
+            switch lower(componentType)
+                % ICA: Show all components
+                case 'ica'
+                    % Explained variance
+                    Singular = [];
+                    if ~isempty(sCat.SingVal)
+                        Singular = sCat.SingVal ./ sum(sCat.SingVal);
```
rcassani, line 721: Same comments as above:
https://github.com/brainstorm-tools/brainstorm3/pull/685/files#r1745702951

rcassani: Solved [6081e1e](https://github.com/brainstorm-tools/brainstorm3/pull/685/commits/6081e1e0a11774441e8012cb3f010824086947b3)

toolbox/gui/panel_ssp_selection.m
rcassani: Solved [6081e1e](https://github.com/brainstorm-tools/brainstorm3/pull/685/commits/6081e1e0a11774441e8012cb3f010824086947b3)

toolbox/io/out_fopen_bst.m
```diff
@@ -37,7 +37,8 @@
 sFileOut.header.starttime = sFileOut.prop.times(1);
 sFileOut.header.navg      = sFileOut.prop.nAvg;
 % sFileOut.header.version   = 51;   % April 2019
-sFileOut.header.version   = 52;   % March 2023
+% sFileOut.header.version   = 52;   % March 2023
+sFileOut.header.version   = 53;   % August 2024
```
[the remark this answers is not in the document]
rcassani: Solved [7cf5937](https://github.com/brainstorm-tools/brainstorm3/pull/685/commits/7cf5937dce5bf3453c9851990d737c07bfe0410f)

PR #688, opened 2024-02-29
view_channel_3d: take an already created figure as input

Edouard2laire:
This PR allows the following code to work: 

```matlab

hFig  = bst_figures('GetFigureWithSurface', SurfaceFile);
hFig  = view_surface_data(SurfaceFile, OverlayFile, [], hFig, 1);
hFig  = view_surface(ScalpFile, 0, [], hFig, 1);
hFig  = view_channels_3d(channelFile, 'NIRS-BRS', 'scalp', 0, 1,hFig);

```

This code produce the following image and solve my problem from here: https://neuroimage.usc.edu/forums/t/visualizing-sensors-on-top-of-cortical-activity/44950/2

image

hs: is there a better way to create an empty figure than : hFig bst_figures('GetFigureWithSurface', SurfaceFile); ?

### DIFF
--- a/toolbox/gui/view_channels_3d.m
+++ b/toolbox/gui/view_channels_3d.m
@@ -43,7 +43,7 @@ global GlobalData;
 iDS = [];
 iFig = [];
 if (nargin < 6) || isempty(hFig)
-    hFig = [];
+    hFig = 'NewFigure';
 elseif ishandle(hFig)
     hFig = bst_figures('GetFigure', hFig);
 else
@@ -101,9 +101,7 @@ if ~isempty(sSubject)
                     case 'ECOG',  SurfAlpha = .2;
                     otherwise,    SurfAlpha = opaqueAlpha;
                 end
-                if isempty(hFig)
-                    hFig = view_surface(SurfaceFile, SurfAlpha, [], 'NewFigure');
-                end
+                hFig = view_surface(SurfaceFile, SurfAlpha, [], hFig);
             end
         case 'innerskull'
             if ~isempty(sSubject.iInnerSkull) && (sSubject.iInnerSkull <= length(sSubject.Surface))
@@ -115,9 +113,7 @@ if ~isempty(sSubject)
                     case 'ECOG',  SurfAlpha = .2;
                     otherwise,    SurfAlpha = opaqueAlpha;
                 end
-                if isempty(hFig)
-                    hFig = view_surface(SurfaceFile, SurfAlpha, [], 'NewFigure');
-                end
+                hFig = view_surface(SurfaceFile, SurfAlpha, [], hFig);
             end
         case 'scalp'
             if ~isempty(sSubject.iScalp) && (sSubject.iScalp <= length(sSubject.Surface))
@@ -136,9 +132,7 @@ if ~isempty(sSubject)
                     otherwise
                         SurfAlpha = opaqueAlpha;
                 end
-                if isempty(hFig)
-                    hFig = view_surface(SurfaceFile, SurfAlpha, [], 'NewFigure');
-                end
+                hFig = view_surface(SurfaceFile, SurfAlpha, [], hFig);
             end
         case {'anatomy', 'subjectimage'}
             if ~isempty(sSubject.iAnatomy) && (sSubject.iAnatomy <= length(sSubject.Anatomy))
@@ -146,10 +140,9 @@ if ~isempty(sSubject)
                     SurfaceFile = sSubject.Anatomy(sSubject.iAnatomy).FileName;
                 end
                 SurfAlpha = .1;
-                if isempty(hFig)
-                    hFig = view_mri_3d(SurfaceFile, [], SurfAlpha, 'NewFigure');
-                end
+                hFig = view_mri_3d(SurfaceFile, [], SurfAlpha, hFig);
             end
+        otherwise
     end
 end
 % Warning if no surface was found
--- a/toolbox/gui/view_channels_3d.m
+++ b/toolbox/gui/view_channels_3d.m
@@ -1,4 +1,4 @@
-function [hFig, iDS, iFig] = view_channels_3d(FileNames, Modality, SurfaceType, is3DElectrodes, isDetails)
+function [hFig, iDS, iFig] = view_channels_3d(FileNames, Modality, SurfaceType, is3DElectrodes, isDetails,hFig)
 % VIEW_CHANNELS_3D: Display channel files on top of subject anatomy.
 
 % @=============================================================================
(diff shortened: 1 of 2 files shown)


rcassani:
> hs: is there a better way to create an empty figure than : hFig = bst_figures('GetFigureWithSurface', SurfaceFile); ?

Not sure what is the question here. If the surface is not plotted in any figure, hFig is just empty, no figure is created. Otherwise, hFig is the handle of the figure where the surface is plotted.

rcassani:
> `'NewFigure'`

That's why I got confused, that option is just there to be used

rcassani:
@Edouard2laire with the latest changes, the image can be plot as:

```
hFig  = view_surface_data(SurfaceFile, OverlayFile, [], 'NewFigure', 1);
hFig  = view_channels_3d(channelFile, 'NIRS-BRS', 'scalp', 0, 1, hFig);
```

And if the scalp is not needed, just leave that string empty `''`
I think this is ready to be merge, let me know if you agree

Edouard2laire:
> > `'NewFigure'`
> 
> That's why I got confused, that option is just there to be used


Yes, I don't know why didn't see that at first and was trying very complex things... haha

Yes, it's working. 
Thanks a lot

chinmaychinara91:
@rcassani @Edouard2laire this change is creating new figures all the time. I am actually using the same function for SEEG work.

In `tree_callbacks.m`, when I modify this:

```
elseif strcmpi(DisplayMod{1}, 'SEEG')
    hFig = bst_figures('GetFiguresByType', {'3DViz'});
    hFig = DisplayChannels(bstNodes, DisplayMod{1}, 'anatomy', 1, 0, hFig);
```
The `DisplayChannels` function uses `view_channels_3d.m`

So if I load a channel file and double click on the channel file shown below, rather than just updating the figure already displayed it keep generating new figures. Let me know if this is expected.
Screenshot 2024-02-29 132959

Edouard2laire:
> @rcassani @Edouard2laire this change is creating new figures all the time. I am actually using the same function for some SEEG work.
> 
> In `tree_callbacks.m`, when I modify this:
> 
> ```
> elseif strcmpi(DisplayMod{1}, 'SEEG')
>     hFig = bst_figures('GetFiguresByType', {'3DViz'});
>     hFig = DisplayChannels(bstNodes, DisplayMod{1}, 'anatomy', 1, 0, hFig);
> ```
> 
> The `DisplayChannels` function uses `view_channels_3d.m`
> 


Oh that is interesting.  I am not sure if that is expected.  Can you check if hfig is sent to view_channels_3d when called through  DisplayChannels ?

chinmaychinara91:
> > @rcassani @Edouard2laire this change is creating new figures all the time. I am actually using the same function for some SEEG work.
> > In `tree_callbacks.m`, when I modify this:
> > ```
> > elseif strcmpi(DisplayMod{1}, 'SEEG')
> >     hFig = bst_figures('GetFiguresByType', {'3DViz'});
> >     hFig = DisplayChannels(bstNodes, DisplayMod{1}, 'anatomy', 1, 0, hFig);
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The `DisplayChannels` function uses `view_channels_3d.m`
> 
> Oh that is interesting. I am not sure if that is expected. Can you check if hfig is sent to view_channels_3d when called through DisplayChannels ?

I did check it is sent. I think I found the issue:
in `view_channels_3d` if we do a `if strcmpi(hFig, 'NewFigure')` check before every:
`hFig = view_mri_3d(SurfaceFile, [], SurfAlpha, hFig);` that kind of solves it

rcassani:
> @rcassani @Edouard2laire this change is creating new figures all the time. I am actually using the same function for some SEEG work.

I may be wrong, but isn't this the current behaviour in the master branch?

chinmaychinara91:
> > @rcassani @Edouard2laire this change is creating new figures all the time. I am actually using the same function for some SEEG work.
> 
> I may be wrong, but isn't this the current behaviour in the master branch?

@rcassani let me know if the suggestion int the previous comment works

Edouard2laire:
I am curious as to why the if is needed. If hFig is a valid figure, then hFig = view_mri_3d(SurfaceFile, [], SurfAlpha, hFig); should not create a new figure; maybe its an issue to fix inside view_mri_3d instead of adding an if.

rcassani:
@chinmaychinara91, the double plotting comes from `view_mri_3d` at [L133-L137](https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/gui/view_mri_3d.m#L133-L137)

The proposed solution kind of works, as skips to plot the 'anatomy' if a handle is given, but it does not check if the anatomy (3D slices) are already there. It will not work properly with:

```
hFig  = view_surface(ScalpFile, 0, [], [], 1);
hFig  = view_channels_3d(channelFile, 'SEEG', 'anatomy', 0, 1, hFig);
```

How do like the proposed change in [916bc36](https://github.com/brainstorm-tools/brainstorm3/pull/688/commits/916bc36cad67da1f43a314fde2166f9958b5fd34)?

---
> If hFig is a valid figure, then hFig = view_mri_3d(SurfaceFile, [], SurfAlpha, hFig); should not create a new figure; maybe its an issue to fix inside view_mri_3d instead of adding an if. 

yes, it seems that was the case

toolbox/gui/view_channels_3d.m
```diff
@@ -136,20 +132,17 @@
...
                 if isempty(SurfaceFile)
                     SurfaceFile = sSubject.Anatomy(sSubject.iAnatomy).FileName;
                 end
                 SurfAlpha = .1;
-                if isempty(hFig)
-                    hFig = view_mri_3d(SurfaceFile, [], SurfAlpha, 'NewFigure');
-                end
+                hFig = view_mri_3d(SurfaceFile, [], SurfAlpha, hFig);
             end
+        otherwise
```
Edouard2laire, line 145: Should an error be thrown in this case?

rcassani: I left it like this with the idea that any invalid surface type will just plot the sensors

chinmaychinara91:
> @chinmaychinara91, the double plotting comes from `view_mri_3d` at [L133-L137](https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/gui/view_mri_3d.m#L133-L137)
> 
> The proposed solution kind of works, as skips to plot the 'anatomy' if a handle is given, but it does not check if the anatomy (3D slices) are already there. It will not work properly with:
> 
> ```
> hFig  = view_surface(ScalpFile, 0, [], [], 1);
> hFig  = view_channels_3d(channelFile, 'SEEG', 'anatomy', 0, 1, hFig);
> ```
> 
> How do like the proposed change in [916bc36](https://github.com/brainstorm-tools/brainstorm3/pull/688/commits/916bc36cad67da1f43a314fde2166f9958b5fd34)?
> 
> > If hFig is a valid figure, then hFig = view_mri_3d(SurfaceFile, [], SurfAlpha, hFig); should not create a new figure; maybe its an issue to fix inside view_mri_3d instead of adding an if.
> 
> yes, it seems that was the case

works perfect. Thanks @rcassani.

rcassani:
@chinmaychinara91, @Edouard2laire 
This is a nice new feature, thank you both!

PR #690, opened 2024-03-05
[EDF] reading EDF using fieldtrip

Edouard2laire:
Hello, 

The main point of this PR is to use fieldtrip to open edf containing multiple frequencies as the reader of edf is automatically resampling signals with different frequencies. 

it's a bit hacky but it seems to be working and it might be easier than rewritting our current edf reader to account for different frequencies   

Edouard


Edouard2laire:
hi @rcassani, 

What do you think of that PR? Would that solution be ok for you?

Edouard2laire:
up :) i think this is a good solution for now; we can remove it when brainstorm supports edf with multiple frequency :)

rcassani:
Hi @Edouard2laire, just minor improvements to the proposed PR.

Interestingly, although using FieldTrip to upsample works, there is a warning from FieldTrip in the console. Do you get the same? E.g.:

```
Warning: channels with different sampling rate not supported, selecting subset of 10 channels at 256.000000 Hz 
```

Edouard2laire:
yes, i get the warning but fieldtrip ends up upsampling them anyway. So I didn't payed to much attention to it

PR #693, opened 2024-03-11
[struct_copy_fields] works with imbricated structure

Edouard2laire:
Hello, 

This PR allows struct_copy_fields to works with imbricated structure. 

Here is an example: 

```matlab

a = struct();
a.ex = struct('x', 1);


b = struct();
b.ex = struct('y', 2);

c = struct_copy_fields(a,b,0); 
% Before : c.ex only contains the field x. 
% After: c.ex contains the fields x and y

```

### DIFF
--- a/toolbox/misc/struct_copy_fields.m
+++ b/toolbox/misc/struct_copy_fields.m
@@ -35,6 +35,12 @@ elseif isempty(sDest)
 else
     namesSrc = fieldnames(sSrc);
     for i = 1:length(namesSrc)
+        % if subfield are both structure, we do a "deep copy"
+        if isfield(sDest, namesSrc{i}) && isstruct(sDest.(namesSrc{i})) && isfield(sSrc, namesSrc{i}) && isstruct(sSrc.(namesSrc{i}))
+             sDest.(namesSrc{i})  = struct_copy_fields(sDest.(namesSrc{i}), sSrc.(namesSrc{i}), override);
+             continue;
+        end
+
         if override || ~isfield(sDest, namesSrc{i})
             sDest.(namesSrc{i}) = sSrc.(namesSrc{i});
         end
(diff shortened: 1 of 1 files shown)


toolbox/misc/struct_copy_fields.m
```diff
@@ -35,6 +35,12 @@
 else
     namesSrc = fieldnames(sSrc);
     for i = 1:length(namesSrc)
+        % if subfield are both structure, we do a "deep copy"
+        if isfield(sDest, namesSrc{i}) && isstruct(sDest.(namesSrc{i})) && isfield(sSrc, namesSrc{i}) && isstruct(sSrc.(namesSrc{i}))
```
rcassani, line 39: We may need to allow the recursive behaviour only for no override `... && ~override`

Otherwise, the override happens only in the deepest level. For example, in the provided code, if override is set to 1, `c.ex` should only contain `y`

Edouard2laire: does it ? 

I would expect following code: 
```matlab
a = struct();
a.ex = struct('x', 1);
b = struct();
b.ex = struct( 'x', 2, 'y', 2);
c = struct_copy_fields(a,b,1); 


```

to produce  c.ex containing the fields x = 2 (overwritten)  and y = 2. 
Maybe we have instead a new flag to the function that if true does the copy of the fields only as suggested here.

Edouard2laire: yes, I would expect c.ex to contain x and y.

Edouard2laire: Would https://github.com/brainstorm-tools/brainstorm3/pull/693/commits/9427abd1fca3dd5a80bcdc0e1138e3da2358e044 works for you ?

rcassani: It seems good, as the current behaviour is not changed

rcassani:
@Edouard2laire, do you want to add something else?

Edouard2laire:
no, that's good for me; if that's ok with you :)

PR #694, opened 2024-03-13
[bugfix] import snirf with event having only one occurance

Edouard2laire:
Fix the following bug happening when importing a snirf where one event  has only one occurrence: 

Maybe its not the best fix. 
```matlab


** Error: Line 258: Index in position 2 exceeds array bounds. Index must not exceed 1.
** 
** Call stack:
** >in_data_snirf.m at 258
** >in_fopen.m at 221
** >import_raw.m at 132
** >bst_call.m at 28
** >tree_callbacks.m>@(h,ev)bst_call(@import_raw,[],[],iSubject) at 672
** 
********

```

Edit: maybe its not the best fix. jnirs.nirs.stim(iEvt).data should be nEvent x 3

### DIFF
--- a/toolbox/io/in_data_snirf.m
+++ b/toolbox/io/in_data_snirf.m
@@ -254,7 +254,7 @@ for iEvt = 1:length(jnirs.nirs.stim)
     if isfield(jnirs.nirs.stim(iEvt), 'dataLabels')
         nStimDataCols = length(jnirs.nirs.stim(iEvt).dataLabels);
     end
-    % Transpose if needed
+    % Transpose to match number of columns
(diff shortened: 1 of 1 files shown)


rcassani:
Based on the [SNIRF specification](https://github.com/fNIRS/snirf/blob/master/snirf_specification.md#nirsistimjdata), a better approach would be to search first for the optional record `/nirs(i)/stim(j)/dataLabels` to decide if transpose or not. If this is not available, transpose could be required in this condition: 

```
if size(data, 1) < 3 && diff(size(data) < 0)
  % Transpose
```

with the hope that if it's 3 by 3, it's correct

rcassani:
@Edouard2laire, I added the code directly in the master branch to avoid double commit (code and version update)

Edouard2laire:
thanks :)

PR #695, opened 2024-03-14
FIX: various fix for itab files

robbisg:
We made few fixes for itab reader:

- header datatype now includes another case (6) both in `in_fopen_itab` and `in_fread_itab`.
- Coil orientation now is normalized.
- Marker positions now include only marker orientation.

### DIFF
--- a/toolbox/io/in_fopen_itab.m
+++ b/toolbox/io/in_fopen_itab.m
@@ -34,9 +34,9 @@ end
 hdr = read_itab_mhd(HeaderFile);
 % Get endianness
 switch (hdr.data_type)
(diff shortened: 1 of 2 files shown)


rcassani:
@robbisg, thanks for the contribution.
For sake of completeness, where can we find information about the ITAB format?

robbisg:
Hi, thanks for accepting the PR.
Actually, our documentation is @vpizzella!! But we can provide the source code of the header to have some information about the format.

rcassani:
@robbisg and @vpizzella, we would really appreciate that documentation for our archives and debugging, thanks!

PR #696, opened 2024-03-25
[bugfix] panel_ieeg: reading coordinates must always happen from MRI Viewer 

chinmaychinara91:
Steps:

1. Right click on CT in anatomy and do SEEG/ECOG Implantation
2. Panel iEEG opens up. Click **+**  to create a new electrode. Give a name and press **OK**. If the current active window is 3D Viz then the error below occurs:
Screenshot 2024-03-25 152829

### DIFF
--- a/toolbox/gui/panel_ieeg.m
+++ b/toolbox/gui/panel_ieeg.m
@@ -1190,11 +1190,7 @@ end
 function [sElectrodes, iDSall, iFigall, hFigall] = GetElectrodes()
     global GlobalData;
     % Get current figure
-    [hFigall,iFigall,iDSall] = bst_figures('GetCurrentFigure');
-    if isempty(hFigall)
-        % Try MRI viewer figure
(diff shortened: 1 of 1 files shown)


toolbox/gui/panel_ieeg.m
```diff
@@ -1190,11 +1190,7 @@ function SetElectrodeVisible(isVisible)
 function [sElectrodes, iDSall, iFigall, hFigall] = GetElectrodes()
     global GlobalData;
     % Get current figure
-    [hFigall,iFigall,iDSall] = bst_figures('GetCurrentFigure');
-    if isempty(hFigall)
-        % Try MRI viewer figure
-        [hFigall,iFigall,iDSall] = bst_figures('GetFiguresByType','MriViewer');
-    end
+    [hFigall,iFigall,iDSall] = bst_figures('GetFiguresByType','MriViewer');
```
chinmaychinara91, line 1193: we don't need to get all the figures. Just the MRI Viewer is enough because for 3D Viz, the user clicks on the surface and the respective point gets updated in the MRI viewer.

PR #697, opened 2024-03-27
[bugfix] tess_isosurface, panel_ieeg

chinmaychinara91:
**ISSUES (tess_isosurface):**
1. In functional view, go to an **implantation node> SEEG/ECOG (with some channels)**
2. Double click to open 3D VIZ and MRI Viewer
3. Use **Thresh** slider to update iso surface 
     **Updated figure is generated in  a new 3D Viz.**
5. Add coordinates panel. Repeat step 3.
     **A bunch of errors pop up**

**SOLUTION:** https://github.com/brainstorm-tools/brainstorm3/pull/697/commits/424a5da6f118561ffa153408e46ee598485a2003

**ISSUES (panel_ieeg):**
When trying to update view type between depth electrode and sphere, the update always occurs in MRI Viewer only. 

**SOLUTION:** https://github.com/brainstorm-tools/brainstorm3/pull/697/commits/ddcefc856efdf3f1c1dbb394dd07bd9abc1af894

### DIFF
--- a/toolbox/anatomy/tess_isosurface.m
+++ b/toolbox/anatomy/tess_isosurface.m
@@ -171,9 +171,7 @@ if isSave
     % Display mesh with 3D orthogonal slices of the default MRI
     MriFile = sSubject.Anatomy(1).FileName;
     hFig = bst_figures('GetFiguresByType', '3DViz');
-    if ~isempty(hFig)
-        hFig = view_mri_3d(MriFile, [], 0.3, hFig);
-    else
+    if isempty(hFig)
         hFig = view_mri_3d(MriFile, [], 0.3, []);
     end
     view_surface(MeshFile, 0.6, [], hFig, []);    
--- a/toolbox/anatomy/tess_isosurface.m
+++ b/toolbox/anatomy/tess_isosurface.m
@@ -170,7 +170,12 @@ if isSave
     iSurface = db_add_surface(iSubject, MeshFile, sMesh.Comment);
     % Display mesh with 3D orthogonal slices of the default MRI
     MriFile = sSubject.Anatomy(1).FileName;
-    hFig = view_mri_3d(MriFile, [], 0.3, []);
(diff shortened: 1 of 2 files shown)


toolbox/anatomy/tess_isosurface.m
```diff
@@ -171,9 +171,7 @@
     % Display mesh with 3D orthogonal slices of the default MRI
     MriFile = sSubject.Anatomy(1).FileName;
     hFig = bst_figures('GetFiguresByType', '3DViz');
-    if ~isempty(hFig)
-        hFig = view_mri_3d(MriFile, [], 0.3, hFig);
-    else
+    if isempty(hFig)
```
chinmaychinara91, line 174: this check now assures that all updates happen in the same 3D Viz figure

toolbox/gui/panel_ieeg.m
```diff
@@ -1979,7 +1979,7 @@ function UpdateFigures(hFigTarget)
 %% ===== SET DISPLAY MODE =====
 function SetDisplayMode(DisplayMode)
     % Get current figure
-    [sElectrodes, iDS, iFig, hFig] = GetElectrodes();
```
chinmaychinara91, line 1982: After yesterday's merge https://github.com/brainstorm-tools/brainstorm3/commit/cc7513f67fc150ad2db5098a13e70a7558cbd72c, this was affected. But `sElectrodes, iDS, iFig` are never used here and nor is this function used as any dependency anywhere outside panel_ieeg. Essentially the idea is to get the current figure handle.

toolbox/gui/panel_ieeg.m
```diff
@@ -1979,7 +1979,7 @@ function UpdateFigures(hFigTarget)
...
+    hFig = bst_figures('GetCurrentFigure');
```
chinmaychinara91, line 1982: This change should do the task.

PR #698, opened 2024-04-12
Some bugfixes from UTAH workshop

chinmaychinara91:


toolbox/gui/panel_ieeg.m
```diff
@@ -338,6 +338,11 @@ function ElecListClick_Callback(h, ev)
     function ContListClick_Callback(h, ev)
         % IF SINGLE CLICK
         if (ev.getClickCount() == 1)
+            ctrl = bst_get('PanelControls', 'iEEG');
+            % if contact list is empty then dont't proceed
+            if ctrl.jListCont.isSelectionEmpty()
```
chinmaychinara91, line 343: [bugfix] If the contact list was not rendered and and user clicks on the panel then the below error occured:
Screenshot 2024-04-11 120435

toolbox/gui/panel_coordinates.m
```diff
@@ -179,8 +179,9 @@ function UpdatePanel()
     if isempty(ctrl)
         return
     end
+
     % Get current figure
-    hFig = bst_figures('GetFiguresByType', '3DViz');
+    hFig = bst_figures('GetCurrentFigure', '3D');
```
chinmaychinara91, line 184: [bugfix] Opening multiple 3D Figures when panel coordinates in open:
1. Go to functional tab
2. Open panel coordinates
3. Try opening multiple 3D figures. The 1st fig opens fine but opening any further after that it throws error.

PR #699, opened 2024-04-17
Update Panel iEEG

chinmaychinara91:
User can now select coordinates from 3D Viz by choosing the **Select** button within **panel ieeg** without having to open **panel coordinates**.


rcassani:
Some general comments:

1. As far as I understand, the aim is to have a shortcut of the "+Select" button. If that is so, it behaves quite different, selecting a point in a cortical surface.

    a. Open a cortex surface, open Coordinates panel, click on +Select in this panel, and click on surface:
    Cursor appears on the surface.
  
    b. Open a cortex surface, open iEEG panel, click on +Select in this panel, and click on surface:Cursor appears on the surface.


3. GUI: the first row of buttons was already busy, the new "+Select" button makes it longer and the "Contact" button may go over the edge. Image below, see also screen captures of the iEEG panel in the [ECoG/sEEG](https://neuroimage.usc.edu/brainstorm/Tutorials/ECoG) and 
[SEEG epileptogenicity maps](https://neuroimage.usc.edu/brainstorm/Tutorials/Epileptogenicity) tutorials.

chinmaychinara91:
Thank you for the inputs. Let me work on them.

rcassani:
Also, there is the bug with closing figures:
1. Open MRI in MRI Viewer
2. Open cortex as 3D figure
3. Close MRI viewer figure
4. Both figures are closed, this is not the desired behaviour

chinmaychinara91:
> Also, there is the bug with closing figures:
> 
> 1. Open MRI in MRI Viewer
> 2. Open cortex as 3D figure
> 3. Close MRI viewer figure
> 4. Both figures are closed, this is not the desired behaviour

I verified this with an [old commit](https://github.com/brainstorm-tools/brainstorm3/commit/c747a70c459fca0ad6b76ba48328bd0ff2cc549b) at random and it seems to have been there like this quite some time. Will debug it.

chinmaychinara91:
It is ready for review.

rcassani:
Commit [4ef79ec](https://github.com/brainstorm-tools/brainstorm3/pull/699/commits/4ef79ecdcd6900aeae04c0cb878760fe9f62dc54) aims for [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). Repeating code makes it hard to maintain. In addition both selectors are now in sync.

toolbox/gui/figure_mri.m
```diff
@@ -883,6 +883,9 @@ function ButtonView3DHead_Callback(hFig)
         Handles.hView3DHeadFig = view_surface_matrix(Vertices, Faces, SurfAlpha);
         % Start coordinates selection
         setappdata(Handles.hView3DHeadFig, 'isSelectingCoordinates', 1);
+        if gui_brainstorm('isTabVisible', 'iEEG') 
+            setappdata(Handles.hView3DHeadFig, 'isSelectingCoordinates', 1);
+        end
```
rcassani, line 888: This is only used when the MRI is being edited and the scalp surface is computed from MRI. Why do we need to have a case for iEEG?

toolbox/gui/figure_3d.m
```diff
@@ -1085,6 +1093,11 @@ function FigureKeyPressedCallback(hFig, keyEvent)
                     if ismember('control', keyEvent.Modifier)
                     	ViewAxis(hFig);
                     end 
+                % C : Enable / disable Select coordinates mode
+                case 'c'
+                    gui_brainstorm('ShowToolTab', 'Coordinates');
+                    pause(0.01);
```
rcassani, line 1099: This pause is needed on my side to display correctly the cross cursor. Please check if that is the case in your side.

toolbox/gui/figure_3d.m
```diff
@@ -1085,6 +1093,11 @@ function FigureKeyPressedCallback(hFig, keyEvent)
                     if ismember('control', keyEvent.Modifier)
                     	ViewAxis(hFig);
                     end 
+                % C : Enable / disable Select coordinates mode
+                case 'c'
+                    gui_brainstorm('ShowToolTab', 'Coordinates');
```
rcassani, line 1098: The shortcut `C` also opens the Coordinates panel, so it does the same as the option "Get coordinates..." in the popmenu. What do you think?

toolbox/gui/figure_3d.m
```diff
@@ -1085,6 +1093,11 @@ function FigureKeyPressedCallback(hFig, keyEvent)
                     if ismember('control', keyEvent.Modifier)
                     	ViewAxis(hFig);
                     end 
+                % C : Enable / disable Select coordinates mode
+                case 'c'
```
rcassani, line 1097: Is `C` as shortcut good for you? 
What about `Ctrl+P` as per Point?

toolbox/gui/figure_3d.m
chinmaychinara91: Yes it is the case in my side as well. But as you suggested below we can replace it by Ctrl+P. It is better if we keep them separate. 
'c' for Coordinates Panel
'Ctrl+P' for enabling/disabling crosshair

toolbox/gui/figure_3d.m
chinmaychinara91: looks good

toolbox/gui/figure_3d.m
chinmaychinara91: agreed. let's do ctrl+P

chinmaychinara91:
> Commit [4ef79ec](https://github.com/brainstorm-tools/brainstorm3/pull/699/commits/4ef79ecdcd6900aeae04c0cb878760fe9f62dc54) aims for [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). Repeating code makes it hard to maintain. In addition both selectors are now in sync.

Thank you for this info.

toolbox/gui/figure_mri.m
rcassani: Sure, we can removed. My question was why it was there in the first place.

toolbox/gui/figure_3d.m
rcassani: What would be the objective to keep them separated?

In [a8390a4](https://github.com/brainstorm-tools/brainstorm3/pull/699/commits/a8390a4299b7e2b44d2bdb72a9f3a93b97d828d9), if the iEEG panel is shown, although the Coordinates panels is open, the iEEG remains the shown panel.

toolbox/gui/figure_mri.m
chinmaychinara91: It was my mistake I did not understand the use case correctly before when I added it.

toolbox/gui/figure_3d.m
chinmaychinara91: Yes agreed. This is a much better way to handle it. 

However, I think you need to add the pause statement before `panel_coordinates('SetSelectionState', ~panel_coordinates('GetSelectionState'));` because without that the cursor issue you posted above remains.

chinmaychinara91:
Missed this use case where the user might have other surface types added for a Subject and theywould want to see the electrodes on them as well using the **Display Sensors** option as mentioned in the forum post above (and not just on 3D MRI Slices).

https://github.com/brainstorm-tools/brainstorm3/pull/699/commits/60cc1273f879314276f61f2edae683b4689eeb22 is a suggestion for issue https://neuroimage.usc.edu/forums/t/ieeg-panel-bugs/45884.

toolbox/gui/view_channels_3d.m
```diff
@@ -142,13 +142,16 @@
                 SurfAlpha = .1;
                 hFig = view_mri_3d(SurfaceFile, [], SurfAlpha, hFig);
                 if strcmpi(Modality, 'SEEG')
-                    % For SEEG, display 3D slices (MRI) + isosurface, and MRI viewer
+                    % For SEEG, display 3D slices (MRI) + isosurface
                     panel_ieeg('DisplayIsosurface', sSubject, hFig);
```
chinmaychinara91, line 146: isosurface makes sense only with 3D MRI slices as for all other surface types it will come in the way while we try accessing the isosurface for any kind of point selection.

toolbox/gui/figure_3d.m
rcassani: Pause is added back in [602ff60](https://github.com/brainstorm-tools/brainstorm3/pull/699/commits/602ff606dd7fe1a8bf2afb1772c25bb95314542a). 

Interestingly, adding the extra code to check the panels existence was enough delay to not needing the pause any more in my setup.

toolbox/gui/panel_ieeg.m
```diff
@@ -1203,7 +1203,7 @@ function SetElectrodeVisible(isVisible)
     [hFigall,iFigall,iDSall] = bst_figures('GetFiguresByType','MriViewer');
 
     % Check if there are electrodes defined for this file
-    if isempty(hFigall) || isempty(GlobalData.DataSet(iDSall).IntraElectrodes) || isempty(GlobalData.DataSet(iDSall).ChannelFile)
+    if isempty(hFigall(end)) || isempty(GlobalData.DataSet(iDSall(end)).IntraElectrodes) || isempty(GlobalData.DataSet(iDSall(end)).ChannelFile)
```
chinmaychinara91, line 1206: This handles the error issue mentioned above

toolbox/gui/panel_ieeg.m
```diff
@@ -3162,6 +3162,11 @@ function CreateNewImplantation(MriFile) %#ok<DEFNU>
         end
     end
 
+    % If MRI Viewer is open don't open another one
+    hFig = bst_figures('GetFiguresByType', 'MriViewer');
```
chinmaychinara91, line 3166: this assures that only one MRI Viewer is open

chinmaychinara91:
good for review now.

rcassani (approved):

rcassani:
The PR looks good to be merged. 

Note: if the PR has the legend "This branch has no conflicts with the base branch" there is no need to synch the branch to the master (as it was made in [40ba04f](https://github.com/brainstorm-tools/brainstorm3/pull/699/commits/40ba04f24a094e9b7bcaa701c4132034b9b6965c)). That just adds extra commits and noise at reading the PR.

chinmaychinara91:
Thank you @rcassani

PR #701, opened 2024-05-01
[bugfix] panel_ieeg

chinmaychinara91:
**Issue-1:** If implantation is started and there is a 3DMRI+isosurface figure open, unexpected things happens.
Solution: https://github.com/brainstorm-tools/brainstorm3/pull/701/commits/fad6568ee413e061af912e16c8270a3d836fdab6

**Issue-2:** Double-clicking on the SEEG/ECOG ("number of channels") channel file in the functional tab only opens the cortex figure for ECOG and ECOG+SEEG.
Solution: https://github.com/brainstorm-tools/brainstorm3/pull/701/commits/9452f2f21fc3685b2ff60ae9705360b7246bc8cc

**Discussion:** https://neuroimage.usc.edu/forums/t/display-sensors-cortex-opens-mri-viewer-in-addition-to-cortex-view/46083/2
Do we need to have MRI Viewer display for all surface types by default ?
If no then this is the proposed solution: https://github.com/brainstorm-tools/brainstorm3/pull/701/commits/3ccd367251ad92d75fde12d652c21427ebd45c0e

**Issue-3:** Reported in https://neuroimage.usc.edu/forums/t/ieeg-panel-error/46550.
Solution: [30e5648](https://github.com/brainstorm-tools/brainstorm3/pull/701/commits/30e5648a07c51288588b0cd1ac0798cd16f7e531)

### DIFF
--- a/toolbox/gui/panel_ieeg.m
+++ b/toolbox/gui/panel_ieeg.m
@@ -3054,6 +3054,9 @@ end
 
 %% ===== CREATE NEW IMPLANTATION =====
 function CreateNewImplantation(MriFile) %#ok<DEFNU>
+    % Unload all figures and datasets before starting a new implantation
+    bst_memory('UnloadAll', 'Forced');
+    
     % Find subject
     [sSubject,iSubject,iAnatomy] = bst_get('MriFile', MriFile);
     % Get study for the new channel file
--- a/toolbox/gui/panel_ieeg.m
+++ b/toolbox/gui/panel_ieeg.m
@@ -3054,9 +3054,21 @@ end
 
 %% ===== CREATE NEW IMPLANTATION =====
 function CreateNewImplantation(MriFile) %#ok<DEFNU>
-    % Unload all figures and datasets before starting a new implantation
-    bst_memory('UnloadAll', 'Forced');
+    % Get all open figures
+    hFig = bst_figures('GetAllFigures');
     
+    % If figures are open, inform the users that to continue all figures must be closed
+    if ~isempty(hFig)
+        isCloseAllFigures = java_dialog('confirm', ['All figures will be closed. Do you want to continue?' 10 ...
+                                        '<HTML>Choose <B>No</B> to revisit the figures before starting implantation</HTML>'], 'SEEG/ECOG implantation');
+        if ~isCloseAllFigures
+            return
+        end
+
+        % Unload all figures and datasets before starting a new implantation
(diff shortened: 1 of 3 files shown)


rcassani:
> Discussion: https://neuroimage.usc.edu/forums/t/display-sensors-cortex-opens-mri-viewer-in-addition-to-cortex-view/46083/2
Do we need to have MRI Viewer display for all surface types by default ?
If no then this is the proposed solution: https://github.com/brainstorm-tools/brainstorm3/commit/3ccd367251ad92d75fde12d652c21427ebd45c0e

This was introduced in #699 with the commit 60cc127 to patch another bug.

[https://github.com/brainstorm-tools/brainstorm3/pull/699/commits/60cc1273f879314276f61f2edae683b4689eeb22](https://github.com/brainstorm-tools/brainstorm3/pull/699/commits/60cc1273f879314276f61f2edae683b4689eeb22)

toolbox/gui/panel_ieeg.m
rcassani, line 1: @chinmaychinara91, could you explain to me what is the reasoning of this proposed change?

---
There is not need to use `cell2mat` for this trouble.
The problem lies on the fact that `sContact` is an array of `[3, nContacts]`, and the indexation in L521 is done with `length` which "returns the length of the largest array dimension" in this case 3, rather than 2.

https://www.mathworks.com/help/matlab/ref/length.html

chinmaychinara91: As you can see here`L1265 sContacts = [sContacts, {ChannelData(i).Loc}]`, the coordinates are stacked as individual cells and then used `cell2mat` to convert it back to matrix where necessary. On doing so the `length(sContacts)` then gave the correct result of 2 rather than 3. 

I guess I took an unnecessary lengthy approach to solve a simple problem. 

A simpler solution would be remove any cell conversion,  to use `size` in place of  `length` to capture each of the dimensions `[3, nContacts]` and then use `nContacts` to loop.

rcassani: > As you can see hereL1265 sContacts = [sContacts, {ChannelData(i).Loc}], the coordinates are stacked as individual cells and then used cell2mat to convert it back to matrix where necessary. 

Of course, because it was added in the same commit. So they were saved in cells to never be used as cells, just to be converted to matrices. Does it sound logic? It does not make sense. It seems there is some lack of understanding of the real problem.

We will work it out in the other PR

PR #706, opened 2024-05-13
Contactsheets for 3D-slices behave as 3D-surfaces

rcassani:
This PR has as goal to have the same functionality in ContactSheets for 3D figures: MRI slices and surfaces. The difference was reported in this Forum thread: 

https://neuroimage.usc.edu/forums/t/45757/10

With these changes, contact sheets from 3D MRI slices will now have:

* Time stamp
* Colorbar 

These two elements will be hidden (as the current behaviour) if and only if there is not data (sources and TF) plotted on the MRI slices, AND volume contact sheets are requested.

For MRI slices: contactsheet `Time` with `Data` (source values)
| Before      | With this PR |
| ----------- | ----------- |
| img_1_1 | img_1_2 |

For MRI slices: contactsheet `Volume` without Data
| Before      | With this PR |
| ----------- | ----------- |
| img_2_1 | img_2_2 |

EDIT: Images have been updated to current state of PR

### DIFF
--- a/toolbox/gui/view_contactsheet.m
+++ b/toolbox/gui/view_contactsheet.m
@@ -77,9 +77,9 @@ ContactSheetOptions = bst_get('ContactSheetOptions');
 % Get dimension
 switch lower(orientation)
     case 'fig',  dim = 0;
-    case 'x',    dim = 1;
-    case 'y',    dim = 2;
-    case 'z',    dim = 3;
+    case 'x',    dim = 1; % Sagittal
+    case 'y',    dim = 2; % Coronal
+    case 'z',    dim = 3; % Axial
 end
 % Time / volume / freq
 switch lower(inctype)
@@ -133,7 +133,7 @@ switch lower(inctype)
         Samples = TimeVector(bst_closest(linspace(TimeRange(1), TimeRange(2), nImages), TimeVector));
         % If reading MRI slices: need to get the surface definition
         if (dim ~= 0)
-            [img, TessInfo, iTess] = GetImage(hFig, dim);
+            [~, TessInfo, iTess] = GetImage(hFig);
         end
         % Do not use progress bar
         isProgress = 0;
@@ -157,7 +157,7 @@ switch lower(inctype)
         initPos = GlobalData.UserFrequencies.iCurrentFreq;
         % If reading MRI slices: need to get the surface definition
         if (dim ~= 0)
-            [img, TessInfo, iTess] = GetImage(hFig, dim);
+            [~, TessInfo, iTess] = GetImage(hFig);
         end
         % Do not use progress bar
         isProgress = 0;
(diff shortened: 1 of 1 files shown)


rcassani:
@tmedani, @chinmaychinara91 could you provide some feedback on this PR. 
It does look good to me.

chinmaychinara91:
Open Default Anatomy MRI > Right click > Snapshot > Choose any of the **View contact sheet** and choose the defaults from **Contact sheet GUI**. Getting the error below:
```Error using view
When specified as a graphics object, first argument must be one or more axes objects.

Error in view_contactsheet (line 221)
    [az,el] = view(hAxes);

Error in figure_mri>@(h,ev)view_contactsheet(hFig,'volume','y',DefaultOutputDir) (line 1103)
        gui_component('MenuItem', jMenuSave, [], 'Volume contact sheet: Coronal',  IconLoader.ICON_CONTACTSHEET, [], @(h,ev)view_contactsheet(hFig, 'volume', 'y', DefaultOutputDir));
```

tmedani:
all looks good to me, 
thanks Ray

tmedani:
One comment: is it better to have just one color bar?

rcassani:
@tmedani, it is possible that the maximum of the colormap is set to `local` then it will be different across contactsheets.

@chinmaychinara91, commit [94f5313](https://github.com/brainstorm-tools/brainstorm3/pull/706/commits/94f5313e71350f3f2a26016597fbaab127128520) handles the case you found. How about now?

tmedani:
> @tmedani, it is possible that the maximum of the colormap is set to `local` then it will be different across contactsheets.

I see, so it is then easier to tune each figure locally.

The thing that bothers me is the size of the figure, which is much smaller.

chinmaychinara91:
@rcassani looks good now.

tmedani:
Awesome :)

rcassani:
Ok, this is now ready.

One of the changes, is that all contactsheets are obtained from 3D figures, even when initiated from the MRI viewer. This is to handle the time legend and colorbar.

Also the removal of the useless black background is now done across all the contact images, so the final composite image can be cut into equal size pieces, and all the pieces have the same spatial scale. This was not possible before as the removal was different for each (contact) image.

toolbox/gui/view_contactsheet.m
```diff
@@ -278,29 +324,55 @@
...
-    background = double(AlphaSheet == 0);
-    % If there are no transparent points on the surface: detect "black" points
-    if (nnz(background) == 0)
-        background = double(sqrt(sum(double(ImgSheet).^2,3)) < .05);
-    end
+    % Detect "black" points for all images as background
+    background = all(double(sqrt(sum(double(ImgBuffer).^2,3)) < .05), 4);
     % Grow background region, to remove all the small parasites
-    kernel = ones(5,5);
+    kernel = ones(2,2);
```
rcassani, line 352: Less aggressive removal of small non-black. 
This to avoid removing time legend, and scale in the colorbar

PR #708, opened 2024-05-27
Add exclusion zones in surface and volume grids for SEEG source estimation

rcassani:


rcassani:
Currently handle in #723

PR #710, opened 2024-05-29
Refactor updated IEEG features (1/2)

rcassani:
# Refactor updated IEEG features
This PR aims to refactor the IEEG new features introduced in the main PR: #689.

Through the in-house testing, tutorials and the forum, we have become aware of some issues with the new IEEG features. Some of these issues have been hot-fixed in subsequent already merged PRs: #696, #697, #698, #699, #700, and additional hot-fixes were under development in #701 (closed without merging). 

## Solved issues
- [x] Error for electrodes with less than 3 contacts. Solved at 30146a4 and 52898a5
 https://neuroimage.usc.edu/forums/t/46550
- [x] Implantation started from anatomy view should be located at the ~~`@intra`~~ "Implantation" folder. If it exist, allow the user to either 'continue', or to 'replace' it.
  Implemented in 20d3c34 and 96bb08c
- [x] Behaviour of double-click on channel file cannot be achieved with any option in the popup menu
 It corresponds to "SEEG/ECOG implantation" on the Channel file popupmenu: 3543fff
- [x] Coordinate radio button for contact coordinates is not kept when selecting different electrodes. Solved in 837ea1a
- [x] Incorrect behaviour selecting contacts. If multiple selection with keyboard (shift+arrows, or Ctrl+A), selected contacts are not highlighted in 3DFig. Solved at d39d556
- [x] Incorrect behaviour selecting contacts. Selected contact remains: (1) select one contact for one electrode (contact get highlighted in 3DFig), (2) change electrode (contact remain highlighted), (3) return to first electrode (no contact is selected, but the contact for step (1) remains highlighted). Solved in 63370a5
- [x] ~~Display sensors in MRI viewer. Load isosurface although it is not used~~ 
  Used by `ieeg('CreateGeometry3DElectrode')` Loads surface (innerskull), to obtain normals, to orient ECOG cylinders, visible as cylinders only when MIP: Functional
- [x] ~~Display sensors in MRI 3D slices, Load isosurface, but not added to display~~ 
  Used by `ieeg('CreateGeometry3DElectrode')` Loads surface (innerskull), to obtain normals, to orient ECOG cylinders, visible as cylinders when display contacts (instead of spheres) is selected in IEEG panel

## Pending issues (for other PR)

- [ ] Display menu (for channel file with SEEG) does something different to the stated
 https://neuroimage.usc.edu/forums/t/46083
- [ ] Avoid closing all figures when starting a new implantation. If figures need to be closed, this must be selective:
[...]

PR #713, opened 2024-06-22
support SEEG with one contact

Edouard2laire:
Hello, 

This PR fixes some issues when there is only one SEEG contact (doesn't usually happen but can happen with some dataset shared online)


rcassani:
Hi @Edouard2laire, we (@chinmaychinara91 and I) are currently working on this other PR #710 to address some issues with the recent update of the iEEG functionalities. The issue fixed in this PR is (will be) addressed in this PR in #710.

PR #716, opened 2024-06-28
Automatic EEG localization and labelling using Revopoint 3D

chinmaychinara91:
**Tutorial:** https://neuroimage.usc.edu/brainstorm/Tutorials/TutDigitize3dScanner#Digitize_using_Revopoint_Digitize_GUI_in_Brainstorm

**Test Data:** Can be downloaded from [here](https://drive.google.com/drive/folders/1n9IlcMljsEiOjcED6KRxKH3k4nMT8WAN?usp=sharing).

Goal is to allow automatic localization and labelling of EEG caps using 3D scanned mesh of a person wearing the cap (captured using [Revopoint 3D](https://www.revopoint3d.com/products/handheld-3d-scanner-range)). This represents a new low-cost, fast, easy-to-use approach for digitizing heads and electrodes as compared to the traditional [digitizer](https://neuroimage.usc.edu/brainstorm/Tutorials/TutDigitize).

**NOTE: Currently tested well with two configurations of EEG caps: ~~https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/d40c719ede6ee167c290c7b526836d57811761ca~~**  https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/c49ebc6cc474979bfd634731f5be67dcca33e6bc

**Proposed GUI within Brainstorm**
**LEGACY MODE**
Screenshot 2024-09-27 113338

**2024 MODE**
Screenshot 2024-09-27 113125

- [x] Load OBJ file into brainstorm (mesh with vertices, faces and color)
- [x] Add color field to brainstorm surface structure 
- [x] Color should be updated when mesh downsampled in BST as well
- [x] Refactor the Digitize menu to add Revopoint
- [x] Refactor the Digitizer GUI for Revopoint tasks
- [x] Menu to load default available caps in BST as Montages in Revopoint GUI
- [x] Handle surface color update for  **Views** menu option on 3D figure (Clone figure, etc.)
- Test nothing breaks for the already available Digitizer functionalities:
   - [x] Simulation mode 
   - [x] With Polhemus connected
- [x] **19th Aug 2024:** Make it work on both the "legacy" and 2024 Digitize panel (https://github.com/brainstorm-tools/brainstorm3/pull/682)

@tmedani @ajoshiusc @yashvakilna


toolbox/db/db_template.m
rcassani, line 1: Do we want to add this field to the template?
Can we save on the `.mat` file and only read it when needed?

toolbox/io/import_surfaces.m
```diff
@@ -204,7 +207,11 @@
     NewTess = bst_history('add', NewTess, 'import', ['Import from: ' TessFile]);
     % Produce a default surface filename (surface of volume mesh)
     if isfield(NewTess, 'Faces')
-        BstTessFile = bst_fullfile(ProtocolInfo.SUBJECTS, subjectSubDir, ['tess_' importedBaseName '.mat']);
+        if isfield(NewTess, 'Color')
```
rcassani, line 210: Since `NewTess` is created with `db_template('surfacemat')` it will always have the `Color` field, thus all surfaces are imported as `tess_textured_*.mat`

rcassani (changes requested):
Please test exhaustively against legacy code in Brainstorm: tutorials in GUI and tutorial scripts.

The proposed changes are not compatible with what is currently in Brainstorm. They produce errors in fundamental functions such as `import_anatomy_fs.m` and potentially all the others `import_anatomy_*.m` scripts since all surfaces have `Color` field:

---
Error stack:
```
Line 159: Index in position 1 exceeds array bounds.

Call stack: 
>tess_downsize.m at 159 
>import_anatomy_fs.m at 312 
>process_import_anatomy.m>Run at 182 
>process_import_anatomy.m at 24 
>bst_process.m>Run at 236 
>bst_process.m>CallProcess at 2316 
>bst_process.m at 38
```

chinmaychinara91:
> tutorials in GUI and tutorial scripts

Can you point me to this ?

toolbox/db/db_template.m
chinmaychinara91: I think yes it will be better to save it to the mat file rather than the template. Let me work on that.

rcassani:
> Can you point me to this ?


@chinmaychinara91 

### 1. Running tutorials on GUI
There are many tutorials in this page, that you can follow to verify the changes are backwards compatible
https://neuroimage.usc.edu/brainstorm/Tutorials/

### 2. Running tutorials with scripts
Most tutorials can be run with scripts, located in main Brainstorm repository. Look for the `tutorial_*.m` scripts:
https://github.com/brainstorm-tools/brainstorm3/tree/master/toolbox/script

There is also this private repo: https://github.com/brainstorm-tools/bst-tests

toolbox/db/db_template.m
chinmaychinara91: check [ced72dd](https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/ced72ddafea0acc67a0b75d9752f600df3daf32d)

toolbox/io/import_surfaces.m
chinmaychinara91: This should now work fine after [d7fbfcc](https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/d7fbfcc2dd41749cffc031a3b60db71ad932e9f1) [ced72dd](https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/ced72ddafea0acc67a0b75d9752f600df3daf32d)

rcassani:
Hey guys, unfortunately, we do not have a revopoint yet. 
Can you provide the following to tests the code in the PR?

* Some raw files to test the import of the new surface format
* A protocol where the sensor localization has been already performed

In addition, have you had the chance to run the digitization with Polhemus, to see that everything is fine with that?

chinmaychinara91:
> Hey guys, unfortunately, we do not have a revopoint yet. Can you provide the following to tests the code in the PR?
> 
> * Some raw files to test the import of the new surface format
> * A protocol where the sensor localization has been already performed
> 
> In addition, have you had the chance to run the digitization with Polhemus, to see that everything is fine with that?

@rcassani: 

1. Based on the tutorial in the description above, raw test data is attached [here] (https://drive.google.com/drive/folders/1n9IlcMljsEiOjcED6KRxKH3k4nMT8WAN?usp=sharing) as **sub1_Waveguard_ANT_65.zip**
2. The test protocol can also be found in the link above as **TutorialDigitizer.zip**

I ran it in the simulated mode without the Polhemus connected. Functionalities looked good and nothing broke. Polhemus here at USC has gone for some repair. Once it is back I will check it.

Also ran the exhaustive testing with various tutorials as you instructed above. All looked good.

tmedani:
> In addition, have you had the chance to run the digitization with Polhemus, to see that everything is fine with that?

we can test that at USC, we have the digitizer around.

but if I understand it, this will not be an issue since the Digitizer simulation was working fine.

rcassani:
@chinmaychinara91, thank you for the data

> but if I understand it, this will not be an issue since the Digitizer simulation was working fine.

Correct, it should not be an issue. Though we must test it before merging the PR

chinmaychinara91:
> @chinmaychinara91, thank you for the data
> 
> > but if I understand it, this will not be an issue since the Digitizer simulation was working fine.
> 
> Correct, it should not be an issue. Though we must test it before merging the PR

Yes definitely. We will test it here and let you know.

tmedani:
> > @chinmaychinara91, thank you for the data
> > > but if I understand it, this will not be an issue since the Digitizer simulation was working fine.
> > 
> > 
> > Correct, it should not be an issue. Though we must test it before merging the PR
> 
> Yes definitely. We will test it here and let you know.

It may require some time to do that on the digitizer itself. 
we need to set it up and run it.

rcassani:
@Moo-Marc, can you give us a hand with the test with the Polhemus?

chinmaychinara91:
> > > @chinmaychinara91, thank you for the data
> > > > but if I understand it, this will not be an issue since the Digitizer simulation was working fine.
> > > 
> > > 
> > > Correct, it should not be an issue. Though we must test it before merging the PR
> > 
> > 
> > Yes definitely. We will test it here and let you know.
> 
> It may require some time to do that on the digitizer itself. we need to set it up and run it.

chinmaychinara91:
> It may require some time to do that on the digitizer itself.
> we need to set it up and run it.

Will be testing the PR with Polhemus today. Will keep posted.

chinmaychinara91:
The Polhemus here at USC is being but of a task as it just got back from repairs. @Moo-Marc could you help us in testing this PR out at your end ? Just want to make sure already available functionalities do not break.

chinmaychinara91:
@Moo-Marc thank you so much. Do let me know once you test it out.

chinmaychinara91:
[d753abd](https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/d753abdd514b0d77c69aa1963ff10849dc2bd861) commit takes into account that the New Digitize Panel changes (https://github.com/brainstorm-tools/brainstorm3/pull/682) are merged to this PR.

chinmaychinara91:
@rcassani this is ready for review. Both the legacy and new panel support Revopoint now. Do test it out on your end based on the tutorial in description and let me know if you face any issues.

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -1216,6 +1705,198 @@ function BytesAvailable_Callback(h, ev)
...
+    
+        if kk<NIT/2
+            centerssketch(:,1) = 0.9*centerssketch(:,1) + 0.1*xsR;
+            centerssketch(:,2) = 0.9*centerssketch(:,2) + 0.1*ysR;
+        else
+            centerssketch(:,1) = xsR;
+            centerssketch(:,2) = ysR;
+        end
+
+        centerssketch = max(min(centerssketch,512-15),15);
```
rcassani, line 1883: These hard-coded values, where are they coming from?

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -1216,6 +1705,198 @@ function BytesAvailable_Callback(h, ev)
...
+    
+        for i=1:4
+            DeletePoint_Callback();
+        end
+    
+        [Ozx, Ozy] = bst_project_2d(EegPoints(1,1), EegPoints(1,2), EegPoints(1,3), '2dcap');
+        [T8x, T8y] = bst_project_2d(EegPoints(2,1), EegPoints(2,2), EegPoints(2,3), '2dcap');
+        [GNDx, GNDy] = bst_project_2d(EegPoints(3,1), EegPoints(3,2), EegPoints(3,3), '2dcap');
+        [T7x, T7y] = bst_project_2d(EegPoints(4,1), EegPoints(4,2), EegPoints(4,3), '2dcap');
+        cap_pts = ([Ozx,Ozy;T8x,T8y;GNDx,GNDy;T7x,T7y]+1)*256;
```
rcassani, line 1836: This is inefficient and unnecessary.
Only one call to `bst_project_2d` is needed

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -1216,6 +1705,198 @@ function BytesAvailable_Callback(h, ev)
...
+        centerssketch = [centerssketch; centerssketch_temp(find(cellfun(@(c)strcmpi(c, 'T7'), {ChannelRef.Name})),:)];
+
+        for i=1:nEEG
+            if ~strcmpi(ChannelRef(i).Name, 'Oz') &&...
+               ~strcmpi(ChannelRef(i).Name, 'T8') &&...
+               ~strcmpi(ChannelRef(i).Name, 'Fpz') &&...
+               ~strcmpi(ChannelRef(i).Name, 'T7')
+                centerssketch = [centerssketch; centerssketch_temp(i, :)];
+            end
+        end
```
rcassani, line 1777: This just sorts the coordinates, use indexing instead of doing unnecessary FOR loops

rcassani (commented):
@chinmaychinara91, @tmedani 
## General comments

The code works with the provided data, and following the steps in the tutorial. I have pushed commits to improve and fix bugs.

There is not a clear workflow for the case where the subject has its own anatomy. Indeed, the code is hard-coded to use the first available anatomy template.

@chinmaychinara91, I'm tagging you in some of the commits I pushed to indicate why they were done, to use similar approaches to improve your code.

### Panel digitize

* It seems there is no reason to have a different name for the panel
  See commits: c857f3e and 8d6bc3bd

### Automated detection of electrodes. 

* There are many places where values are hard-coded and will not work with `.obj` with different resolutions or with different caps.

* The code in the warpping section is not properly commented. It only works for two caps. It is not scalable. And it relies hundreds of lines copy-pasted lines of code. This makes it hard to review, and will make maintaining and extending a nightmare.

### Other comments
* Most comments in Brainstorm start with capitals, please follow that

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -60,6 +60,11 @@ function Start(DigitizerType)
         'iPoint',           0, ...
         'Transf',           []);
     
+    % Fix old structure (bef 2024) for Digitize.Options.Montages
```
rcassani, line 63: @chinmaychinara91, if adding new things into the well defined structures (in this case `bst_get('DigitizeOptions')`, you MUST support older versiong

toolbox/gui/view_surface_matrix.m
```diff
@@ -186,7 +186,7 @@
 end
 % Register in the GUI
 GlobalData.Surface(end + 1).Color = sLoadedSurf.Color;
-GlobalData.Surface(end + 1) = sLoadedSurf;
+GlobalData.Surface(end) = sLoadedSurf;
```
rcassani: @chinmaychinara91, this caused almost-empty entries in `GlobalData.Surface`, which cause a lot of warning in the Matlab console window when clearing the memory. 

Did you notice it?

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -1869,7 +1869,7 @@ function BytesAvailable_Callback() %#ok<INUSD>
         ind(isoutlier) = [];
         vec_atlas_pts(isoutlier) = [];
     
-        [warp,L,LnInv,bendE] = tpsGetWarp(lambda, centerssketch(vec_atlas_pts,1)', centerssketch(vec_atlas_pts,2)', centerscap(ind,1)', centerscap(ind,2)' );
+        warp = tpsGetWarp(lambda, centerssketch(vec_atlas_pts,1)', centerssketch(vec_atlas_pts,2)', centerscap(ind,1)', centerscap(ind,2)' );
```
rcassani, line 1871: @chinmaychinara91, why to store those variables?

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -153,19 +153,17 @@ function Start(DigitizerType)
     db_reload_studies(Digitize.iStudy);
 
     if strcmpi(Digitize.Type, '3DScanner')
-        % import surface
-        iTargetSurface = find(cellfun(@(x)~isempty(regexp(x, '3dscanner', 'match')), {sSubject.Surface.Comment}));
-        if(isempty(iTargetSurface))
-            [~, OutputSurfacesFiles, ~] = import_surfaces(iSubject);
-            sSurf = bst_memory('LoadSurface', OutputSurfacesFiles{end});
-            sSubject = bst_get('Subject', Digitize.SubjectName);
```
rcassani: @chinmaychinara91 Do not query the DB if there is not a real reason to do it.
When we move to the SQLite version, these calls will be slower

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -542,6 +794,39 @@ function ManualCollect_Callback()
...
+    ctrl.jButtonRandomHeadPts.setEnabled(0);
+
+    hFig = bst_figures('GetCurrentFigure','3D');
+    [~, TessInfo, ~, ~] = panel_surface('GetSurfaceMri', hFig);
+    TessMat.Vertices = double(TessInfo.hPatch.Vertices);
+    TessMat.Faces = double(TessInfo.hPatch.Faces);
+    TessMat.Color = TessInfo.hPatch.FaceVertexCData;
+    dsFactor = 100 / size(TessMat.Vertices, 1); 
+    % Reduce number of vertices
+    [NewTessMat.Faces, NewTessMat.Vertices] = reducepatch(TessMat.Faces, TessMat.Vertices, dsFactor);
```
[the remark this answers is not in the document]
chinmaychinara91: @rcassani what would you recommend here then ? We observed this while testing before as well. One way I am thinking is maybe we can use reduce patch for maybe say 200 points and then out of those just select 100 points by filtering out the points detected in the edges.

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -96,36 +135,59 @@ function Start()
...
+        % Import surface
+        iSurface = find(cellfun(@(x)~isempty(regexp(x, '3dscanner', 'match')), {sSubject.Surface.Comment}));
+        if isempty(iSurface)
+            [~, surfaceFiles] = import_surfaces(iSubject);
+            if isempty(surfaceFiles)
+                return
+            end
+            surfaceFile = surfaceFiles{end};
+        else
+            surfaceFile = sSubject.Surface(iSurface(end)).FileName;
```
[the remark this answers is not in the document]
chinmaychinara91: Good point. We should let the user select. Will make the change.

toolbox/sensors/panel_digitize_2024.m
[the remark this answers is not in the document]
chinmaychinara91: Will have these fixed.

defaults/eeg/Colin27/channel_ANT_Waveguard_65.mat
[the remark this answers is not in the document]
chinmaychinara91: We just added it for testing in Colin27 anatomy. Will work it out with @tmedani to decide on how to distribute.

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -823,47 +1228,96 @@ function SelectMontage(iMontage)
...
+            newMontage.Labels{end+1} = 'T7';
+            for i=1:col
+                if ~strcmpi(ChannelMat.Channel(i).Name, 'Oz') &&...
+                   ~strcmpi(ChannelMat.Channel(i).Name, 'T8') &&...
+                   ~strcmpi(ChannelMat.Channel(i).Name, 'Fpz') &&...
+                   ~strcmpi(ChannelMat.Channel(i).Name, 'T7')
+                    newMontage.Labels{end+1} = ChannelMat.Channel(i).Name;
+                end
+            end
+        end
```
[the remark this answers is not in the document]
chinmaychinara91: This was bugging me as well when I wrote it as well. I will handle it better. Let me work on it.

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -1216,6 +1708,203 @@ function BytesAvailable_Callback(h, ev)
...
+    [curMontage, nEEG] = GetCurrentMontage();
+    if ~isempty(regexp(curMontage.Name, 'ActiCap', 'match'))
+        vc_sq = imcomplement(vc_sq);
+    end
+
+    % toggle comment depending on cap
+    if ~isempty(regexp(curMontage.Name, 'ActiCap', 'match'))
+        [centers, radii, metric] = imfindcircles(vc_sq,[6 55]); % 66 easycap
+    elseif ~isempty(regexp(curMontage.Name, 'Waveguard', 'match'))
+        [centers, radii, metric] = imfindcircles(vc_sq,[1 25]); % 65 ANT waveguard
```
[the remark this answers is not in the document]
chinmaychinara91: We did some testing and faced this issue of the range. I agree on what you say. Since for now we have a very limited set of caps, we need to more rigorous testing to get a good range.

toolbox/io/in_tess_wftobj.m
```diff
@@ -0,0 +1,181 @@
...
+    'Faces',    [], ...
+    'Vertices', [], ...
+    'Color',    [], ...
+    'Comment', '3dscanner head surface');
+
+% Convert face data
+TessMat.Faces = faces;
+
+% Convert vertices data
+TessMat.Vertices = vertices ./ 1000;
```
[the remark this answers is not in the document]
chinmaychinara91: That is correct. @yashvakilna had this in his version and it did not strike it then. This conversion is not required.

toolbox/gui/view_surface_matrix.m
chinmaychinara91: I was getting them as well but could not get the root cause. Thanks for this.

toolbox/sensors/panel_digitize_2024.m
chinmaychinara91: Yuu men the `L,LnInv,bendE` ? They are not required. I forgot to remove them.

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -36,51 +38,88 @@
...
+    end
+
+    [sSubject, iSubject] = bst_get('Subject', Digitize.SubjectName);
+    % Create if subject doesnt exist
+    if isempty(iSubject)
+        % Default anat / one channel file per subject
+        if strcmpi(Digitize.Type, '3DScanner')
+            [sSubject, iSubject] = db_add_subject(Digitize.SubjectName, iSubject);
+            sTemplates = bst_get('AnatomyDefaults');
+            db_set_template(iSubject, sTemplates(1), 1);
```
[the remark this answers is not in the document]
chinmaychinara91: I have confusion here. Doesn't the `Digitizer` use the same default anatomy approach as well ? We just drilled down the same logic for this now. But this is a good point as well. Also, how do we handle things for the `Digitizer` if the subject has a individual anatomy ?

toolbox/sensors/panel_digitize_2024.m
chinmaychinara91: I will double check and get back.

chinmaychinara91:
@rcassani thank you for your comments. I have replied to them individually. Do go through them. But to summarize yes it needs another round of cleanup from my end which I will work on this week. Definitely it is not generalized well and there are some hard coded values but we are yet to receive more samples with variety of caps to get a good sense of generalizing them.

rcassani:
@chinmaychinara91, related to [53f43dd](https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/53f43dde11cbd2ea28e7a2bb0477bdac0fa7e7d8). 

Do not patch code just to make it work. 
You have to understand why there was a problem, and how it can be solved from origin. 

Moreover, remember to keep previous behaviours otherwise: 

* All the calls must be updated. E.g. the proposed changes brakes the calls `brainstorm digitize`
https://github.com/brainstorm-tools/brainstorm3/blob/23b805bea0cecaa09ef41de416fd95b8e65176b1/brainstorm.m#L141

* If user's code rely on this they will need to change their codes.

In addition, it would need to update the usage strings for `panel_digitize` and `panel_digitize_2024`

What about solving the bug from root like below?
This solution is compatible with former ways using 'Start' with `panel_digitize` and `panel_digitize_2024`

```
--- a/toolbox/sensors/panel_digitize.m
+++ b/toolbox/sensors/panel_digitize.m
@@ -40,7 +40,7 @@ end
 %#function icinterface
 
 %% ===== START =====
-function Start(DigitizerType) %#ok<DEFNU>
+function Start(varargin) %#ok<DEFNU>
     global Digitize    
     % ===== INITIALIZE CONNECTION =====
     % Intialize global variable
@@ -68,10 +68,11 @@ function Start(DigitizerType) %#ok<DEFNU>
             'trans',    []));
     
     % ===== PARSE INPUT =====
-    Digitize.Type = 'Digitize';
-    if nargin > 0 && ~isempty(DigitizerType)
-        Digitize.Type = DigitizerType;
+    DigitizerType = 'Digitize';
+    if nargin > 0 && ~isempty(varargin{1})
+        DigitizerType = varargin{1};
     end
+    Digitize.Type = DigitizerType;
     switch DigitizerType
         case 'Digitize'
             % Do nothing
```

Same thing must be done for `panel_digitize_2024`

defaults/eeg/Colin27/channel_ANT_Waveguard_65.mat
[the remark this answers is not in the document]
tmedani: @chinmaychinara91  here

toolbox/gui/view_headpoints.m
chinmaychinara91, line 1: @rcassani This seems to be not working:

Steps for new panel:
1. Start Digitize->Digitizer
2. Click on collect point 3 times to get NAS, LPA, RPA

ERROR IS THROWN. **Unable to find hFig**

rcassani: If digitization was done previously with 3DScanner (`Digitize.Type = '3DScanner'`), right-click on channel file > Digitized head points > View head points failed.

It should be properly addressed now in 70992eeb

chinmaychinara91:
@rcassani 

I have made all the changes based on the comments and discussions above. Let me know if you find any errors on your side.

To summarize:
1. User can now just right click on the imported texture surface > **Digitize (3D Scanner)** to start the digitization.
1. Regarding **There is not a clear workflow for the case where the subject has its own anatomy**, 
the user can right click on the subject > **Import surfaces** and import the textured surface. Then right click on the imported  textured surface > **Digitize (3D Scanner)** to start the digitization.follow step 1 above.
1. User now gets to choose what to do when starting 3D Scanner digitization (Use existing surface or Add a new surface).
1. Better way of sampling without having to use `reducepath`.
1. `auto_3dscanner` now handles all the automatic detection and labelling of 3D Scanner acquired mesh. For new caps, we just need to change the `getEegCapLandmarkLabels` function and that should be it.
1. Did some rigorous testing and was able to come up with a single range for  `imfindcircles` with the caps we have available. Check https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/3403c9ef4359220ca9b189425b14ea901ef3f079.
1. Updated the cap templates for Colin27 and ICBM152. Check https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/c49ebc6cc474979bfd634731f5be67dcca33e6bc.
1. The textured surface in database is updated with `Digitize.Transf` transformation once the fiducials are digitized.
1. Had forgot about this https://forum.revopoint3d.com/t/revopoint-generated-mesh-coordinate-system/25662. The Revopoint vertices are in 'mm'. Did some digging here https://www.fieldtriptoolbox.org/tutorial/electrode/ and I think we need this approach https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/b1124bf6e0fc558fd81ffb303ff36d0c4210859d if we want to support any generic head surface acquired through other devices. 
1. Needed to do some ordering changes in the way the initialization points were captured for scalability. 

Will update the tutorial to match the changes for you to test.

chinmaychinara91:
For importing an OBJ into brainstorm:

1. The in_tess_wftobj function in actually a slightly modified submodule of https://www.fieldtriptoolbox.org/reference/fileio/ft_read_headshape. Should we use the fieldtrip function directly or since the license allows us, just continue to do it as it is right now (and just put a reference to the fieldtrip function).
1. We also use the https://www.fieldtriptoolbox.org/reference/forward/ft_convert_units function to determine the units of the vertices based on the shape of the head. This is a good function to have.

Both these functions are available as part of spm12 plugin (/external/fieldtrip) and also the fieldtrip plugin. How do we go about this ?

rcassani:
In short, it does not makes too much sense to request the user to install FieldTrip as plugin and Load it to perform trivial operations and things that already in the Brainstorm source code. Moreover, not all functions in FieldTrip are available in the compiled version of Brainstorm.

---

> 1. The in_tess_wftobj function in actually a slightly modified submodule of https://www.fieldtriptoolbox.org/reference/fileio/ft_read_headshape. Should we use the fieldtrip function directly or since the license allows us, just continue to do it as it is right now (and just put a reference to the fieldtrip function).

* The waveform .obj files are simple text files that can be parsed quickly, take a look to a Brainstorm function such as [`in_tess_off.m`](https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/io/in_tess_off.m). Do you have any trouble reading this format of file? Do we want to install full FieldTrip for this?



> 2. We also use the https://www.fieldtriptoolbox.org/reference/forward/ft_convert_units function to determine the units of the vertices based on the shape of the head. This is a good function to have.

* Same thing as the point above. Do we want to install full FieldTrip for this?

* A conversion of surface from Brainstorm > FieldTrip > Brainstorm would be needed.

* Choosing the right scaling for distance units is shown in the [Brainstorm tutorials](https://neuroimage.usc.edu/brainstorm/Tutorials/Epileptogenicity#Import_the_contacts_positions). This is done with a function that is already in Brainstorm, [`channel_fixunits`](https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/sensors/channel_fixunits.m)

chinmaychinara91:
@rcassani thanks for pointing out the resources. That makes things much simpler now. I don't think we will need FT now.

toolbox/sensors/auto_3dscanner.m
```diff
@@ -0,0 +1,215 @@
...
+    curMontage = panel_fun('GetCurrentMontage');
+
+    % For white caps change the color space by inverting the colors
+    % NOTE: only 'Acticap' is the tested white cap (needs work on finding a better aprrooach)
+    if ~isempty(regexp(curMontage.Name, 'ActiCap', 'match'))
+        capImg2d = imcomplement(capImg2d);
+    end
+    
+    % Detect the centers of the electrodes which appear as circles in the flattened image whose radii are in the range below
+    capCenters2d = imfindcircles(capImg2d, [minRadius maxRadius]);
```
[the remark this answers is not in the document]
rcassani: Addressed in XXX

toolbox/sensors/auto_3dscanner.m
```diff
@@ -0,0 +1,215 @@
...
+    % Interpolation of the fitted points to the image space of the layout   
+    ll=linspace(-1,1,capImgDim);
+    [X1,Y1]=meshgrid(ll,ll);    
+    capLayoutPts2dU = interp2(X1,xsR,ysR);
+    capLayoutPts2dV = interp2(Y1,xsR,ysR);
+    
+    % Get the desired electrode locations on the 3D EEG cap 
+    capPoints3d(:,1) = griddata(surface3dscannerUv.u, surface3dscannerUv.v, surface3dscannerUv.Vertices(:,1), capLayoutPts2dU, capLayoutPts2dV);
+    capPoints3d(:,2) = griddata(surface3dscannerUv.u, surface3dscannerUv.v, surface3dscannerUv.Vertices(:,2), capLayoutPts2dU, capLayoutPts2dV);
+    capPoints3d(:,3) = griddata(surface3dscannerUv.u, surface3dscannerUv.v, surface3dscannerUv.Vertices(:,3), capLayoutPts2dU, capLayoutPts2dV);
```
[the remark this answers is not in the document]
rcassani: Addressed in XXX

toolbox/sensors/auto_3dscanner.m
```diff
@@ -0,0 +1,215 @@
...
+    % Perform image processing to detect the electrode locations
+    % Convert to grayscale
+    grayness = surface3dscanner.Color*[1;1;1]/sqrt(3);
+    
+    % Interpolate and fit flattended mesh image to a 512x512 grid 
+    % NOTE: Should work with any flattened cap mesh but needs more testing
+    ll=linspace(-1,1,512);
+    [X,Y]=meshgrid(ll,ll);
+    capImg2d = 0*X;
+    capImg2d(:) = griddata(surface3dscannerUv.u(1:end),surface3dscannerUv.v(1:end),grayness,X(:),Y(:),'linear');
```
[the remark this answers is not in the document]
rcassani: Addressed in XXX

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -524,6 +765,68 @@ function UpdateList()
...
+    ctrl = bst_get('PanelControls', 'Digitize');
+    
+    % Get current montage
+    curMontage = GetCurrentMontage();
+
+    % Get EEG cap landmark labels used for initialization
+    [nLandmarkLabels, eegCapLandmarkLabels] = auto_3dscanner('getEegCapLandmarkLabels',curMontage.Name);
+
+    if numel(GlobalData.DataSet(Digitize.iDS).Channel) < nLandmarkLabels
+        bst_error(['Please set the first ' num2str(nLandmarkLabels) ' initialization points in order [ ' sprintf('%s ',eegCapLandmarkLabels{:}) ']'], Digitize.Type, 0);
```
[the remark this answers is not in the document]
rcassani: Addressed in XXX

toolbox/io/import_surfaces.m
```diff
@@ -204,7 +207,11 @@
     NewTess = bst_history('add', NewTess, 'import', ['Import from: ' TessFile]);
     % Produce a default surface filename (surface of volume mesh)
     if isfield(NewTess, 'Faces')
-        BstTessFile = bst_fullfile(ProtocolInfo.SUBJECTS, subjectSubDir, ['tess_' importedBaseName '.mat']);
+        if isempty(NewTess.Color)
+            BstTessFile = bst_fullfile(ProtocolInfo.SUBJECTS, subjectSubDir, ['tess_' importedBaseName '.mat']);
+        else
+            BstTessFile = bst_fullfile(ProtocolInfo.SUBJECTS, subjectSubDir, ['tess_textured_' importedBaseName '.mat']);
+        end
```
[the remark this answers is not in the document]
rcassani: Addressed in commit XXXX

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -331,63 +531,96 @@ function Close_Callback()
...
+        end
+        if ~isempty(res{3})
+            Digitize.Options.DistThresh = str2double(res{3}) / 1000; % mm to m
+        end
+        if ~isempty(res{4})
+            Digitize.Options.isBeep = str2double(res{4});
+        end
+
+        % Parse fiducials.
+        Digitize.Options.Fids = str_split(res{1}, '()[],;"'' ', true); % remove empty
```
rcassani, line 581: There is no need to have copies of the code to verify the same parameters

rcassani: Addressed in commit: XXXX

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -622,8 +991,8 @@ function CreateHeadpointsFigure()
     end 
 end
 
-%% ===== PLOT next point, or remove last =====
-function PlotCoordinate(isAdd) %(Loc, Label, Type, iPoint)
+%% ===== PLOT next point, or remove last or remove selected point =====
```
rcassani, line 994: Sections are all capitals

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -770,11 +1159,87 @@ function CreateMontageMenu(jMenu)
...
+                        jMenuDir.add(jMenuEgi);
+                    end
+                    if (jMenuNs.getMenuComponentCount() > 0)
+                        jMenuDir.add(jMenuNs);
+                    end
+                    if (jMenuWs.getMenuComponentCount() > 0)
+                        jMenuDir.add(jMenuWs);
+                    end
+                end
+            end
```
rcassani, line 1236: This part is a copy/paste of the menu generation for  `tree_callbacks` (https://github.com/brainstorm-tools/brainstorm3/blob/60fe915bc485de5f60ef5ed275eb6021f080df96/toolbox/tree/tree_callbacks.m#L2937)

Can you think in a way to reuse the code?
Maintaining this will be very annoying and prone to errors

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -823,47 +1288,70 @@ function SelectMontage(iMontage)
...
+        newMontage.Name = ChannelMat.Comment;
+        newMontage.Labels = {};
+        newMontage.ChannelFile = ChannelFile;
+
+        % Get cap landmark labels
+        [~, capLandmarkLabels] = auto_3dscanner('getEegCapLandmarkLabels', newMontage.Name);
+        if isempty(capLandmarkLabels)
+            bst_error('EEG cap not supported', Digitize.Type, 0);
+            return;
+        end
```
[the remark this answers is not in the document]
rcassani: Addressed in commit XXX

toolbox/sensors/panel_digitize.m
rcassani, line 1: It will be needed to apply the changes that were performed in `panel_digitize_2024` into this one

rcassani (commented):
## Overall 
Code looks good, almost ready to merge.

- Most of the code duplication mentioned in the previous review has been addressed.

- @chinmaychinara91, I have addressed several of the comments in this review. Please take a look to those changes so you can improve future PRs

- @chinmaychinara91, please remove unnecessary changes that make harder to look at the diff (see comments)

- @chinmaychinara91, also improvement in the comments is needed

## Other points
@tmedani and @chinmaychinara91 

- Using caps without landmarks. The code did not allow using caps that did not have defined "landmark markers", I think is a mistake. The automatic labelling is a tool to help avoiding doing tens of clicks, however, if the users decide to do so they should be able to do it. 

- Automatic labelling results. The automatic labelling should be mentioned in the tutorial (and maybe in the GUI) as an experimental tool, is has not been deeply tested, and the found positions are not always great.

- The code has been refactor, so now it can accept more than the "landmark markers" if the user has already acquire them. This greatly improves the automatic labelling.

toolbox/sensors/panel_digitize.m
rcassani: Changes in `panel_digitize_2024` from before the review (2e1aae8) to my last commit (d8b7b97):
https://github.com/brainstorm-tools/brainstorm3/compare/2e1aae8..d8b7b97#diff-989687fa8cf3c9eaec6f84f35347469e711da0e0ab109bd9a92ab7f7845e92b4

chinmaychinara91:
@rcassani thank you for the review. I will work on the comments you gave and get back.

toolbox/io/in_tess.m
```diff
@@ -272,6 +277,21 @@
...
+    subjectName = split(sMri.FileName, '/');
+    sSubject = bst_get('Subject', subjectName{1});
+    iTexSurf = find(cellfun(@(x)~isempty(regexp(x, 'tess_textured', 'match')), {sSubject.Surface.FileName}));
+    if ~isempty(iTexSurf)
+        iTexSurfComment = find(cellfun(@(x)~isempty(regexp(x, fileBase, 'match')), {sSubject.Surface(iTexSurf).Comment}));
+        if ~isempty(iTexSurfComment)
+            fileBase = sprintf('%s_%02d', fileBase, length(iTexSurfComment)+1);
+        end
+    end
+end
```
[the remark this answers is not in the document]
chinmaychinara91: The reason why I only did it for textured surface is because there is a use case i.e. lets says the user starts the digitization through from `File > Digitize > 3D scanner`, and then uses the `Use existing` option, then this unique naming of the surface would help the user to choose the one he wants easily. 

I agree we can drill it down to all other surface types as well but I did not find a use case

.

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -215,11 +336,24 @@ function Start()
         jTextFieldExtra = gui_component('text', jPanelInfo, [], '0', [], 'Head shape points digitized', @(h,ev)bst_call(@ExtraChangePoint_Callback), largeFontSize);
         initSize = jTextFieldExtra.getPreferredSize();
         jTextFieldExtra.setPreferredSize(Dimension(initSize.getWidth()*1.5, initSize.getHeight()*1.5))
+        if strcmpi(Digitize.Type, '3DScanner')
+            % Add Random 100 points generation button
```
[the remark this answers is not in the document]
chinmaychinara91: We can have it. I had given it a thought but then do we really need more then 100 points because this is just for doing the alignment correct ?
If we decide to have it then we should keep a lower/upper bound to avoid any under/over clustering. Let me know what you think.

toolbox/io/in_tess_wftobj.m
```diff
@@ -0,0 +1,191 @@
...
+    bst_error('Invalid call. Please specify the mesh file to be loaded.', 'Importing tesselation', 1);
+end
+
+%% ===== Parse the OBJ file: Set up import options and import the data =====
+% Specify range and delimiter
+% MATLAB R2016b to R2018a had 'DelimitedTextImportOptions'
+if (bst_get('MatlabVersion') >= 901) && (bst_get('MatlabVersion') <= 904)
+    opts = matlab.io.text.DelimitedTextImportOptions();
+else 
+    opts = delimitedTextImportOptions('NumVariables', 10);
```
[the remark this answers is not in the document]
chinmaychinara91: No it does not. I will convert the parsing logic similar to that in `in_channel_pom.m` to support older versions.

toolbox/sensors/auto_3dscanner.m
```diff
@@ -0,0 +1,215 @@
+function varargout = auto_3dscanner(varargin)
```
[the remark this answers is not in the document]
chinmaychinara91: `channel_detect_eegcap_auto` name sounds good. Will make the change.

toolbox/io/in_tess_wftobj.m
```diff
@@ -0,0 +1,180 @@
...
+        color = color./255;
+    end
+end
+
+% Centering vertices
+vertices = vertices - repmat(mean(vertices,1), [size(vertices, 1),1]);
+
+% For scalability, we determine the vertices' unit and convert them to 'meters' to match the 
+% coordinate space as that of Polhemus as implemented in the 'panel_digitize.m'
+vertices = channel_fixunits(vertices, 'm', 1, 1);
```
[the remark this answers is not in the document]
chinmaychinara91: @rcassani correct me if I am wrong. 

in the `channel_fixunits`, `FileUnits    : String, expected units` is where it is confusing me. Yes the vertices in the file are already in `mm` in the file and we want to convert it to `m` (multiplying the factor of 0.001 does that). So the "expected units" here is `m` correct ?

defaults/eeg/ICBM152/channel_WearableSensing_DSI_24.mat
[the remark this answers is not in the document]
rcassani: > The Generic >10-20 (19) cap layout inside BST has some of the points inside scalp.

And that should no be.

@tmedani, we would need to check all the caps and project the sensors on the respective scalps.

toolbox/io/in_tess_wftobj.m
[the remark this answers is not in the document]
rcassani: This will be low priority, we can release it only for R2026b or higher

toolbox/sensors/panel_digitize_2024.m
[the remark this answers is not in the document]
rcassani: @chinmaychinara91 you could tests to see if more points are actually helpful.

Related to head points, we need to update the tutorial, it mentions that it is good practice to take points in the boney part of the nose and eyebrows, but that is not what it is shown:

https://neuroimage.usc.edu/brainstorm/Tutorials/TutDigitize3dScanner#Collect_the_desired_number_of_head_shape_points

toolbox/io/in_tess_wftobj.m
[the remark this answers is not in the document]
rcassani: Yes, they way it is described is confusing, it means expected units in the file that is being read; not the units at the output of the script. Remember that locations in Brainstorm is saved as meters. Also, if you check the code, this `FileUnits` argument is used only for displaying info to the user when confirmation is needed.

toolbox/io/in_tess_wftobj.m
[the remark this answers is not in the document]
chinmaychinara91: Should we just disable the 3DScanner menu option and importing WFTOBJ surface type for < R2016b

defaults/eeg/ICBM152/channel_WearableSensing_DSI_24.mat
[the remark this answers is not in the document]
chinmaychinara91: As per slack, we will handle it in a separate PR

toolbox/sensors/panel_digitize_2024.m
chinmaychinara91: Addressed in https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/f1fc564a0fb5f331749fe6263ed8b29e9b5f4a2e

toolbox/sensors/panel_digitize.m
chinmaychinara91: Addressed in [82831c3](https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/82831c3f62c9d35c81cb8e4f3b6d2aee6a35cede)

toolbox/gui/view_surface_matrix.m
```diff
@@ -141,8 +183,8 @@
     sLoadedSurf.FileName = file_unique(sLoadedSurf.FileName, {GlobalData.Surface.FileName});
 end
 % Register in the GUI
-GlobalData.Surface(end + 1) = sLoadedSurf;
-    
+GlobalData.Surface(end+1) = sLoadedSurf;
+
```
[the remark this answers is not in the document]
chinmaychinara91: Addressed in [a1e2eb1](https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/a1e2eb12c4c45709fab05d993df6069ad353ae94)

toolbox/io/in_tess_wftobj.m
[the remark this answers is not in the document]
chinmaychinara91: > Should we just disable the 3DScanner menu option and importing WFTOBJ surface type for < R2016b

Did this here https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/21d0a0932aef2583bfc69a7a24c33d8eb0b01e3b

toolbox/sensors/auto_3dscanner.m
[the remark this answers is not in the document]
chinmaychinara91: Addressed in https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/7c0680596bddaf1bc9b43f473eb293cacb362da1

toolbox/sensors/auto_3dscanner.m
```diff
@@ -0,0 +1,215 @@
...
+end
+
+%% ===== FIND ELECTRODES ON THE EEG CAP =====
+function [capCenters2d, capImg2d, surface3dscannerUv] = findElectrodesEegCap(surface3dscanner)
+    % Hyperparameters for circle detection
+    % NOTE: these values can vary for new caps
+    minRadius = 1;
+    maxRadius = 25;
+    
+    % create a copy of the input mesh to add UV texture information to it as well
```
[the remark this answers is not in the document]
chinmaychinara91: Addressed in https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/81d17908111cb677ebcfe2f183c101dd85aff160 https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/20bc51a157ec19d411c611bcd45f337e30102e34

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -449,8 +682,15 @@ function ResetDataCollection(isResetSerial)
     Digitize.Transf = [];
     % Reset figure (also unloads in global data)
     if ~isempty(Digitize.hFig) && ishandle(Digitize.hFig)
-        %close(Digitize.hFig);
         bst_figures('DeleteFigure', Digitize.hFig, []);
+
+        % For 3D Scanner reload the surface
+        if strcmpi(Digitize.Type, '3DScanner')
+            % load the surface
```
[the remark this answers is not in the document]
chinmaychinara91: Addressed in https://github.com/brainstorm-tools/brainstorm3/commit/81d17908111cb677ebcfe2f183c101dd85aff160 https://github.com/brainstorm-tools/brainstorm3/commit/20bc51a157ec19d411c611bcd45f337e30102e34

toolbox/io/in_tess_wftobj.m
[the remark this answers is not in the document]
chinmaychinara91: Addressed in https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/cb0267b4155d98522803341741975c53a3f61734

toolbox/tree/tree_callbacks.m
```diff
@@ -1154,8 +1160,8 @@
                         case 'other'
                             jItemSetSurfTypeOther.setSelected(1);
                     end
-                end
-                
+                end 
+
```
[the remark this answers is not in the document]
chinmaychinara91: Addressed in https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/a1e2eb12c4c45709fab05d993df6069ad353ae94

toolbox/sensors/panel_digitize_2024.m
chinmaychinara91: Addressed in https://github.com/brainstorm-tools/brainstorm3/commit/81d17908111cb677ebcfe2f183c101dd85aff160 https://github.com/brainstorm-tools/brainstorm3/commit/20bc51a157ec19d411c611bcd45f337e30102e34

toolbox/sensors/panel_digitize_2024.m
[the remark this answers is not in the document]
chinmaychinara91: @rcassani proceeded with the tutorial recommended way (almost). Check https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/4265a266e57f3eb4f7515642aa31243ced6827c9. This does a pretty decent job in refining. What do you think ?

toolbox/sensors/panel_digitize_2024.m
rcassani: This code creates a GUI menu, it does not makes much sense to have them as a subfunction inside `auto_3dscanner`.

Create a new file called `menu_default_eegcaps.m` in `/toolbox/gui/`

`menu_default_eegcaps(jMenu, iAllStudies, isAddLoc)`

For treecallbacks:
`menu_default_eegcaps(jMenu, iAllStudies, isAddLoc)`

For digitizer:
`menu_default_eegcaps(jMenu)`
- Check that the Digitizer panels is open
- Check the digitizer version, to call the correct `AddMontage` function

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -546,11 +546,11 @@ function Close_Callback()
                     num2str(Digitize.Options.isBeep), {}};
 
     % Options to show for each type
-    switch lower(Digitize.Type)
-        case 'other'
-            iOptionsType = [1:7];
+    switch lower(Digitize.Type)     
         case '3dscanner'
             iOptionsType = [4:7];
+        otherwise
```
rcassani, line 552: Use explicit the case 'Digitize'

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -573,7 +573,7 @@ function Close_Callback()
         return;
     end
     % Digitizer: COM port
-    Digitize.Options.ComPort  = res{1};
+    Digitize.Options.ComPort  = str2double(res{1});
```
rcassani, line 576: Are you 100% sure of this change? 
Before the PR the parameter was a string, and it has been working without troubles

toolbox/io/in_tess.m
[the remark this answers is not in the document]
rcassani: Since the comment comes from the filename. To be the same Comment, the user would need to importg textured files with the same filenames, not likely. 

If you want to keep it, extend to all the surfaces types.

P.S. `strcmpi(FileFormat, 'WFTOBJ')` just checks for the format. While we do not support other textured files, we would certainly add more formats.

toolbox/sensors/panel_digitize_2024.m
chinmaychinara91: My mistake. Mistook `ComPort` for a number. Reverted it back in https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/04400ad04fc79350a2ba934dfa67c8dfa71bff1b.

toolbox/sensors/panel_digitize_2024.m
chinmaychinara91: Addressed in https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/9ab42ab23e0c44b551e1fdb40b6b1954f97ab532

toolbox/sensors/panel_digitize_2024.m
chinmaychinara91: Addressed in https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/9f2346ccf5821dd97537f4fe1e11ca296ec1d851

toolbox/io/in_tess.m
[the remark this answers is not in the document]
chinmaychinara91: Addressed in https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/65b4fbdb4349c2bb52ab3a27d7a47b72f6b82017

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -546,11 +546,11 @@ function Close_Callback()
...
-        case 'other'
-            iOptionsType = [1:7];
+    switch lower(Digitize.Type)     
         case '3dscanner'
             iOptionsType = [4:7];
+        otherwise
+            iOptionsType = [1:7];
     end
     
     % Ask options
```
chinmaychinara91, line 556: I am getting this error here:
Screenshot 2024-11-03 174826

It just happened randomly. Steps to reproduce:
1. Create a new subject by `File>Digitize>3D scanner`
2. Click `Edit Settings`. Click `OK`. Close the digitizer menu.
3. Click `File>Digitize>Digitizer`.
4. Click `Edit Settings`. Click `OK`. Close the digitizer menu.
5. Repeat steps 1 to 4, around 3-4 times
6. While doing step-4 I get this error. Once I get it then for every other time I try doing step-4 this error persists.

On debugging, I see that the `ComPort` which is set to `COM1` and `UnitType` which is set to `fastrak` as default, just goes empty at end of step-5. @rcassani can you check if it happens at your end as well.

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -331,66 +528,92 @@ function Close_Callback()
...
+    % Options to show for each type
+    switch lower(Digitize.Type)     
+        case '3dscanner'
+            iOptionsType = [4:7];
+        case 'digitize'
+            iOptionsType = [1:7];
+    end
+    
+    % Ask options
+    [resType, isCancel] = java_dialog('input', options_all(iOptionsType,1), [Digitize.Type ' configuration'], [], options_all(iOptionsType,2));
```
chinmaychinara91, line 557: I am getting this error here:
Screenshot 2024-11-03 174826

It just happened randomly. Steps to reproduce:
1. Create a new subject by `File>Digitize>3D scanner`
2. Click `Edit Settings`. Click `OK`. Close the digitizer menu.
3. Click `File>Digitize>Digitizer`.
4. Click `Edit Settings`. Click `OK`. Close the digitizer menu.
5. Repeat steps 1 to 4, around 3-4 times
6. While doing step-4 I get this error. Once I get it then for every other time I try doing step-4 this error persists.

On debugging, I see that the `ComPort` which is set to `COM1` and `UnitType` which is set to `fastrak` as default, just goes empty at end of step-5. @rcassani can you check if it happens at your end as well.

toolbox/io/in_tess_wftobj.m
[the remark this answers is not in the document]
chinmaychinara91: I tried using two methods to test if we can do some obj loading for versions < 2016b:
1. https://www.mathworks.com/matlabcentral/fileexchange/18957-readobj
2. https://www.mathworks.com/matlabcentral/fileexchange/134606-read-obj-files

Method-1 is parsing using line by line parsing considering the obj as a text file using `sscanf` but the problem is it is very very slow. I ran with R2014a and for loading the test obj file we are using it took 40 minutes which is not ideal at all. Most of the scans that users will be using will be around this dense or more.

Method-2 is an optimized way of handling method-1. But still for versions < 2016b it does not work well. I feel the current implementation is already a very good method if we want to keep things >= R2016b.

I have spent significant time on this already. Let me know what your thoughts are on this @rcassani or if you could recommend something better.

toolbox/gui/gui_brainstorm.m
```diff
@@ -147,7 +147,12 @@
             jMenuFile.addSeparator();
         end
         % === DIGITIZE ===  
-        gui_component('MenuItem', jMenuFile, [], 'Digitize', IconLoader.ICON_CHANNEL, [], @(h,ev)bst_call(@panel_digitize, 'Start'), fontSize);
+        jSubMenu = gui_component('Menu', jMenuFile, [], 'Digitize', IconLoader.ICON_CHANNEL,[],[], fontSize);
+            gui_component('MenuItem', jSubMenu, [], 'Digitizer',  IconLoader.ICON_CHANNEL,  [], @(h,ev)bst_call(@panel_digitize, 'Start'), fontSize);
+            % Not supported for < R2016b MATLAB versions
+            if bst_get('MatlabVersion') >= 901
+                gui_component('MenuItem', jSubMenu, [], '3D scanner', IconLoader.ICON_SNAPSHOT, [], @(h,ev)bst_call(@panel_digitize, 'Start', '3DScanner'), fontSize);
+            end
```
rcassani, line 155: Do not disable this feature because the reader for wavefront `.obj` files is not compatible with versions < R2016b

toolbox/io/in_tess.m
```diff
@@ -272,6 +277,21 @@
 
 %% ===== COMMENT =====
 % Add a comment field to the TessMat structure.
+
+% If there exists a surface with the same comment, make it unique
+if ~isempty(sMri)
+    % Get the current subject
+    subjectName = split(sMri.FileName, '/');
+    sSubject = bst_get('Subject', subjectName{1});
+    iSurf = find(cellfun(@(x)~isempty(regexp(x, 'tess_', 'match')), {sSubject.Surface.FileName}));
```
rcassani, line 286: All the files in `sSubject.Surface.FileName` are surfaces.
There is no need to test if they start with `tess`; just check the comments of the existent surfaces.

toolbox/io/in_tess_wftobj.m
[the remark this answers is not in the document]
rcassani: > Method-1 is parsing using line by line parsing considering the obj as a text file using sscanf but the problem is it is very very slow.

Parsing line by line is a bad idea:
1. There are 
2. The file is small to be read at once

> Method-2 is an optimized way of handling method-1. But still for versions < 2016b it does not work well as the readtable was not very well developed to handle files as table very well.

@chinmaychinara9, I'll push a snippet on how this can be done
It is a text file, parsing should not be hard

toolbox/tree/tree_callbacks.m
```diff
@@ -1172,6 +1178,15 @@
...
                 end
+
+                % === DIGITIZE (3D SCANNER) OPTION ===
+                % Not supported for < R2016b MATLAB versions
+                if strcmpi(nodeType, 'other') && ~isempty(regexp(filenameRelative, 'tess_textured', 'match')) && bst_get('MatlabVersion') >= 901
+                    gui_component('MenuItem', jPopup, [], 'Digitize (3D scanner)', IconLoader.ICON_SNAPSHOT, [], @(h,ev)bst_call(@panel_digitize, 'Start', '3DScanner', sSubject, iSubject, filenameRelative));
+                    % Separator
+                    AddSeparator(jPopup);
+                end
+
```
rcassani, line 1189: Remove this limitation

toolbox/io/in_tess_wftobj.m
```diff
@@ -0,0 +1,201 @@
...
+        elementTmp = regexp(allData(iLines, 2), '([-\.|\d])*', 'match')';
+        elementTmp = cat(1, elementTmp{:});
+        elementSize = size(elementTmp);
+        elementTmp = sscanf(sprintf(' %s', elementTmp{:}), '%f'); % Faster than str2double
+        elementData{iElement} = reshape(elementTmp, elementSize);
+    end
+    vertices   = elementData{1};
+    texture    = elementData{2};
+    faces      = elementData{3}; % TODO Select the proper columns
+    textureIdx = elementData{3}; % TODO Select the proper columns
```
rcassani, line 69: @chinmaychinara91 complete a test the text parsing

chinmaychinara91: Thanks @rcassani will test and get back

chinmaychinara91: Addressed in https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/249961aa496e4a4ac2844dbce2e6c83d3de31f4e

toolbox/io/in_tess_wftobj.m
```diff
@@ -45,49 +42,69 @@
...
+    allData = regexp(txtStr, '(\w)+ ([^\n])*\n', 'tokens'); % (\w) ignores comments (#)
+    allData = cat(1,allData{:});
+    % Read data for each element type
+    elementTags = {'v', 'vt', 'f'}; % Vertices, Texture, Faces
+    elementData = cell(1, length(elementTags));
+    % Parse element data
+    for iElement = 1: length(elementTags)
+        iLines = strcmp(elementTags{iElement}, allData(:,1));
+        elementTmp = regexp(allData(iLines, 2), '([-\.|\d])*', 'match')';
+        elementTmp = cat(1, elementTmp{:});
```
chinmaychinara91, line 61: fails at this line here for this data which came directly from Revopoint scan software: 
https://drive.google.com/file/d/1vYLkujKJX6dsVWxlOtEDeTO_xL0RAqpx/view?usp=sharing 

Error:
```
Error using cat
Dimensions of matrices being concatenated are not consistent.
```

Works well for this data which was cleaned and reexported using MeshLab: https://drive.google.com/file/d/1TyUONrM24NeQukFpkvNDGA-0FaFzgzVz/view?usp=sharing

toolbox/io/in_tess.m
chinmaychinara91: Addressed in https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/b9f2c6f4b3e009e2d3853e5b178dce227d7f0d53

toolbox/tree/tree_callbacks.m
chinmaychinara91: Addressed in https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/3852eeab0caeef1859926f4457e5da59d5b7e38e

toolbox/gui/gui_brainstorm.m
chinmaychinara91: Addressed in https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/3852eeab0caeef1859926f4457e5da59d5b7e38e

toolbox/io/in_tess.m
chinmaychinara91: Addressed in https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/b9f2c6f4b3e009e2d3853e5b178dce227d7f0d53

toolbox/io/in_tess_wftobj.m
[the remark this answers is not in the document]
chinmaychinara91: > From 40 min to 15 s, not bad a x160 speed up. Although, longer than the `readtable` performance

Indeed. This is a really good implementation @rcassani 💯

toolbox/io/in_tess_wftobj.m
chinmaychinara91: > fails at this line 61 here for this data which came directly from Revopoint scan software: https://drive.google.com/file/d/1vYLkujKJX6dsVWxlOtEDeTO_xL0RAqpx/view?usp=sharing
> 
> Error:
> 
> ```
> Error using cat
> Dimensions of matrices being concatenated are not consistent.
> ```


On debugging, for the failed file: While parsing for `vt` and concatenating it, it fails nbecause of this:

```
K>> elementTmp{574982}

ans = 

    '3.50314'    '-05'    '0.763701'
 ```

Comment (rcassani) [reply toolbox/io/in_tess_wftobj.m:61]: Solved in [c8af0d7](https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/c8af0d758009e225342708da00ac978364a95b95)

File: toolbox/io/in_tess.m (hunk 38 above)
Comment (rcassani) [reply toolbox/io/in_tess.m]: Avoid reinventing the wheel. There are functions in our code base to do these task.
See: [86c3ccf](https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/86c3ccf48f9272ee9076be0ee2b0ee7b334f3041)

File: toolbox/sensors/panel_digitize_2024.m (hunk 36 above)
Comment (rcassani) [reply toolbox/sensors/panel_digitize_2024.m:557]: Yes, this happens as those settings were cleared and saved in the global variable.
Solved in [366303f](https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/366303f23e2d789bb89a73c237323a97dd6cf3c8)

**Comment by rcassani** [comment]
@chinmaychinara91, @tmedani
It seems this PR is ready.

Please add a warning in the tutorials and in the GUI (e.g. when clicking on the Auto button) that the Auto feature is experimental, and results must be carefully verified.

**Comment by chinmaychinara91** [comment]
Did the tests. All look good now. It can be merged.

**Comment by chinmaychinara91** [comment]
Tutorials updated. Found some more bugs while doing some tests and fixed them in https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/53f04ca38eab3370a2b707eef5485ed7af60cd44 and https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/bcacbcaea9a71017d82ecb3f112bb3f89c1a59b4. It is good to go now.

File: toolbox/sensors/panel_digitize.m [hunk 42]
```diff
@@ -1023,7 +1023,9 @@ function SetSelectedButton(iButton)
 %% 3DSCANNER: AUTOMATICALLY DETECT AND LABEL EEG CAP ELECTRODES
 function EEGAutoDetectElectrodes(h, ev)
     global Digitize
-
+    
+    % Add disclaimer to users that 'Auto' feature is experimental
+    java_dialog('warning', 'This is an experimental feature. Please verify the results carefully.', 'Auto Detect EEG electrodes');
```
rcassani, line 1033: Give an option to cancel the action. 
Currently if the user cancels the window, the auto localization still continues.

toolbox/sensors/panel_digitize_2024.m
```diff
@@ -1664,7 +1664,7 @@ function BytesAvailable_Callback() %#ok<INUSD>
         UpdateList();
     end
     % Enable 'Auto' button IFF all landmark fiducials have been acquired
-    if strcmpi(Digitize.Type, '3DScanner')
+    if strcmpi(Digitize.Type, '3DScanner') && ~strcmpi(Digitize.Points(Digitize.iPoint).Type, 'EXTRA')
```
rcassani, line 1667: Isn't this change also needed in `panel_digitize`?

toolbox/sensors/panel_digitize.m
rcassani: You could do:

```
if ~java_dialog('confirm', ....
    return
end
```

rcassani: Done at [fb2a8eb](https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/fb2a8eb4ffeb452e3ddd6477421897c5f5de4443)

toolbox/sensors/panel_digitize_2024.m
rcassani: Did you try this? Because this is not how it works. If all the EEG positions are acquired, it will enable the Auto button even when EXTRA points are being acquired as all the required EEG are already there.

rcassani: Solved in [0e4637d](https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/0e4637d4c65ca8fa92d1da4d747a520779888289)

rcassani: In the same part of the code there was this bug: [67e0f8](https://github.com/brainstorm-tools/brainstorm3/pull/716/commits/f67e0f8610b739bf49bdc303b06dbce1147ac73f)

rcassani (approved):

PR #660, opened 2023-11-23
Synchronise multiple signal

Edouard2laire:
Hello, 

This process aims at aligning multiple signals acquired with different devices but sharing synchronizing events (typically sent to each recording device at the time of the acquisition).  This situation can arise easily when recording multimodal data such as EEG - iEEG,  EEG - fMRI, fNIRS - fMRI, or EEG - fMRI. 

Having the signal synchronised makes making easier to perform multimodal analysis, notably allowing the simultaneous exploration of the different signals. For example, here is a figure I recently presented at a conference, showing how we can understand better what is happening during a seizure by looking at the EEG, fNIRS, and pulse oximeter 

image

What's left to do: 

- [x]  Fix the code to save continuous files as output 
- [x]  Improve file history

Future work 

Here is a list of future things we might consider implementing but that is not required for this PR

- [x]  Allowing the simultaneous visualization of the different signals from Brainstorm (#656)
- [x]  Explicitly linking the different synchronized files. This could notably be used so that if an event is added to a file, then the same event would be added to all the events the file is synchronized with. 
- [ ]  Add the option to either crop the longer signal  keeping only the common signal (current implementation) or pad the shorter signal with either 0 or NaN

Let me know if you want me to send you an example dataset. I can generate a minimum brainstorm database with 2 files to synchronize. 

Regards, 
Edouard

Issue #656, opened 2023-11-11
opening multiple continous viewer at the same time.


Edouard2laire:
Hello, 

Do you think it would be possible to remove the limitation of having only one continuous signal displayed at a time?  https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/core/bst_memory.m#L765-L766

I think it's quite important especially if Brainstorm is encouraging toolbox to manipulate continuous files instead of imported files which support nicely to have multiple files displayed at the same time. 

Edouard

Edouard2laire:
Happy New Year @rcassani !! Do you think you would have time to work on this in january? Let me know if i can help :)

Edouard2laire:
any update on this ? :)

PR #717, opened 2024-07-06
Combine recordings

Edouard2laire:
Hello, 

This is the continuation of https://github.com/brainstorm-tools/brainstorm3/pull/660 in the attempt to enhance multimodal data analysis in Brainstorn. 

Since, having multiple raw viewer recording with multiple sampling rate is not accessible (https://github.com/brainstorm-tools/brainstorm3/issues/656), i made this process to combine recording and resampling them to the highest frequency. 

This allows to make such visualization directly from Brainstorm: 

image

This assume that the signals were synchronized

Todo: 

- [x] User-specified folder name
- [x] Copy channel flag
- [x] Copy video if present (not necessary actually. can be loaded from the other folder)

Question: 
- [ ] Better way to merge channels? 
I am not sure if the way I am doing is ok; especially if I need to merge headpoints / fiducials ... 

- [ ] How to deal with events ? 
Same thing, I am only copying events from the first file ? Should we merge them adding a sufixe to the name to identify the file ?


Edouard2laire:
hello @rcassani 

This PR is ready for review. 

The only things I am not sure are: the events ( how to merge event from multiple file - esp after synchronization there will be a lot of duplicate if we just concatenate the events), and the channel flag. 

Let me know if you have any questions. 

I'll try to complete all the PR I have as draft soon :)

Edouard2laire:
ok i completed my todo list. 

@rcassani, let me know when you would have time to review this. Let me know if you need an example dataset

rcassani:
> Better way to merge channels?
I am not sure if the way I am doing is ok; especially if I need to merge headpoints / fiducials ...



How to deal with events ?
Same thing, I am only copying events from the first file ? Should we merge them adding a sufixe to the name to identify the file ?

rcassani:
@Edouard2laire, apologies for closing the PR, I was trying to close my comment —duh!

rcassani:
@Edouard2laire, thanks for noticing the part for events with same name.

Don't you think it is necessary to keep the events separated per file?

If we merge the events at combining recordings, it will not possible to split them after. Also, it is possible that events with the same name are different type (single / extended) in that case merging is not even possible. On the other hand, if we keep the events separated in the combined file, the user can easily decide to merge them. This is done with the `process_evt_merge.m` or in the GUI (which calls the process), and this process will handle the cases when there is same name, different types, merging if same time sample, etc.

This could be implementing by reverting to 54d1cb47 and this patch:

```
diff --git a/toolbox/process/functions/process_combine_recordings.m b/toolbox/process/functions/process_combine_recordings.m
index 3926c2b9..37bc1b1f 100644
--- a/toolbox/process/functions/process_combine_recordings.m
+++ b/toolbox/process/functions/process_combine_recordings.m
@@ -135,6 +135,8 @@ function OutputFiles = Run(sProcess, sInputs)
         % Concatenate events
         for iEvent = 1 : length(sMetaData(iInput).F.events)
             tmpEvent = sMetaData(iInput).F.events(iEvent);
+            % Rename event if duplicated
+            tmpEvent.label = file_unique(tmpEvent.label, {NewEvents.label});
             % Add channel info
             addedChannelNames = {NewChannelMat.Channel(sIdxChNew{iInput}).Name};
             nOccurences = size(tmpEvent.times, 2);
```

rcassani:
> I don't see why it is necessary. At least I think it is important not to duplicate identical events. (since we duplicate them during the synchronization) it would lead to so many events in the combined file.

Makes sense for the event that was used for synchronization.

> But even then, it is easy to split: If an event has no channel information:it belong to all channel and can be split easily. just copy to each file.
If an event has channel information, you can attribute to file based on the channel

Yes, this is true. But there is no way to separate events based on the channels that the belong in the GUI or with a process. The user would need to program it. But by not merging (in the combining process), the user can decide to merge them, which can be done with the GUI and there is already a process. 

> Btw, i copied the code from the GUI and it doesnt seems it is calling the process : https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/gui/panel_record.m#L1963-L2029

Oh! I'll add this on my TODO list, outside of this PR. Thanks for the heads-up

rcassani:
> and yes we can merge the motion latter using the GUI or process

Sounds good

rcassani:
@Edouard2laire, two changes in [83b10d0](https://github.com/brainstorm-tools/brainstorm3/pull/717/commits/83b10d055da49c2f3e11e34a905729c42f8fefae):

1. With your recent changes, if an event was not channel-wise and was present in only on file. Once combined, this event was also not channel-wise, so it suggested that it applied to all channels, not only for channels of the file that had the event. This is now fixed.

2. The fact that duplicate names started in `_02` was bugging me. Now duplicated events are `_01`, ... 
So the intuitive thing to do at the moment of merging will be to set the event label to the name without the numbers.

PR #719, opened 2024-07-08
[bugfix] better identify raw file when exporting file

Edouard2laire:
Hello, 

This fix a bug where some raw file would not be identified as raw file during export.

### DIFF
--- a/toolbox/io/export_data.m
+++ b/toolbox/io/export_data.m
@@ -82,7 +82,7 @@ end
 % ===== LOAD DATA =====
 % CALL: export_data( DataFile, [], ExportFile )
 if ischar(DataFile) 
-    isRawIn = ~isempty(strfind(DataFile, '_0raw'));
+    isRawIn = ~isempty(strfind(DataFile, '_0raw')) || ~isempty(strfind(DataFile, '_raw')) 
     % Load initial file
     DataMat = in_bst_data(DataFile);
     % Get raw file structure


rcassani:
Hi @Edouard2laire, how did you get to have a file with such a name?
Raw files should have `_0raw` in their names

Edouard2laire:
it was created by https://github.com/brainstorm-tools/brainstorm3/pull/717

rcassani:
> i copied what we did for the synchronisation to get filename (https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/process/functions/process_sync_recordings.m#L255 ) :

Yeah, that's not right. It should have been `_0raw`

> It is recognize everywhere else in brainstorm as raw file despite not having _0raw

It does work for most of the cases as different checks for 'raw' type just find the `raw`  string, or check if the loaded `F` is a struct. However, it may have unexpected behaviours, e.g., setting a bad channel for a `_raw` file will not change the `ChannelFlag` inside `F`: https://github.com/brainstorm-tools/brainstorm3/blob/ec632b9c6579a01dad7c05df8919fc4c89c2909f/toolbox/sensors/panel_channel_editor.m#L722 , and in other places that do check for `_0raw`

I'll push the correction on `process_sync_recordings.m` directly on the master

Edouard2laire:
Is there a way I can fix the file already created without having to recompute ?

rcassani:
Apologies for the inconveniences. 
With Brainstorm close, you could renaming the file to add the zero and reloading the Protocol

PR #724, opened 2024-07-18
change default color order  for nirs

Edouard2laire:
By convention, we prefer to have HbO in red, HbR in blue, and HbT in green. This changes the order so the colors are good without modifying the figure. 

Before; 
image

After:

image

one small limitation with this fix is that displaying multiple HbO signals will make them all red.

### DIFF
--- a/toolbox/gui/figure_timeseries.m
+++ b/toolbox/gui/figure_timeseries.m
@@ -3250,6 +3250,10 @@ function PlotHandles = PlotAxes(iDS, hAxes, PlotHandles, TimeVector, F, TsInfo, 
     if (TsInfo.ShowLegend)
         if isempty(LinesColor)
             ColorOrder = panel_scout('GetScoutsColorTable');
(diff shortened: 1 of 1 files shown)


Edouard2laire:
Hello, 
i just tried and it doesnt work: 
strcmpi(TsInfo.Modality, 'NIRS') never goes in the if as TsInfo.Modality contains 'results'

Edouard2laire:
The result of MNE or MEM on the cortex. Created either using https://github.com/Nirstorm/nirstorm/blob/master/bst_plugin/inverse/process_nst_cmem.m or https://github.com/Nirstorm/nirstorm/blob/master/bst_plugin/inverse/process_nst_wmne.m

rcassani:
🙏 thanks for the extra information 
This has been fixed in: [1c331b6](https://github.com/brainstorm-tools/brainstorm3/commit/1c331b6d476cca3a88943f04375e2ea9b1b01bc6)

Edouard2laire:
works great. Thanks :)

PR #726, opened 2024-07-21
Adding Zeffiro Plugin 

tmedani:
Add Zeffiro Plugin & FEM mesh generation module with two possibility of usage
- Using basic options from the Brainstorm interface
- Using advanced option managed by the Zeffiro team


toolbox/process/functions/process_fem_mesh.m
```diff
@@ -190,6 +207,16 @@
         bst_report('Error', sProcess, [], 'Invalid downsampling factor.');
         return
     end
+    % Zeffiro: Mesh resolution -Edge Length-
+    OPTIONS.ZefMeshResolution = sProcess.options.zefMeshResolution.Value{1};
+    if isempty(OPTIONS.ZefMeshResolution) || (OPTIONS.ZefMeshResolution < 0.5) || (OPTIONS.ZefMeshResolution > 4.5)
+        bst_report('Error', sProcess, [], 'Invalid Mesh resolution value, please use value [0.5 - 4.5] mm.');
```
rcassani, line 213: What is the correct range for valid values? 
Are the limits included?

toolbox/process/functions/process_fem_mesh.m
```diff
@@ -1278,6 +1485,43 @@ function ComputeInteractive(iSubject, iMris, BemFiles) %#ok<DEFNU>
...
+                OPTIONS.ZefAdvancedInterface = 0;
+                % Get mesh resolution
+                [res, isCancel] = java_dialog('input', 'Mesh Resolution (edge length) in [mm]:', 'Zeffiro FEM mesh', [], num2str(OPTIONS.ZefMeshResolution));
+                if isCancel || isempty(res)
+                    return
+                end
+                OPTIONS.ZefMeshResolution = str2num(res{1});
+                % Check the values
+                if isempty(OPTIONS.ZefMeshResolution) || (OPTIONS.ZefMeshResolution < 0.5) || (OPTIONS.ZefMeshResolution > 4.5)
+                    errMsg = ['Invalid Mesh resolution value.' 10 'Please use value from this interval [1 - 4.5] mm.'];
```
rcassani, line 1509: What is the correct range for valid values? 
Are the limits included?

rcassani (approved):

toolbox/process/functions/process_fem_mesh.m
tmedani: the correct range can be outside of these limits, but we have noticed some issue when < 1 or >4.5

if the value is less than 1mm, the computation may crash since the resolution is to high and require a lot of memory.
if the value is > 5mm, the mesh is too coarse, and some elements maybe bigger than the layers between two faces, so the mesh will look messy. 

so to you question, you are right, please include the limits please change it to: <= or >= 

Thanks @rcassani

tmedani (commented):
adding some reviews

PR #729, opened 2024-07-24
Model selection in specparam and SPRiNT

lucwilson:
Additions and modifications to implement model selection in specparam and SPRiNT. 

Also removes GUI options for peak types. These were always experimental but can be slowly phased out as they are no longer being studied and have not been extensively validated (nor used in an article). Instead, users can select between default optimization settings (conventional implementations) and model selection (experimental).

This build still generates warnings when either the user does not possess the parallel processing toolbox or on first use of either specparam or SPRiNT. In this way, it is **not** ready to be merged quite yet. However, algorithm behaviour is as expected.

### DIFF
--- a/toolbox/process/functions/process_fooof.m
+++ b/toolbox/process/functions/process_fooof.m
@@ -260,7 +260,7 @@ function [fs, fg, errMsg] = FOOOF_matlab_nll(TF, Freqs, opt, hOT)
             'error',            [],...
             'r_squared',        []);
     % Iterate across channels
-    bst_progress('text',['Standby: ms-specparam is running in parallel']);
+    bst_progress('text','Standby: ms-specparam is running in parallel');
     try
         parfor chan = 1:nChan
             bst_progress('set', bst_round(chan / nChan,2) * 100);
@@ -273,7 +273,6 @@ function [fs, fg, errMsg] = FOOOF_matlab_nll(TF, Freqs, opt, hOT)
                 opt.peak_width_limits/2, opt.proximity_threshold, opt.border_threshold, opt.peak_type);
             model = struct();
             for pk = 0:size(est_pars,1)
-                params = [];
                 aperiodic_pars_tmp = [];
                 peak_pars_tmp = [];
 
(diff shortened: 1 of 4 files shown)


toolbox/process/functions/process_fooof.m
rcassani, line 1: Example warning:
``` 
Warning: File: process_fooof.m Line: 303 Column: 17
The temporary variable 'params' will be cleared at the beginning of each iteration of the parfor-loop. If 'params' is used before it is
set, a runtime error will occur. For more information, see Parallel for Loops in MATLAB, "Uninitialized Temporaries".
```

toolbox/process/functions/process_fooof.m
```diff
@@ -303,7 +307,11 @@
                 params = fmincon(@err_fm_constr,guess,[],[],[],[], ...
                     lb,ub,[],options,fs,spec(chan,:),opt.aperiodic_mode,opt.peak_type);
             catch
-                a = 0; % for catching errors
+                % for catching errors
+            end
+            if isempty(params)
+                % TODO: Improve handling of errors at estimating 'params'
```
rcassani, line 313: @lucwilson, this error handling needs to be improved

rcassani:
@lucwilson, in an unrelated topic, there is a problem if if the requested window is smaller than the data. It is partially handled here: https://github.com/brainstorm-tools/brainstorm3/blob/e08897975c28effa586893804bdec27b95104e60/toolbox/timefreq/bst_sprint.m#L77-L83

However, it seems `opt.nAverage` needs to be updated there as well, otherwise, it leads to unexpected shapes for `TF`
```
TF = nan(size(F,1), Nwin-(opt.nAverage-1), size(FreqVector,2));
```

toolbox/process/functions/process_fooof.m
lucwilson: So I think that we should throw an error if line 307 doesn't run correctly, in which case we either always have 2 +3n params or we error out. So we can integrate a more informative error message for failure to converge in line 307 but otherwise I would remove 312-315.

toolbox/process/functions/process_fooof.m
```diff
@@ -288,9 +438,6 @@
         fg(chan).peak_fit         = 10.^(model_fit-ap_fit); 
         fg(chan).error            = MSE;
         fg(chan).r_squared        = rsq_tmp(2);
-        if opt.return_spectrum
-            fg(chan).power_spectrum = spec(chan,:);
-        end
```
lucwilson, line 293: So this was added so that SPM (if I remember correctly) can call the Brainstorm version of specparam within their software, and this was only here so that those users could have access to the original power spectrum. return_spectrum is not accessible in the GUI because it is redundant with the TF field, but I think we should consider keeping this in so as to not kill compatibility with SPM. But let me know your thoughts :)

toolbox/process/functions/process_fooof.m
rcassani: @lucwilson, sounds good to me.

A good approach would be to add an `errMsg` output into `FOOOF_matlab_nll` (as well as in `FOOOF_matlab` and `FOOOF_python`), fill it an early return in the `catch` section. 

So if `errMsg` is not empty in `process_fooof('Run')` the error message will be displayed and added to the report.

toolbox/process/functions/process_fooof.m
rcassani: Agree, we need to keep it

toolbox/process/functions/process_fooof.m
rcassani: Addressed in [6cead9f](https://github.com/brainstorm-tools/brainstorm3/pull/729/commits/6cead9f18dd7f45f7b7eebaf1e5041f856c2b185)

PR #733, opened 2024-08-22
[MEM] Reduce file size when using wavelet-MEM 

Edouard2laire:
Hi, 

This implements the idea I started here: https://neuroimage.usc.edu/forums/t/stange-bug-when-saving-mem-results/45901/7 when using wMEM. The main idea is to attempt to reduce the file size. this is mostly important for nirs as we are about to use wMEM for long-duration data but it is also nice for EEG/MEG. 

The principle is the following: for wMEM, the data is first transformed using a time-frequency representation (nSensor x nBox), then each time-frequency box is localized, and finally, the time-course is recreated by inverting the time-frequency transformation. 

The last step can be seen here: 
This can be seen here: https://github.com/Edouard2laire/best-brainstorm/blob/wMEM-nirs/best/wmem/solver/be_wmem_solver.m#L202-L204

```matlab

    inv_proj    =   be_wavelet_inverse( wav, OPTIONS );
    inv_proj    =   inv_proj(:,obj.info_extension.start:obj.info_extension.end);
    obj.ImageGridAmp =  obj.ImageGridAmp * inv_proj;

```

The objective of this PR is instead of saving the result of the multiplication obj.ImageGridAmp (nVertex x nBox) * inv_proj (nBox x nTimes) to save each matrix separately and compute the multiplication on loading. 

```matlab

% Results (full temporal sequence)
Results = struct(...
    'ImageGridAmp',     obj.ImageGridAmp, ...
    'ImagingPostKernel',inv_proj, ...
    'ImagingKernel',    [], ...
    'MEMoptions',       OPTIONS, ...
    'MEMdata',          obj);

```

On the EEG data for the epilepsy tutorial, it means that we go from a matrix ImageGridAmp: [15002x206 double] 
to ImageGridAmp:      [15002x29 double] and ImagingPostKernel: [29x206 double] 

so that the size of the file in brainstorm db goes down from 120356918 Bytes to 77919214 Bytes ( 35% reduction). 
Let me know if this change would be ok for you to integrate in Brainstorm. 

Note; I was not sure about the name ImagingPostKernel, so if you have a better idea :) 

Edouard

### DIFF
--- a/toolbox/process/functions/process_inverse_2018.m
+++ b/toolbox/process/functions/process_inverse_2018.m
@@ -694,9 +694,7 @@ function [OutputFiles, errMessage] = Compute(iStudies, iDatas, OPTIONS)
                 OPTIONS.FunctionName  = 'mem';
                 % Call the mem solver
                 [Results, OPTIONS] = be_main(HeadModel, OPTIONS);
-                if isfield(Results, 'nComponents')
-                    Results.nComponents = Results.nComponents;
-                else
+                if ~isfield(Results, 'nComponents') || isempty(Results.nComponents)
                     Results.nComponents = round(max(size(Results.ImageGridAmp,1),size(Results.ImagingKernel,1)) / nSources);
                 end
 
--- a/toolbox/process/functions/process_inverse_2018.m
+++ b/toolbox/process/functions/process_inverse_2018.m
@@ -694,7 +694,12 @@ function [OutputFiles, errMessage] = Compute(iStudies, iDatas, OPTIONS)
                 OPTIONS.FunctionName  = 'mem';
                 % Call the mem solver
                 [Results, OPTIONS] = be_main(HeadModel, OPTIONS);
-                Results.nComponents = round(max(size(Results.ImageGridAmp,1),size(Results.ImagingKernel,1)) / nSources);
+                if iscell(Results.ImageGridAmp)
+                    Results.nComponents = round(max(size(Results.ImageGridAmp{1},1),size(Results.ImagingKernel,1)) / nSources);
+                else
+                    Results.nComponents = round(max(size(Results.ImageGridAmp,1),size(Results.ImagingKernel,1)) / nSources);
+                end
+                
                 % Get outputs
                 DataFile = OPTIONS.DataFile; 
                 Time     = OPTIONS.DataTime;
--- a/toolbox/process/functions/process_inverse_2018.m
+++ b/toolbox/process/functions/process_inverse_2018.m
@@ -694,12 +694,12 @@ function [OutputFiles, errMessage] = Compute(iStudies, iDatas, OPTIONS)
                 OPTIONS.FunctionName  = 'mem';
                 % Call the mem solver
                 [Results, OPTIONS] = be_main(HeadModel, OPTIONS);
-                if iscell(Results.ImageGridAmp)
-                    Results.nComponents = round(max(size(Results.ImageGridAmp{1},1),size(Results.ImagingKernel,1)) / nSources);
+                if isfield(Results, 'nComponents')
+                    Results.nComponents = Results.nComponents;
                 else
                     Results.nComponents = round(max(size(Results.ImageGridAmp,1),size(Results.ImagingKernel,1)) / nSources);
                 end
-                
+
                 % Get outputs
                 DataFile = OPTIONS.DataFile; 
                 Time     = OPTIONS.DataTime;
--- a/toolbox/io/in_bst_results.m
+++ b/toolbox/io/in_bst_results.m
@@ -70,7 +70,6 @@ else
     if any(ismember({'ImageGridAmp','ImagingKernel'}, FieldsToRead))
         FieldsToRead{end + 1} = 'ImagingKernel';
         FieldsToRead{end + 1} = 'ImageGridAmp';
-        FieldsToRead{end + 1} = 'ImagingPostKernel';
         FieldsToRead{end + 1} = 'GoodChannel';
         FieldsToRead{end + 1} = 'OPTIONS';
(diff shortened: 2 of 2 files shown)


rcassani:
@Edouard2laire, it makes sense to store the factors to save space. 

Though, a new field may not be the optimal place.
A better solution would be to save the factors inside `ImageGridAmp`:

```
ImageGridAmp = {obj.ImageGridAmp, inv_proj};
```

Then when reading
```
if iscell(ImageGridAmp)
    tmp = ImageGridAmp{1};
    for im = 2 : length(ImageGridAmp)
        tmp = tmp * ImageGridAmp{im};
    end
    ImageGridAmp = tmp;
end
```

Edouard2laire:
Works for me. 

I also realized i could save inv_proj as a sparse matrix to gain even more space: 

```
  Name         Size                   Bytes  Class     Attributes

  inv_proj         2788x11233              8443632  double    sparse    
  inv_proj         2788x11233            250540832  double           
```

Here are the non-zero coeficient of inv_proj: 
image

Edouard2laire:
I just tried saving as you said, but it is causing some issue with existing code: 

Error using size
Dimension argument must be a positive integer scalar within
indexing range.

Error in process_inverse_2018>Compute (line 697)
                Results.nComponents = round(max(size(Results.ImageGridAmp,1),size(Results.ImagingKernel,1)) / nSources);

Edouard2laire:
ok. this is ready for review :) 

Just, to double check, if all the latter script uses in_bst_results to load their results; then the change of format should not impact them. 
Should we allow to get the factor separatly depending on the flag LoadFull ? 

One thing we could do later is to allow the reconstruction of specific frequency bands of the signal by only reconstructing specific boxes. 

For example, here is the code to reconstruct only the timecourse for scale 3: 

>> idx = tmp.MEMoptions.automatic.selected_samples(2,:) == 3;
>> data = tmp.ImageGridAmp{1}(:,idx) * tmp.ImageGridAmp{2}(idx,:);

rcassani:
> Just, to double check, if all the latter script uses in_bst_results to load their results; then the change of format should not impact them.

That's is right, the changes in this PR will not only be used for cell `ImageGridAmp`.
Not sure what "change of format should" you mean here.

> Should we allow to get the factor separatly depending on the flag LoadFull ?
I don't think so, the expected result of using the `LoadFull` flag is "full" sources [nSources, nTime]

toolbox/process/functions/process_inverse_2018.m
```diff
@@ -694,7 +694,12 @@
                 OPTIONS.FunctionName  = 'mem';
                 % Call the mem solver
                 [Results, OPTIONS] = be_main(HeadModel, OPTIONS);
-                Results.nComponents = round(max(size(Results.ImageGridAmp,1),size(Results.ImagingKernel,1)) / nSources);
+                if iscell(Results.ImageGridAmp)
+                    Results.nComponents = round(max(size(Results.ImageGridAmp{1},1),size(Results.ImagingKernel,1)) / nSources);
+                else
+                    Results.nComponents = round(max(size(Results.ImageGridAmp,1),size(Results.ImagingKernel,1)) / nSources);
+                end
+
```
rcassani, line 702: For simplicity. Can the number of components be added inside the `be_main()` call in the next version of Brainentropy which I guess will have the cell ImageGridAmp?

If so, we can add a check for the installed Brainentropy version, if it is not the one with cell, we can get the components as before.

rcassani (changes requested):

toolbox/process/functions/process_inverse_2018.m
Edouard2laire: I guess we could. Since MEM only supports dipole perpendicular to the cortical surface, for now, the number of components is always 1.  (also, MEM never returns Kernel)

i am not sure how checking the version number is easier than checking for cell/not cell.

Edouard2laire:
> > Just, to double check, if all the latter script uses in_bst_results to load their results; then the change of format should not impact them.
> 
> That's is right, the changes in this PR will not only be used for cell `ImageGridAmp`. Not sure what "change of format should" you mean here.

i just meant, we are not going to break some other process by saving the decomposition in the file as when loading the full source is automatically computed during loading. So for another script, it doesn't change anything.



> > Should we allow to get the factor separatly depending on the flag LoadFull ?
> 
> I don't think so, the expected result of using the `LoadFull` flag is "full" sources [nSources, nTime]

oh i see. thx

toolbox/process/functions/process_inverse_2018.m
Edouard2laire: would https://github.com/brainstorm-tools/brainstorm3/pull/733/commits/d2aa00ba9293528833608db63052e31e768521e6 works ?

rcassani:
> [what did i just wrote....](https://github.com/brainstorm-tools/brainstorm3/pull/733/commits/954a1e3794d90b3bc90717a2f73c0a0eb310d2d1) 

🤣

toolbox/process/functions/process_inverse_2018.m
rcassani: > i am not sure how checking the version number is easier than checking for cell/not cell.

In computation it may be the same, but with the change ([954a1e3](https://github.com/brainstorm-tools/brainstorm3/pull/733/commits/954a1e3794d90b3bc90717a2f73c0a0eb310d2d1)) it is now easier to read and grasp

rcassani:
@Edouard2laire, just to be sure. The two changes: (1) Adding the `nComponent` field, and (2) `ImageGridAmp` as cell will be added at the same time in next version on BrainEntropy, right?

If so, we can merge this PR now.

Edouard2laire:
Yes, that is correct.

PR #735, opened 2024-09-03
[bugfix] exit view_surface_data peacefully

Edouard2laire:
Currently, if you open two files with different time definition, Brainstorm will show the nice error: 

'''
** Error: Time definition for this file is not compatible with the other files
** already loaded in Brainstorm.
** Close existing windows before opening this file, or use the Navigator.

''' 

followed by a matlab error: 

''''
** Error: Line 175:  isempty
** Not enough input arguments.
** 
** Call stack:
** >view_surface_data.m at 175
** >tree_callbacks.m at 293
** >bst_call.m at 28
** >panel_protocols.m>CreatePanel/protocolTreeClicked_Callback at 125
** >bst_call.m at 28
** >panel_protocols.m>@(h,ev)bst_call(@protocolTreeClicked_Callback,h,ev) at 75
** 
'''

This PR ensure that view_surface_data doesn't continue if an error occurs during loading. Removing the matlab error.

### DIFF
--- a/toolbox/gui/view_surface_data.m
+++ b/toolbox/gui/view_surface_data.m
@@ -167,6 +167,10 @@ switch (DataType)
         iDS = bst_memory('GetDataSetSubject', sSubject.FileName, 1);
 end
 
+if isempty(iDS)
+    % Something went wrong before
+    return;
+end
 
 %% ===== MODALITY =====
 if isempty(Modality)


rcassani:
Beside detecting the empty iDS, we can ask users if they want to unload the other files.

@Edouard2laire, test this solution and check if it saves you few clicks.

Same idea can be implemented of other files (matrix, time-freq and dipoles), where the time definition does not match the current loaded files

Edouard2laire:
it does work. There is however a small issue: 

If you open a  nirs file first, some montage a created to have the color nicely done: 
image


but they dont appears after doint the unload thing: 

image


then montage is: 

image

Edouard2laire:
Moving the loading of the channel after the eventual unloading is fixing the issue. See https://github.com/brainstorm-tools/brainstorm3/pull/735/commits/ee4b25a89bad693b4f93f1a80169a404194e7b43 let me know if that works for you.

rcassani:
@Edouard2laire, the bug of not loading the montages seems that was there from a long long time.
Good thing it is now fixed.

I added support for different time definition to matrices as well

Edouard2laire:
thanks a lot.  I don’t really have data to test on those, so I think you can merge if that’s ok :)

PR #736, opened 2024-09-05
Add atlas selection as an option for Brainsuite segmentation process

chinmaychinara91:
Currently BrainSuite processing in Brainstorm just uses the USC Brain atlas by default and though the USC Lobes atlas (https://brainsuite.org/usclobesatlas/) is available in the BrainSuite installation package there is no way to use it in Brainstorm through the `process_segment_brainsuite` .  This PR exposes this option to the users.

@rcassani  let me know if there is a better way. 

@tmedani @ajoshiusc FYI based on our discussion.


chinmaychinara91:
Closing this PR because USC lobes currently is not distributed as part of BrainSuite. Had a wrong understanding. Will open a new PR to handle this.

@tmedani @ajoshiusc

PR #740, opened 2024-09-26
EEG-MEG fusion: remove warning when using MEM

Edouard2laire:
Remove the waning for the fusion of EEG/MEG when using MEM since it was extensively tested in https://pubmed.ncbi.nlm.nih.gov/26016950/ and https://pubmed.ncbi.nlm.nih.gov/29164737/


Edouard2laire:
ok @rcassani . i made the change you asked to only modify the panel_inverse_2018

PR #743, opened 2024-10-07
fix nComponent in process_inverse

Edouard2laire:
This applies the changes of https://github.com/brainstorm-tools/brainstorm3/pull/733/files to process_inverse. 

Btw, shouldn't process_inverse_mem call process_inverse_2018 (like what happens in the panel) instead of process_inverse ?


rcassani:
> Btw, shouldn't process_inverse_mem call process_inverse_2018 (like what happens in the panel) instead of process_inverse ?

Yes, it should. That process is part of the [best-brainstorm](https://github.com/Edouard2laire/best-brainstorm/tree/master) plugin, and it seem that file has not be maintained for a while:
https://github.com/Edouard2laire/best-brainstorm/commits/master/processes/process_inverse_mem.m

Edouard2laire:
could we still merge this while i look at the change to make in process_inverse_MEM ? or if you could help, that would get great :)  I never looked into that process

PR #744, opened 2024-10-10
Recompile mex file for PAC (fix #741)

### LINKED ISSUE #741: Inconsistent behaviour on macOS (Intel) of `direct_pac_mex.c`
**Description**
The [provided .mex files](https://github.com/brainstorm-tools/brainstorm3/tree/master/toolbox/connectivity/private) to compute PAC do not behave the same on macOS using Intel CPU.

**To reproduce**
Run the script [`tutorial_resting`](https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/script/tutorial_resting.m) on a macOS (Intel) system, using the options `ismex = 1` and `parallel = 0` in the process `process_pac`.

**Expected behaviour**
On line 181 of [`bst_pac`](https://github.com/brainstorm-tools/brainstorm3/blob/7ef5e6fef78bc41bec942506065abb942e919532/toolbox/connectivity/bst_pac.m#L181):
```
DirectPAC = direct_pac_mex(permute(PHASE, [2,3,1]), permute(AMP, [2,3,1]));
```

the function `direct_pac_mex` should return a complex double matrix of shape `[16,19]` (it does in Linux and Windows). 
However in macOS (Intel) it returns a complex double matrix of shape `[0, 2177491158568009729]`.

Edouard2laire:
This recompile the Mex direct_pac_mex.c on mac to solve the discrepancy with windows mex (see #741)

Test code 
```matlab

load("bst_pac_test_data.mat");

DirectPAC = direct_pac_mex(permute(PHASE, [2,3,1]), permute(AMP, [2,3,1]));

size(DirectPAC)

all(all(DirectPAC == DirectPACGoodResult))

```

On matlab2023b, produce : 

```
ans =

    16    99


ans =

  logical

   1

``` 

On matlab2021b, produce, before recompiling : 

```
ans =

   1.0e+18 *

         0    7.0915
```

After recompiling: 

``` 

ans =

    16    99


ans =

  logical 
1
```

The mex file were produced using the command: mex -largeArrayDims -compatibleArrayDims  direct_pac_mex.c using matlab2023b for ARM Mac, and 2021b for the intel version. 

note: i had to copy the mex files outside of the private folder as i was not able to add it to matlab path. (not sure if that is expected)

note2: i tried to use the mex before commit https://github.com/brainstorm-tools/brainstorm3/commit/cd2568357ba104e000c6657fd6f856c4ef09dcc6 and its working well too. it seems this commit introduced a bug during compilation


Edouard2laire:
@rcassani seems recompiling the mex fixed the issue

Edouard2laire:
Edit: After doing some test, the reason why the previous mex was not working was probably because the flag -compatibleArrayDims was not used during compilation. 



```

>> mex -largeArrayDims  direct_pac_mex.c
Building with 'Xcode with Clang'.
/Users/edelaire1/Documents/script/test-bst/direct_pac_mex.c:60:40: warning: incompatible pointer types passing 'int[3]' to parameter of type 'const mwSize *' (aka 'const unsigned long *') [-Wincompatible-pointer-types]
   60 |         plhs[0] = mxCreateNumericArray(nDimA, dimsOut, mxDOUBLE_CLASS, 1);      
      |                                               ^~~~~~~
/Applications/MATLAB_R2021b.app/extern/include/matrix.h:955:49: note: passing argument to parameter 'dims' here
  955 | mxCreateNumericArray(mwSize ndim, const mwSize *dims, mxClassID classid, mxComplexity flag);
      |                                                 ^
1 warning generated.
```

Those warning are not present when using the 32-bit API (compatibleArrayDims flag) (https://www.mathworks.com/help/matlab/matlab_external/upgrading-mex-files-to-use-64-bit-api.html)

rcassani:
> note: i had to copy the mex files outside of the private folder as i was not able to add it to matlab path. (not sure if that is expected)

Yes, this is expected, it is using the "Private functions" approach in Matlab
https://www.mathworks.com/help/matlab/matlab_prog/private-functions.html

> Edit: After doing some test, the reason why the previous mex was not working was probably because the flag -compatibleArrayDims was not used during compilation.

Thanks for going deeper in this issue. I just tested on Intel and Apple Silicon and both work as they should.

PR #745, opened 2024-10-25
[bugfix] Ensure that wavelength are integer

Edouard2laire:
See error reported here: https://neuroimage.usc.edu/forums/t/nirstorm-error-mbll-raw-to-delta-hbo-hbr-hbt/48659/4


rcassani:
@Edouard2laire, can you check if the last commit solved the wrong labels on your case.
Otherwise, please share the problematic channel file

Edouard2laire:
Seems its working

PR #747, opened 2024-11-01
Plugins: Update iso2mesh to 1.9.8, adds support for ARM macOS

Edouard2laire:
This PR updates the plugins for the ARM Mac: 

- iso2mesh to 1.9.8
A new version of iso2mesh was released last week by @fangq notably adding the support for arm computer. 
The details of the changes can be seen here: https://github.com/fangq/iso2mesh/releases/tag/v1.9.8 
This updates the plugin version of iso2mesh to this new release. 

- xdf to master. 
xdf is now working on ARM Mac.  (see https://github.com/xdf-modules/xdf-Matlab/pull/20)

- Improve the support of conflict between plugins

We are very close to having all the plugins working for Mac with only Duneuro not being compatible.


rcassani:
This could addressed in two ways:

1. Not allow `iso2mesh` to be loaded when `jsnirfy` is being loaded.
   Done by adding `iso2mesh` in the field `UnloadPlugs` for `jsnirfy`
   
2. Removing the copy of `jsnirfy` from `iso2mesh` when this second is installed.
   We can add the plugin `jsnirfy` as dependency for `iso2mesh`, but this will cause troubles if they start to differ.

Edouard2laire:
Note: I opened a PR to have xdf available on arm here: https://github.com/xdf-modules/xdf-Matlab/pull/20 
Once this is merged there will only be duneneuro left not working on Mac with arm processors. 

So maybe we should remove the warning at startup: https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/core/bst_startup.m#L160-L164

PR #748, opened 2024-11-01
Call process_inverse_2018  without noiseMethod (MEM)

Edouard2laire:
See https://github.com/multifunkim/best-brainstorm/pull/28


rcassani:
It seems we added it in parallel. See: [7c9e3ea](https://github.com/brainstorm-tools/brainstorm3/commit/7c9e3ea07eb46f64346133b769e419fbe30ae878)

PR #751, opened 2024-11-12
[bugfix] import non-skull stripped version of  T2

Edouard2laire:
Hello, 

Currently,  when importing T2 image from freesurfer segmentation, we import the image that has been skull stripped 
image

However, this can be an issue later when we do segmentation inside Brainstorm using SPM (to get tissue map) as SPM is expecting a full T1 and T2. 

This is importing the full T2 image: 

image

Regards, 
Edouard


Edouard2laire:
Edit it seems actually SPM is working worse on with this T2.. will investigate before reopening

PR #754, opened 2024-11-26
Functions for MRI/CT skull stripping

chinmaychinara91:
While importing CT ~and co-registering using  SPM or CT2MRI~, give option for skull stripping either using **BrainSuite (Brain Surface Extractor)** or **SPM (Tissue segmentation)**

Screenshot 2024-11-27 213247

EDIT: @rcassani 

In additon:
* Add popmenu callbacks to perform registration with BrainSuite
* Add popmenu callbacks to perform skull stripping (with SPM or BrainSuite)


toolbox/anatomy/mri_coregister.m
```diff
@@ -470,3 +468,91 @@
     bst_progress('stop');
 end
 
+%% ===== GENERATE SKULL MASK BY SKULL STRIPPING =====
+function [sMriMask, MaskTag] = generate_skull_mask(NiiRefFile, TmpDir, MaskMethod)
```
rcassani, line 472: Masked CT does not have any track of the skull stripping procedure in its history.

toolbox/anatomy/mri_coregister.m
```diff
@@ -470,3 +468,91 @@
...
+            % Set the SPM logo
+            bst_plugin('SetProgressLogo', 'spm12');
+            % Perform skull stripping using SPM Tissue Segmentation
+            bst_progress('text', 'Skull Stripping: SPM Segment...');
+            % Reset matlabbatch to start fresh
+            clear matlabbatch;
+            % Get the TPM atlas
+            TpmFile = bst_get('SpmTpmAtlas', 'SPM');
+            % Read the reference MRI
+            sMri = in_mri_nii(NiiRefFile, 0, 0, 0);
```
rcassani, line 535: sMRI converted to `.nii` before the function, then converted back to sMRI. Not very efficient

toolbox/io/import_mri.m
```diff
@@ -320,24 +320,42 @@
             errMsg = 'Reslice: No SCS transformation available for the reference volume. Set the fiducials first.';
             RegMethod = ''; % Registration will not be performed
         end
-
+        
+        % === ASK SKULL STRIPPING ===
+        if isInteractive && isCt && (strcmpi(RegMethod, 'SPM') || strcmpi(RegMethod, 'CT2MRI'))
+            % Ask if the user wants to mask out region outside skull in CT
+            [MaskMethod, isCancel] = java_dialog('question', ['<HTML><B>Do skull stripping to clean the CT volume?</B><BR>' ...
+                                                            'This operation cleans the CT to exclude any thing outside the skull.<BR><BR>' ...
```
rcassani, line 328: Skull-stripping also removes the skull, not only "any thing outside the skull"

rcassani (changes requested):
The PR implements the different methods for skull stripping when coregistering a CT to a MRI volume.

1. As it is skull stripping is inside `mri_coregister` then it only available on importing C, so there are serious limitations:

- It cannot be used on an already imported coregisted CT volume
- It cannot be used if a CT volume is imported and then coregistered from popup menu

The tags are in wrong order, e.g., '...CT_spm_masked_spm_reslice' it implies masked happened before reslicing, which is not the case

toolbox/anatomy/mri_coregister.m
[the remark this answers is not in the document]
rcassani: Fixed in [1dc3dab](https://github.com/brainstorm-tools/brainstorm3/pull/754/commits/1dc3dab4480d54f5b2d422a233a639bba9e49ad7)

toolbox/anatomy/mri_coregister.m
rcassani: Done in `mri_skullstrip` in [1dc3dab](https://github.com/brainstorm-tools/brainstorm3/pull/754/commits/1dc3dab4480d54f5b2d422a233a639bba9e49ad7)

toolbox/io/import_mri.m
rcassani: See [915277a](https://github.com/brainstorm-tools/brainstorm3/pull/754/commits/915277a2f3bec4d27ad83a3800839b34d3de5a1e)

toolbox/anatomy/mri_skullstrip.m
```diff
@@ -0,0 +1,197 @@
+function [MriFileMask, errMsg, fileTag, binBrainMask] = mri_skullstrip(MriFileSrc, MriFileRef, Method)
```
chinmaychinara91, line 1: @rcassani Should we also add the `sMriMask` structure ? While it might be ambiguous since users can use the output `MriFileMask` with `bst_memory(LoadMri)` to load it if they want but I was just comparing it with `mri_reslice` and thinking if we should.

rcassani: I would say that there is no value in adding `sMriMask` as output argument. If the user is working with structures, this output argument will be repeated. There is not point in adding it.

> I was just comparing it with mri_reslice and thinking if we should.

Did you find a use for that output argument? 
That `mri_reslice` has it is not a good enough reason to add it.

chinmaychinara91: > Did you find a use for that output argument?

No I did not. I think it is ambiguous.

rcassani: What do you mean with ambiguous? 

The same happens with the input arguments they can be either structures or filenames.
You can see it a way to function overloading

toolbox/anatomy/mri_skullstrip.m
rcassani, line 1: I appreciate your enthusiasm, in expanding the help for the function.

However, I think the USAGE was clear, it shows that the if the function has as input filesnames for files in the DB e.g., `MriFileSrc` it returns a filename `MriFileMask` which implies is already register in the DB. If the inputs are structures `sMriSrc` (with the start `s` as we do it in the code and mentioned in the naming conventions), the result is a  structure `sMriMask`.

Moreover, the added descriptions has troubles:
1. It does not mention that the inputs can be structures
2. It misleads saying that the input paths are relative, they can also be absolutes
3. It does not mention that the output can be structure

rcassani (approved):

PR #709, opened 2024-05-28
NIRS: Allow finding channels using Source, Detector and/or Wavelenth info

Edouard2laire:
Hello, 

In NIRS, sometimes can be useful to mark a detector, or a source as bad as we know from the acquisition that this detector or source was not well placed.  Currently, to all channels that involve `S1`, for example, we have to write `S1D1WL680`, `S1D1WL830`, `S1D2WL680`, `S1D2WL830`, ... to list all the channels containing S1 which is awful :) 

This PR allows findings all NIRS channels using an specific Source, Detector or Wavelength.

Examples: 
* Searching `S1`: Find NIRS channels using: Source `S1`, any Detectors, any Wavelengths
* Searching `D1`: Find NIRS channels using: any Sources, Detector `D1`, any Wavelengths
* Searching `S1D1`: Find NIRS channels using: Source `S1`, Detector `D1`, any Wavelengths
* Searching `WL830`: Find NIRS channels using: any Source, any Detector, Wavelength `WL830`
* Searching `S1D1WL830`: Find NIRS channel using: Source `S1`, Detector `D1`, Wavelength `WL830`

EDIT: @rcassani

PR #757, opened 2025-01-07
Bugfix: On `channel_find` NIRS

Edouard2laire:
Fix a bug introduced in https://github.com/brainstorm-tools/brainstorm3/pull/709

Convidering the following channel find: 
image

channel_find would not work to find the channel called 'AUX1' (for trigger identification). 
This PR fix the bug. This quite crucial as it is affecting all the fNIRS pipeline.


rcassani:
@Edouard2laire, @tmedani, I'll be merging this later today.

PR #765, opened 2025-03-05
Replace java_dialog  with helpdlg for dislaying the Help in view leadfield vectors

tmedani:
Remove java_dialog and replace it with Matlab native helpdlg box to display the help to view the leadfield vectors. 
This modification will allow the user to interact with MATLAB/Brainstorm while the message box is open.


rcassani (commented):
It's a good idea. 

However, the proposed implementation completely disregard the formatting of the help, making it hard to read.

| Before    | After |
| -------- | ------- |
| image  | image    |

rcassani:
A better approach is now implemented in `master` with 0e0278ae
There a nonmodal version of `msgbox` is implemented, so it keeps the same look without blocking the use of other windows. Moreover, this nonmodal dialog box is now used in:

* View leadfield vectors, and 
* View leadfield sensitivity

tmedani:
Thanks @rcassani 
Having the non-modal option for the msgbox is better

PR #766, opened 2025-03-14
Bugfix file attrib

rcassani:
This PR has to goals:

1. Double-check the answer from `file_attrib(FILE, 'w')` in Windows, since there are reported false positives.

2. Allow Windows users to load protocols from folders that are symbolic links. Users would need to select either the `anat` or `data` folders inside the folder (which is the symbolic link).

@yashvakilna, @chinmaychinara91, please try this on different scenarios on Windows.

### DIFF
--- a/toolbox/gui/panel_protocol_editor.m
+++ b/toolbox/gui/panel_protocol_editor.m
@@ -305,6 +305,12 @@ function [subjectDir, studyDir, protocolName] = SelectProtocolDir(protocolDir)
         return
     end
 
+    % Check if 'anat' or 'data' dirs were seletect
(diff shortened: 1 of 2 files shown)


rcassani:
Ping @chinmaychinara91 @yashvakilna

yashvakilna:
I tested this on Windows 10 and 11 computers. Everything seems to be working well. Also, I was able to verify that the new `elseif ` statement in `file_attrib.m` is performing its function.

chinmaychinara91:
Thanks @rcassani .

yashvakilna:
Thanks @rcassani  @chinmaychinara91 !

PR #768, opened 2025-03-18
Allow to copy headmodel when there is no file in the destination folder

Edouard2laire:


Edouard2laire:
> I agree, copy to Studies without non-raw data (i.e., with imported data) is a strange limitation.
> 
> Though, the limitation should be to ignore Studies without a channel file. As the head-model is quite useless without a channel file. What do you think?

This seems reasonable.


> 
> P.S. Currently, this operation is possible in the Protocol tree with Ctrl-C / Ctrl-V


I am currently creating a process to simulate nirs data to evaluate the reconstruction algorithm so it's nice to copy the headmodel in the folder with the simulated data to be able to more easily launch the reconstruction. 

Basically the process is like this : 

```matlab 

    iStudy = db_add_condition(sInputs.SubjectName, output_condition);
    sStudy = bst_get('Study', iStudy);

    % Copy head model
    HeadModelFile   = sStudy.HeadModel(sStudy.iHeadModel).FileName;
    db_set_headmodel(HeadModelFile, iStudy);

    % Save channel definition
    [~, iChannelStudy] = bst_get('ChannelForStudy', iStudy);
    db_set_channel(iChannelStudy, ChannelMat, 2, 0);


   % Save Data 
    % ... 

```

rcassani:
@Edouard2laire, [e799f61](https://github.com/brainstorm-tools/brainstorm3/pull/768/commits/e799f61400676a2855ea7d368f9b0efb64e79dfa) should do the trick. Just set the channel file before

Edouard2laire:
Thx. its working great

PR #732, opened 2024-08-08
Refactor updated IEEG features (2/2)

rcassani:
# Refactor updated IEEG features (2/2)

This PR addresses pending issues (from PR #710) with the IEEG features.

## Pending issues

- [ ] Display menu (for channel file with SEEG) does something different to the stated
 https://neuroimage.usc.edu/forums/t/46083
- [ ] Avoid closing all figures when starting a new implantation. If figures need to be closed, this must be selective:
 (Issue 1 in [#701](https://github.com/brainstorm-tools/brainstorm3/pull/701))
- [ ] Some Java exception issues
 https://neuroimage.usc.edu/forums/t/46585
 https://neuroimage.usc.edu/forums/t/46178
- [ ] Thresholding of isosurface without creating a new one
 Creating of a new file was supposed to be temporary (added in [#689](https://github.com/brainstorm-tools/brainstorm3/pull/689))
 **Discussion:** Do we need this file in the DB? It will used only on localizing contacts.
- [ ] Printing contact locations in GUI, better text alignment 
 **Discussion:** Do we really need to print up to one-hundredth of millimeter? 
- [ ] Display sensors in MRI 3D slices, add isosurface to figure (surface tab), add new SEEG electrode (in IEEG tab), enable Select tool (in EEG tab), click on isosurface. Result: MRI viewer opens (why?), all electrodes disappear of GUI and Electrode configuration is disabled. Electrodes are visible in 3D figure.
- [ ] Wrong CT volume can be selected thresholding CT with slider. 
  See code: https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/gui/panel_surface.m#L422-L426
- [ ] Slider in panel surface for thresholding CT is set only at panel creation (`CreatePanel`), thus it's hardcoded to `1` to `4500`
  See code: https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/gui/panel_surface.m#L113 

@tmedani, @chinmaychinara91

PR #770, opened 2025-03-21
Fix IsoSurface threshold slider range

chinmaychinara91:
This PR fixes the issue below (point # 8 in PR https://github.com/brainstorm-tools/brainstorm3/pull/732)
> Slider in panel surface for thresholding CT is set only at panel creation (CreatePanel), thus it's hardcoded to 1 to 4500
> See code: https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/gui/panel_surface.m#L113

Now the slider range is updated in real time based on the IsoSurface.

### DIFF
--- a/toolbox/gui/panel_surface.m
+++ b/toolbox/gui/panel_surface.m
@@ -2745,7 +2745,10 @@ function [ctFile, isoValue, isoRange] = GetIsosurfaceParams(isosurfaceFile)
             isoValue = str2double(ctEntries{iEntries(end)}{1}{2});
             if any(cellfun(@isempty, ctEntries{iEntries(end)}{1}(3:4)))
                 % If range not in last History entry, load from CT file and update History accordingly
-                sCt = bst_memory('LoadMri', ctFile);
+                sCt = load(file_fullpath(ctFile), 'Histogram');
+                if ~isfield(sCt, 'Histogram')
+                    sCt = bst_memory('LoadMri', ctFile);
+                end
                 isoRange = double(round([sCt.Histogram.whiteLevel, sCt.Histogram.intensityMax]));
                 sSurf.History{iEntries(end), 3} = [sSurf.History{iEntries(end), 3} ' minVal = ' num2str(isoRange(1)) ' maxVal = ' num2str(isoRange(2))];
                 sSurfTmp.History = sSurf.History;
--- a/toolbox/gui/panel_surface.m
+++ b/toolbox/gui/panel_surface.m
@@ -2735,7 +2735,7 @@ function [ctFile, isoValue, isoRange] = GetIsosurfaceParams(isosurfaceFile)
     isoValue = [];
     isoRange = [];
     % Load the IsoSurface
-    sSurf = load(file_fullpath(isosurfaceFile));
+    sSurf = load(file_fullpath(isosurfaceFile), 'History');
     if isfield(sSurf, 'History') && ~isempty(sSurf.History)
         % Get CT file, value and range from last History entry
         ctEntries = regexp(sSurf.History(:, 3), '^Thresholded CT:\s(.*)\sthreshold\s*=\s*(\d+)(?:\sminVal\s*=\s*)?(\d+)?(?:\smaxVal\s*=\s*)?(\d+)?', 'tokens');
@@ -2751,8 +2751,7 @@ function [ctFile, isoValue, isoRange] = GetIsosurfaceParams(isosurfaceFile)
                 end
                 isoRange = double(round([sCt.Histogram.whiteLevel, sCt.Histogram.intensityMax]));
                 sSurf.History{iEntries(end), 3} = [sSurf.History{iEntries(end), 3} ' minVal = ' num2str(isoRange(1)) ' maxVal = ' num2str(isoRange(2))];
-                sSurfTmp.History = sSurf.History;
-                bst_save(file_fullpath(isosurfaceFile), sSurfTmp, [], 1);
+                bst_save(file_fullpath(isosurfaceFile), sSurf, [], 1);
             else
                 isoRange = str2double(ctEntries{iEntries(end)}{1}(3:4));
             end
--- a/toolbox/gui/panel_surface.m
+++ b/toolbox/gui/panel_surface.m
@@ -2747,7 +2747,7 @@ function [ctFile, isoValue, isoRange] = GetIsosurfaceParams(isosurfaceFile)
                 % If range not in last History entry, load from CT file and update History accordingly
                 sCt = bst_memory('LoadMri', ctFile);
                 isoRange = double(round([sCt.Histogram.whiteLevel, sCt.Histogram.intensityMax]));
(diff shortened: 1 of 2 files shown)


rcassani (changes requested):
This PR saves teh CT range in the IsoSurface history.

I works fine with isosurfaces created after these commits.
But it crashes with isosurfaces that were created before. It must have backwards compatibility.

toolbox/gui/panel_surface.m
```diff
@@ -2751,20 +2728,34 @@ function ApplyDefaultDisplay() %#ok<DEFNU>
...
     isoValue = [];
-    % Load the IsoSurface history
-    sSurf = load(file_fullpath(isosurfaceFile), 'History');
+    isoRange = [];
+    % Load the IsoSurface
+    sSurf = load(file_fullpath(isosurfaceFile));
     if isfield(sSurf, 'History') && ~isempty(sSurf.History)
         % Get CT file and value from last History entry
-        ctEntries  = regexp(sSurf.History(:, 3), '^Thresholded CT:\s(.*)\sthreshold\s*=\s*(\d+)', 'tokens');
+        ctEntries = regexp(sSurf.History(:, 3), '^Thresholded CT:\s(.*)\sthreshold\s*=\s*(\d+)(\sminVal\s*=\s*(\d+))?(\smaxVal\s*=\s*(\d+))?', 'tokens');
```
rcassani, line 2741: Why don't we get the values `minVal` and `maxVal` directly with the regular expression?

toolbox/gui/panel_surface.m
```diff
@@ -2751,20 +2728,34 @@ function ApplyDefaultDisplay() %#ok<DEFNU>
...
+                sCt = bst_memory('LoadMri', ctFile);
+                isoRange = [round(sCt.Histogram.whiteLevel) round(sCt.Histogram.intensityMax)];
+                sSurf = bst_history('add', sSurf, 'threshold_ct', ['Thresholded CT: ' sCt.FileName ' threshold = ' num2str(isoValue) ...
+                                                                   ' minVal = ' num2str(round(sCt.Histogram.whiteLevel)) ...
+                                                                   ' maxVal = ' num2str(round(sCt.Histogram.intensityMax))]);
+                bst_save(file_fullpath(isosurfaceFile), sSurf, 'v7');
+            else
+                minVal = str2double(strrep(ctEntries{iEntries(end)}{1}{3}, 'minVal = ', ''));
+                maxVal = str2double(strrep(ctEntries{iEntries(end)}{1}{4}, 'maxVal = ', ''));
+                isoRange = [minVal maxVal];
```
rcassani, line 2757: This string manipulation is not needed if parsing is done in the regular expression

toolbox/gui/panel_surface.m
```diff
@@ -2751,20 +2728,34 @@ function ApplyDefaultDisplay() %#ok<DEFNU>
...
         iEntries =  find(~cellfun(@isempty, ctEntries));
-        if any(iEntries) && length(ctEntries{iEntries(end)}{1}) == 2
+        if any(iEntries) && length(ctEntries{iEntries(end)}{1}) == 4
             ctFile   = ctEntries{iEntries(end)}{1}{1};
             isoValue = str2double(ctEntries{iEntries(end)}{1}{2});
+            if isempty(ctEntries{iEntries(end)}{1}{3}) || isempty(ctEntries{iEntries(end)}{1}{4})
+                % If IsoValue range not in IsoSurface History, load CT info and update History accordingly
+                sCt = bst_memory('LoadMri', ctFile);
+                isoRange = [round(sCt.Histogram.whiteLevel) round(sCt.Histogram.intensityMax)];
+                sSurf = bst_history('add', sSurf, 'threshold_ct', ['Thresholded CT: ' sCt.FileName ' threshold = ' num2str(isoValue) ...
```
rcassani, line 2750: Instead of creating a new history entry, edit the one that needs the range

toolbox/gui/panel_surface.m
rcassani: You could even use `textscan`

chinmaychinara91: Will use `textscan`

toolbox/gui/panel_surface.m
```diff
@@ -2743,7 +2743,7 @@ function ApplyDefaultDisplay() %#ok<DEFNU>
         if any(iEntries) && length(ctEntries{iEntries(end)}{1}) == 4
             ctFile   = ctEntries{iEntries(end)}{1}{1};
             isoValue = str2double(ctEntries{iEntries(end)}{1}{2});
-            if isempty(ctEntries{iEntries(end)}{1}{3}) || isempty(ctEntries{iEntries(end)}{1}{4})
+            if isempty(ctEntries{iEntries(end)}{1}{3}) && isempty(ctEntries{iEntries(end)}{1}{4})
```
rcassani: Bugfix? In which conditions this is a bug?

rcassani (commented):
I made changes to address some issues as the incorrect class of the variables in the CT file. Also changes to improve performance were made.

Test exhaustively the changes

toolbox/gui/panel_surface.m
rcassani: But the OR was correct, the range needs to be obtained if one OR the other is missing.
Not ONLY if both are missing.

toolbox/gui/panel_surface.m
```diff
@@ -2751,20 +2728,31 @@ function ApplyDefaultDisplay() %#ok<DEFNU>
...
+        % Get CT file, value and range from last History entry
+        ctEntries = regexp(sSurf.History(:, 3), '^Thresholded CT:\s(.*)\sthreshold\s*=\s*(\d+)(?:\sminVal\s*=\s*)?(\d+)?(?:\smaxVal\s*=\s*)?(\d+)?', 'tokens');
         iEntries =  find(~cellfun(@isempty, ctEntries));
-        if any(iEntries) && length(ctEntries{iEntries(end)}{1}) == 2
+        if any(iEntries) && length(ctEntries{iEntries(end)}{1}) == 4
             ctFile   = ctEntries{iEntries(end)}{1}{1};
             isoValue = str2double(ctEntries{iEntries(end)}{1}{2});
+            if any(cellfun(@isempty, ctEntries{iEntries(end)}{1}(3:4)))
+                % If range not in last History entry, load from CT file and update History accordingly
+                sCt = bst_memory('LoadMri', ctFile);
```
rcassani, line 2748: This can be done even faster.
Use `load()` to get the variable `Histogram`. If it does not exist, then we can `LoadMri` to compute it.

rcassani (changes requested):
Update to make it faster to the the CT range (`Histogram` field)

toolbox/gui/panel_surface.m
chinmaychinara91: The range calculation yes but for updating the history field in this case we need to do a check to ensure which range parameter needs to be updated in the text correct ?

rcassani (approved):
Looks good

chinmaychinara91:
Looks good to be merged.

PR #773, opened 2025-03-30
Add electrode with just 3D figure open

chinmaychinara91:
Does two things:

1. Can set electrodes directly from 3D Viz (without having to open MRI Viewer):
    - For SEEG > Display sensors in MRI 3D slices > add SEEG IsoSurface to figure (Surface tab) > Add electrode and enable Select tool (iEEG tab) > Select points in 3D IsoSurface and Set tip and Skull entry to create the electrode.
    - The steps work for ECoG as well (Either use [test data](https://drive.google.com/drive/folders/1v6IvXRbxUeawT-jB3VAJinrIzPxVPtMT?usp=sharing)) or just use the SEEG IsoSurface above to create dummy ECoG).
    - This gives option to the user to create electrodes directly in 3D without having to open MRI Viewer.
    
2. Commit 17cc586 fixes the issue below (point # 6 in #732)

> Display sensors in MRI 3D slices, add IsoSurface to figure (surface tab), add new SEEG electrode (in IEEG tab), enable Select tool (in EEG tab), click on isosurface. Result: MRI viewer opens (why?), all electrodes disappear of GUI and Electrode configuration is disabled. Electrodes are visible in 3D figure.

### DIFF
--- a/toolbox/gui/panel_coordinates.m
+++ b/toolbox/gui/panel_coordinates.m
@@ -431,8 +431,9 @@ function vi = SelectPoint(hFig, AcceptMri, isCentroid) %#ok<DEFNU>
          'Tag',             'ptCoordinates');
     % Update "Coordinates" panel
     UpdatePanel();
-    % Open MRI viewer for SEEG
-    if isCentroid
+    % Update MRI viewer (if open)
+    hFigMri = bst_figures('GetFiguresByType', {'MriViewer'});
+    if ~isempty(hFigMri)
         ViewInMriViewer();
     end
 end
--- a/toolbox/gui/panel_ieeg.m
+++ b/toolbox/gui/panel_ieeg.m
@@ -2905,41 +2905,20 @@ function SetElectrodeLoc(iLoc, jButton)
     global GlobalData;
 
     % Get selected electrodes
-    [sSelElec, iSelElec, iDS, iFig, hFig] = GetSelectedElectrodes();
-    MriIdx = 1;
-
+    [sSelElec, iSelElec, iDS, iFig] = GetSelectedElectrodes();
     if isempty(sSelElec)
-    	bst_error('No electrode seleced.', 'Set electrode position', 0);
+    	bst_error('No electrode selected.', 'Set electrode position', 0);
         return;
     elseif (length(sSelElec) > 1)
         bst_error('Multiple electrodes selected.', 'Set electrode position', 0);
         return;
-    elseif ~strcmpi(GlobalData.DataSet(iDS(MriIdx)).Figure(iFig(MriIdx)).Id.Type, 'MriViewer')
-        if length(hFig) == 1
-            bst_error('MRI viewer must be open', 'Set electrode position', 0);
-            return;
-        end
-        MriIdx = 2;
     elseif (size(sSelElec.Loc, 2) < iLoc-1)
         bst_error('Set the previous reference point (the tip) first.', 'Set electrode position', 0);
         return;
     end
-    
-    
-    sMri = panel_surface('GetSurfaceMri', hFig(MriIdx));
-    XYZ = figure_mri('GetLocation', 'scs', sMri, GlobalData.DataSet(iDS(MriIdx)).Figure(iFig(MriIdx)).Handles);
-
-    % If SCS coordinates are not available
+    % Get location from the current figure
+    XYZ = GetCrosshairLoc('scs');
     if isempty(XYZ)
-        % Ask to compute MNI transformation
-        isComputeMni = java_dialog('confirm', [...
-            'You need to define the NAS/LPA/RPA fiducial points before.' 10 ...
-            'Computing the MNI normalization would also define default fiducials.' 10 10 ...
-            'Compute the MNI normalization now?'], 'Set electrode position');
-        % Run computation
-        if isComputeMni
-            figure_mri('ComputeMniCoordinates', hFig);
-        end
         return;
     end
     % Make sure the points of the electrode are more than 1cm apart
@@ -3368,3 +3347,34 @@ function ExportChannelFile(isAtlas)
     end
 end
 
+%% ===== GET CROSSHAIR LOCATION FROM CURRENT FIGURE =====
+function XYZ = GetCrosshairLoc(cs)
+    global GlobalData;
+    % Intialize output
+    XYZ = [];
+    % Get figures
+    [hFig, iFig, iDS] = bst_figures('GetCurrentFigure');
+    switch lower(GlobalData.DataSet(iDS).Figure(iFig).Id.Type)
+        case '3dviz'
+            XYZ = figure_3d('GetLocation', cs, hFig);
+            if isempty(XYZ)
+                return;
+            end
+        case 'mriviewer'
+            sMri = panel_surface('GetSurfaceMri', hFig);
+            XYZ = figure_mri('GetLocation', cs, sMri, GlobalData.DataSet(iDS).Figure(iFig).Handles);
+            % If SCS coordinates are not available
+            if isempty(XYZ)
+                % Ask to compute MNI transformation
+                isComputeMni = java_dialog('confirm', [...
+                    'You need to define the NAS/LPA/RPA fiducial points before.' 10 ...
+                    'Computing the MNI normalization would also define default fiducials.' 10 10 ...
+                    'Compute the MNI normalization now?'], 'Get MRI crosshair location');
+                % Run computation
+                if isComputeMni
+                    figure_mri('ComputeMniCoordinates', hFig);
+                end
+                return;
+            end
+    end
+end
--- a/toolbox/gui/figure_3d.m
+++ b/toolbox/gui/figure_3d.m
@@ -4800,6 +4800,31 @@ function JumpMaximum(hFig)
     UpdateMriDisplay(hFig, [1 2 3], TessInfo, iAnatomy);
 end
 
+%% ===== GET LOCATION FROM 3D FIGURE =====
+function XYZ = GetLocation(cs, hFig)
+    XYZ = [];
+    CoordinatesSelector = getappdata(hFig, 'CoordinatesSelector');
+    isSelectingCoordinates = getappdata(hFig, 'isSelectingCoordinates');
+    % Exit early if required data is missing or selection is inactive
+    if isempty(CoordinatesSelector) || isempty(CoordinatesSelector.MRI) || ~isSelectingCoordinates
+        return;
+    end
+    % Determine which coordinate set to return based on input
+    switch lower(cs)
+        case 'mni'
+            XYZ = CoordinatesSelector.MNI;
+        case 'mri'
+            XYZ = CoordinatesSelector.MRI;
+        case 'scs'
+            XYZ = CoordinatesSelector.SCS;
+        case 'voxel'
+            XYZ = CoordinatesSelector.Voxel;
+        case 'world'
+            XYZ = CoordinatesSelector.World;
+        otherwise
+            bst_error('Invalid coordinate conversion', 'Get location (3D)');
+    end
+end
 
 %% ===== SET LOCATION MRI =====
 function SetLocationMri(hFig, cs, XYZ)


toolbox/gui/panel_coordinates.m
```diff
@@ -431,8 +431,9 @@ function SetSelectionState(isSelected)
          'Tag',             'ptCoordinates');
     % Update "Coordinates" panel
     UpdatePanel();
-    % Open MRI viewer for SEEG
-    if isCentroid
+    % Update MRI viewer (if open)
+    hFigMri = bst_figures('GetFiguresByType', {'MriViewer'});
```
rcassani, line 435: No need to be cell. Check the `bst_figures('GetFiguresByType')`

chinmaychinara91: Correct. My bad.

toolbox/gui/panel_ieeg.m
```diff
@@ -3361,3 +3341,34 @@ function ExportChannelFile(isAtlas)
...
+    % Intialize output
+    XYZ = [];
+    % Get figures
+    [hFig, iFig, iDS] = bst_figures('GetCurrentFigure');
+    switch lower(GlobalData.DataSet(iDS).Figure(iFig).Id.Type)
+        case '3dviz'
+            XYZ = figure_3d('GetLocation', cs, hFig);
+            if isempty(XYZ)
+                return;
+            end
```
rcassani, line 3356: Why is this needed?

toolbox/gui/panel_ieeg.m
```diff
@@ -3361,3 +3341,34 @@ function ExportChannelFile(isAtlas)
...
+                % Ask to compute MNI transformation
+                isComputeMni = java_dialog('confirm', [...
+                    'You need to define the NAS/LPA/RPA fiducial points before.' 10 ...
+                    'Computing the MNI normalization would also define default fiducials.' 10 10 ...
+                    'Compute the MNI normalization now?'], 'Get MRI crosshair location');
+                % Run computation
+                if isComputeMni
+                    figure_mri('ComputeMniCoordinates', hFig);
+                end
+                return;
```
rcassani, line 3371: Why?

toolbox/gui/panel_ieeg.m
```diff
@@ -3361,3 +3341,34 @@ function ExportChannelFile(isAtlas)
...
 end
 
+%% ===== GET CROSSHAIR LOCATION FROM CURRENT FIGURE =====
+function XYZ = GetCrosshairLoc(cs)
+    global GlobalData;
+    % Intialize output
+    XYZ = [];
+    % Get figures
+    [hFig, iFig, iDS] = bst_figures('GetCurrentFigure');
+    switch lower(GlobalData.DataSet(iDS).Figure(iFig).Id.Type)
```
rcassani, line 3351: There is no need to use `GlobalData` here. 
The Type of figure is the Figure Tag `hFig.Tag`

toolbox/gui/panel_ieeg.m
rcassani, line 3356: Why is this needed?

toolbox/gui/panel_ieeg.m
rcassani, line 3371: ~Why?~
What is the purpose of this line?

toolbox/gui/panel_coordinates.m
rcassani: Please pay more attention

rcassani (changes requested):
@chinmaychinara91 please address the questions on the review comments, and perform changes if you see they are needed after answering. 

Also, please check the recent bugfixes in `master`, and verify this PR behaves as expected

* c08a2a2dca450a029eef9080101b224990353767
* fff32d0ee75056e4327e7a79ba92eb58752749c8
* ad5e3c7c95858c5208237fdd272c58a7220b4df3

toolbox/gui/panel_ieeg.m
chinmaychinara91: Thanks for catching this. It is not required. It is repeating.

toolbox/gui/panel_ieeg.m
chinmaychinara91: After importing MRI, if user forgets to set the fiducials manually (or do MNI normalization in the beginning) so as to establish a coordinate system and then starts electrode implantation and tries setting tip/entry using the MRI viewer then this prompt allows the user to do so.

toolbox/gui/panel_ieeg.m
chinmaychinara91: Thanks for the tip.

toolbox/gui/panel_ieeg.m
rcassani: The comment is on the line `return;`

---

In other thoughts. If MNI normalization in done at this step, all the figures are updated?

If the user needs to close figures to see the change. Do not compute the MNI normalization here, just indicate it. 

Also, MNI normalization implies SCS, but not the opposite, right? In this train of thought, maybe we should allow to start iEEG implantation if there is not SCS for the Subject. This makes more sense, since all coordinates for the electrodes/contacts are in SCS, so it cannot be done without the fiducials.

chinmaychinara91: > The comment is on the line `return;`

It is not required. Will be careful.

> In other thoughts. If MNI normalization in done at this step, all the figures are updated?

Can you share an example case where there can be multiple figures here ?

>Also, MNI normalization implies SCS, but not the opposite, right?

Correct

> maybe we should allow to start iEEG implantation if there is not SCS for the Subject. This makes more sense, since all coordinates for the electrodes/contacts are in SCS, so it cannot be done without the fiducials.

I too feel that would be a better approach. Two options:
1. Show `SEEG/ECOG implantation` in the menu even if SCS is not there, give a warning to users that fiducials are not set and they need to compute it before proceeding.
2. Do not show `SEEG/ECOG implantation` in the menu if SCS is not there. But this I feel is not very intuitive for the users. There could be confusion as to why they cannot start an implantation when they already have MRI.

rcassani: > Can you share an example case where there can be multiple figures here ?

Two instances of MRIViewer

> I too feel that would be a better approach. Two options:

Then, why is not implemented in this PR that is your development?

> 1. Show SEEG/ECOG implantation in the menu even if SCS is not there, give a warning to users that fiducials are not set and they need to compute it before proceeding.
> 2. Do not show SEEG/ECOG implantation in the menu if SCS is not there. But this I feel is not very intuitive for the users. There could be confusion as to why they cannot start an implantation when they already have MRI.

Option 1 sound more logic.

chinmaychinara91: > Two instances of MRIViewer

No it updates only the figure it is run on. Other figures needs to be closed to see the change.

> Option 1 sound more logic.

Done. Check [6d6144a](https://github.com/brainstorm-tools/brainstorm3/pull/773/commits/6d6144aed162f0d88455dd05f53de978961ce626)  [1bc1474](https://github.com/brainstorm-tools/brainstorm3/pull/773/commits/1bc1474c8676bfbecad0f51b4087ec8624d98fc1). 

The `HasFiducials` function will come handy across Brainstorm.

toolbox/gui/panel_ieeg.m
```diff
@@ -3174,6 +3174,24 @@ function CreateImplantation(MriFile) %#ok<DEFNU>
...
+        cellfun(@(vol) bst_memory('UnloadMri', vol), MriFiles(iMriUnload));
+        switch(length(MriFiles))
+            case 1
+                bst_error('You need to set the fiducial points in the volume first.', 'SEEG/ECOG implantation', 0);
+                return;
+            case 2
+                bst_error('You need to co-register the volumes first.', 'SEEG/ECOG implantation', 0);
+                return;
+        end
+    end
```
rcassani, line 3194: There is some error in this logic.

If two volumes are provided, but only the first volume does not have SCS, both volumes are loaded, then only that one is unloaded, the other remains loaded.

Keep things simple and readable. There is no need for anonymous functions, nor `cellfun`. Just make a FOR loop on the volumes (it runs once if one volume is required)

```
FOR vol in vols
    LOAD vol
    CHECK vol_scs
    IF NOT vol_scs
        UNLOAD vol
        ERROR message depends if 1st or 2dn vol
             if error in 1st volume: Fiducials are needed.
             if error in 2nd volume: Coregister volumes
```

toolbox/gui/panel_ieeg.m
```diff
@@ -3174,6 +3174,24 @@ function CreateImplantation(MriFile) %#ok<DEFNU>
             MriFiles{2} = sSubject.Anatomy(iVol2).FileName;
         end
     end
+    % Check SCS coordinates availability for the MriFiles
+    hasValidSCS = @(sMri) isfield(sMri, 'SCS') && ~isempty(sMri.SCS) && all(isfield(sMri.SCS, {'NAS','LPA','RPA'})) && ~any(cellfun(@isempty, {sMri.SCS.NAS, sMri.SCS.LPA, sMri.SCS.RPA}));
+    % Pre-load all MRI files once
+    sMris = cellfun(@(vol) bst_memory('LoadMri', vol), MriFiles, 'UniformOutput', false);
```
rcassani, line 3180: Why to load both, and then check both?
This makes things way longer, if the error is in the first volume, user needs to wait for both volumes to be loaded.

toolbox/gui/panel_ieeg.m
```diff
@@ -3174,6 +3174,24 @@ function CreateImplantation(MriFile) %#ok<DEFNU>
...
+    % Find the index for volumes to unload
+    iMriUnload = find(~cellfun(hasValidSCS, sMris));
+    if ~isempty(iMriUnload)
+        % Unload from memory
+        cellfun(@(vol) bst_memory('UnloadMri', vol), MriFiles(iMriUnload));
+        switch(length(MriFiles))
+            case 1
+                bst_error('You need to set the fiducial points in the volume first.', 'SEEG/ECOG implantation', 0);
```
rcassani, line 3188: Which volume? Provide full information

toolbox/gui/panel_ieeg.m
```diff
@@ -3174,6 +3174,24 @@ function CreateImplantation(MriFile) %#ok<DEFNU>
...
+                return;
+            case 2
+                bst_error('You need to co-register the volumes first.', 'SEEG/ECOG implantation', 0);
```
rcassani, line 3191: Which volumes need to be coregistered?
Which volume on which volume?

rcassani (changes requested):

toolbox/gui/panel_ieeg.m
chinmaychinara91: > Keep things simple and readable.

Will do

toolbox/gui/panel_ieeg.m
chinmaychinara91: Got you. Makes sense.

toolbox/gui/panel_ieeg.m
chinmaychinara91: Updated.

toolbox/gui/panel_ieeg.m
```diff
@@ -3194,6 +3174,21 @@ function CreateImplantation(MriFile) %#ok<DEFNU>
...
+            switch(iVol)
+                case 1
+                    bst_error('You need to set the fiducial points in the MRI first.', 'SEEG/ECOG implantation', 0);
+                    return;
+                case 2
+                    bst_error(['You need to co-register ' MriFiles{2} ' to the MRI first.'], 'SEEG/ECOG implantation', 0);
+                    return;
+            end
+        end
+    end
```
rcassani, line 3191: Issues: 
- If error in Volume2, Volume1 remains in memory
- For Volume2 co-registration check goes beyond having defined fiducials

chinmaychinara91:
Thank you @rcassani. I tested it and it looks good to be merged.

PR #774, opened 2025-04-01
[connectivity] Remove regions without signal before connectivity

Edouard2laire:
Hello, 

This PR is here to remove regions where the signal is constant (zero) before running the connectivity. This is mainly the case for nirs since we will only have non zero value near the sensors. (it will probably be the same for SEEG)

Before: 

image

After: 

image

Note: I only tried with the following configuration, so It probably doesn't work in the general case, but i can improve the code if this kind of solution works for you. 

image

Edouard


Edouard2laire:
Closing as it will addressed in #806

PR #775, opened 2025-04-02
Allow to apply montage on raw data

Edouard2laire:
This is not ready to merge as we need to fix : 
image

After reloading the study the file appears well in the database.

### DIFF
--- a/toolbox/process/functions/process_montage_apply.m
+++ b/toolbox/process/functions/process_montage_apply.m
@@ -36,8 +36,8 @@ function sProcess = GetDescription() %#ok<DEFNU>
     sProcess.Index       = 307;
     sProcess.Description = 'https://neuroimage.usc.edu/brainstorm/Tutorials/MontageEditor';
     % Definition of the input accepted by this process
-    sProcess.InputTypes  = {'data'};
-    sProcess.OutputTypes = {'data'};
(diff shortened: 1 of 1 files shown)


Edouard2laire:
cc @ShahlaBakian

rcassani:
@yashvakilna, you're working on something similar right?
How is it going?

rcassani:
@Edouard2laire, this is good start.
The error seems to be registration in the DB of the new raw file.

However, applying montages on raw data must be done by blocks rather than the entire raw file. 
We'll be addressing on a different PR

Edouard2laire:
np. Do you think it is 'safe' for @ShahlaBakian to use this version of the fix in the mean time ?

rcassani:
@Edouard2laire, the PR for the block-by-block montage is already open: #776 
The code in #776 should work as the one in this PR, but using a block-by-block approach, and having the option for projectors.

> Do you think it is 'safe' for @ShahlaBakian to use this version of the fix in the mean time ?

It would be safe to use the version in this closed PR, but:

* It needs to be in the dir for USER process (`$HOME/.brainstorm/process`) to avoid deleting it with a Brainstorm update
* It will no have support

Edouard2laire:
Thanks for oppening the new PR :)

PR #776, opened 2025-04-03
Add support for `raw` files in process `Apply montage`

rcassani:
Expand existing process **Standardize > Apply montage** to support `raw` files

For `raw` data, there is the option to apply CTF compensation (only in CTF MEG devices) and SSP/ICA projectors before applying the montages.

If SSP/ICA projectors are not applied, they are removed from the output channel file, as the channels are with montages they are not the same. Although they could, however, in that case montages should not be the way to go, but filtering or projectors themselves. 

FYI @yashvakilna, @jcmosher, @Edouard2laire, @ShahlaBakian

### DIFF
--- a/toolbox/process/functions/process_montage_apply.m
+++ b/toolbox/process/functions/process_montage_apply.m
@@ -322,15 +322,23 @@ function OutputFiles = Run(sProcess, sInputs) %#ok<DEFNU>
                     % Write block
                     out_fwrite(sFileOut, ChannelMatOut, 1, iTimesBlocks(iBlock, :)-1, [], RawDataMat.F);
                 end
-                % Mark the projectors as already applied to the file
-                if isUseSsp && ~isempty(ChannelMatOut.Projector)
-                    for iProj = 1:length(ChannelMatOut.Projector)
-                        if (ChannelMatOut.Projector(iProj).Status == 1)
-                            ChannelMatOut.Projector(iProj).Status = 2;
+                % Update projectors data
+                if isUseSsp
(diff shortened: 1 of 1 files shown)


ShahlaBakian:
Hello Raymundo,

Thanks for your time and help. I have polysomnography data that is imported as "raw" and Im using " apply montage" to be able to separate the channels for further analysis. I unchecked CTF MEG and SSP/ICA and after using your last changes to process_montage_apply.m, I'm getting this error:
 
Screenshot 2025-04-03 at 7 55 28 PM

ShahlaBakian:
@Edouard2laire Thank you so much Edouard. your change solved the issue.

Edouard2laire:
thx @rcassani

ShahlaBakian:
Hello Raymundo,
I wanted to report a new issue with the "apply_montage" process that I just noticed.
regarding the issue I had for data that has been imported as "raw", the process " apply montage" after last update, now can be complete without error, however when I open the file the end of recording has been truncated and has 0 value only and I see this error message in command window:
please see the screen shot below:

 



*I have checked my raw data and I don't have this issue on the raw file. also when i import data that has been imported as "raw" and then i go recording> file> " import in database", then I apply montage this file doesnt have the 0 at the end.

rcassani:
@ShahlaBakian, thanks for reporting this issue.
Could you create a post in the Brainstorm forum?
If possible, provide a link to download a data sample that presents this issue.
Thanks!

https://neuroimage.usc.edu/forums/

rcassani:
ping @ShahlaBakian

PR #777, opened 2025-04-04
[bst_getunit]  fix crash when the value is 0

Edouard2laire:
Fix a small issue with getting the scale when the map is exactly 0.


rcassani:
@Edouard2laire, a clean up is needed in the functions `GetSIFactor` and `getUnit`
(unused variables, repeated code, a one-line function (`n2pAdjust`) used once, etc)

Is this something you can handle in this PR?

rcassani:
@Edouard2laire 🙏

Edouard2laire:
Done. Let me know if that is ok for you

PR #778, opened 2025-04-07
Bugfix: Panel Digitize

chinmaychinara91:
**Both panels:**
- [x] For 3D scanner, while manually collecting/correcting EEG points, enable `Random` button IFF all EEG points have been already collected/corrected. Check c177d50 2fc8adb
- [x] Right click on `texture > Digitize (3D scanner)`, get the correct subject ID from the subject name. Check 4b78658

**2024 panel:** 
- [x] For 3D scanner, after `Auto` detection, change delete button label and callback such that we can delete the last point. Check 3e0a5ae

### DIFF
--- a/toolbox/sensors/panel_digitize.m
+++ b/toolbox/sensors/panel_digitize.m
@@ -124,17 +124,16 @@ function Start(varargin) %#ok<DEFNU>
     end
     
     % ===== PATIENT ID =====
+    % Get Digitize options
+    DigitizeOptions = bst_get('DigitizeOptions');
     if isempty(sSubject)
-        % Get Digitize options
-        DigitizeOptions = bst_get('DigitizeOptions');
         % Ask for subject id
         PatientId = java_dialog('input', 'Please, enter subject ID:', Digitize.Type, [], DigitizeOptions.PatientId);
         if isempty(PatientId)
             return;
         end
         % Save the new default patient id
-        DigitizeOptions.PatientId = PatientId;
-        bst_set('DigitizeOptions', DigitizeOptions);
+        DigitizeOptions.PatientId = PatientId;      
     
         % ===== GET SUBJECT =====
         if strcmpi(Digitize.Type, '3DScanner')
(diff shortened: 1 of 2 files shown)


chinmaychinara91:
@Moo-Marc checking in to see if you were able to review this.

Moo-Marc:
I could test it probably sometime next week.  Our MEG electronics is currently being upgraded so the lab is a mess right now.

chinmaychinara91:
Sounds good. Thank you for the heads up.

Moo-Marc:
Hi team,

First, since I again won't be able to make it to the meeting this week, I wanted to point out that when we decided to keep a legacy version of panel_digitize, it was with the understanding that development would be restricted to the newer version and the legacy would be removed eventually.  We don't want to keep maintaining multiple versions; it's a strain on resources. So I highly recommend to limit updates on the legacy version to critical bug fixes only.  We might even want to consider reverting the legacy panel to the pre-3D-scanner state.  I don't see any advantage of having this new feature in both versions.

That said, I've tested this PR on the 2024 version with the Polhemus Fastrak.  Everything works as before this PR, so I have no objection to a merge.

Note that I've not yet had a chance to fully look at the 3d scanner code additions, and I'd like to because it caused a noticeable drop in responsiveness of the digitization with the Fastrak.  The first point takes a second longer to "click", and holding the stylus button pressed for continuous point collection (something I don't really do, but others might), gives slower and uneven point sampling in time.  It would be nice to find out what was changed that caused these delays when not using a 3d scanner.

Incidentally, I found another bug in the 2024 version, which existed in my original commit but changed behavior after the 3d-scanner feature was added.  I think it best to do a separate PR for that bug, but it has to do with the 3d points figure, when deleting the last point, after doing "add fiducials".  At first, it would not delete any fiducial point in the figure, and more recently, the additional fiducials are not displayed at all and it mistakenly deletes a head point in the figure.  The coordinates list and saved files behave correctly so it's only a visualization issue.  I'll try to get to debug it in the coming days/weeks.

PR #780, opened 2025-04-11
Toggle between centroid and surface point selection (for IsoSurface)

chinmaychinara91:
This PR allows users to switch between centroid/surface point selection while selecting a point in the IsoSurface (addresses point # 10 in PR #732).

> Shortcut to switch between center and non-center selection of 3D contacts

Proposed GUI (button) in iEEG panel :
Screenshot 2025-04-14 185728

Proposed menu item (checkboxmenuitem) when user right clicking on the 3D figure (shows up only when the `Select` button is pressed in iEEG panel and IsoSurface is loaded):
Screenshot 2025-04-14 185852

### DIFF
--- a/toolbox/gui/figure_3d.m
+++ b/toolbox/gui/figure_3d.m
@@ -1999,7 +1999,7 @@ function DisplayFigurePopup(hFig)
             isSelectingCoordinates = getappdata(hFig, 'isSelectingCoordinates');
             isIsoSurf = any(~cellfun(@isempty, regexp({TessInfo.SurfaceFile}, 'tess_isosurface', 'match')));
             if isIsoSurf && isSelectingCoordinates
-                jItem = gui_component('checkboxmenuitem', jPopup, [], 'Select centroid', [], [], @(h,ev)panel_coordinates('SetCentroidSelection', ev.getSource.isSelected()));
+                jItem = gui_component('checkboxmenuitem', jPopup, [], 'Select surface centroid', [], [], @(h,ev)panel_coordinates('SetCentroidSelection', ev.getSource.isSelected()));
                 isSelectingCentroid = getappdata(hFig, 'isSelectingCentroid');
                 jItem.setSelected(isSelectingCentroid);
                 jPopup.addSeparator();
--- a/toolbox/gui/figure_3d.m
+++ b/toolbox/gui/figure_3d.m
@@ -1997,8 +1997,8 @@ function DisplayFigurePopup(hFig)
         % ==== MENU: TOGGLE BETWEEN CENTROID/SURFACE POINT SELECTION ====
         if gui_brainstorm('isTabVisible', 'iEEG')
             isSelectingCoordinates = getappdata(hFig, 'isSelectingCoordinates');
-            iIsoSurf = find(cellfun(@(x) ~isempty(regexp(x, 'tess_isosurface', 'match')), {TessInfo.SurfaceFile})); 
-            if ~isempty(iIsoSurf) && isSelectingCoordinates
+            isIsoSurf = any(~cellfun(@isempty, regexp({TessInfo.SurfaceFile}, 'tess_isosurface', 'match')));
+            if isIsoSurf && isSelectingCoordinates
                 jItem = gui_component('checkboxmenuitem', jPopup, [], 'Select centroid', [], [], @(h,ev)panel_coordinates('SetCentroidSelection', ev.getSource.isSelected()));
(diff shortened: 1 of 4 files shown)


toolbox/gui/figure_3d.m
```diff
@@ -1998,6 +1996,17 @@ function DisplayFigurePopup(hFig)
         jItem = gui_component('checkboxmenuitem', jPopup, [], 'Get coordinates...', IconLoader.ICON_SCOUT_NEW, [], @GetCoordinates);
         jItem.setSelected(panel_coordinates('GetSelectionState'));
         jItem.setAccelerator(KeyStroke.getKeyStroke(KeyEvent.VK_P, KeyEvent.CTRL_MASK));
+        % ==== MENU: TOGGLE BETWEEN CENTROID/SURFACE POINT SELECTION ====
+        if gui_brainstorm('isTabVisible', 'iEEG')
+            ctrl = bst_get('PanelControls', 'iEEG');
+            if ctrl.jButtonSelect.isSelected()
+                jPopup.addSeparator();
```
rcassani, line 2003: This separator does not make sense. The previous item is "Get coordinates" which is strongly related.

toolbox/gui/figure_3d.m
```diff
@@ -1998,6 +1996,17 @@ function DisplayFigurePopup(hFig)
         jItem = gui_component('checkboxmenuitem', jPopup, [], 'Get coordinates...', IconLoader.ICON_SCOUT_NEW, [], @GetCoordinates);
         jItem.setSelected(panel_coordinates('GetSelectionState'));
         jItem.setAccelerator(KeyStroke.getKeyStroke(KeyEvent.VK_P, KeyEvent.CTRL_MASK));
+        % ==== MENU: TOGGLE BETWEEN CENTROID/SURFACE POINT SELECTION ====
```
rcassani, line 1999: This option is available even without iEEG channel information.
Is this the intended behaviour?

1. Close all figures
2. Open cortex surface
3. Add iEEG tab (will be all gray)
4. Add Coordinates tab
5. Click "Select" on coordinate tab
6. Right-click on figure. The menu Select Centroid appears.

Even if that behaviour is intended, it is not consisted.
Just changing the order of actions give something different. 

1. Close all figures
2. Open cortex surface
3. Add Coordinates tab
4. Click "Select" on coordinate tab
5. Right-click on figure. Select centroid menu is not there.
5. Add iEEG tab (will be all gray)
6. Right-click on figure. Select centroid menu is not there.
7. In Coordinate tab, turn OFF and ON the Select button
8. Right-click on figure. Now the Select centroid menu is there.

rcassani (changes requested):
Not very intuitive behaviour.

1. How to turn off Centroid selection? From iEEG tab without turning off and on the coordinate selection?

2. There is no indication in iEEG tab that the select centroid is selected or not

3. "Select centroid" option is available (tab and context menu) if no IsoSurface is present in the Figure. Is this expected?

4. For the iEEG tab button. Did you consider other alternatives, for example another button with a different icon (⨁, or similar) to make the behaviour explicit? As you know we have this kind of behaviour in the Scout tab, when we can select either [ALL] or [SEL] or none of them, but not both at the same time.

---
Minor:

1. Do not capitalize "Centroid" in "Select Centroid"

toolbox/gui/figure_3d.m
chinmaychinara91: Corrected

chinmaychinara91:
> 4. For the iEEG tab button. Did you consider other alternatives, for example another button with a different icon (⨁, or similar) >to make the behaviour explicit? As you know we have this kind of behaviour in the Scout tab, when we can select either [ALL] or >[SEL] or none of them, but not both at the same time.

- Added a separate button as you suggested and that definitely makes things more clear. 
- For now I just added a filler icon for the centroid selection (iconReset). Can you create a icon like you suggested and change it ?

> 1. How to turn off Centroid selection? From iEEG tab without turning off and on the coordinate selection?
> 2. There is no indication in iEEG tab that the select centroid is selected or not
> 3. "Select centroid" option is available (tab and context menu) if no IsoSurface is present in the Figure. Is this expected?

I agree these were not properly handled earlier but now they should be

> Do not capitalize "Centroid" in "Select Centroid"

Updated

I have updated the description with the new GUI changes

rcassani (approved):
I made minimal changes

chinmaychinara91:
Thank you @rcassani. Love the idea of the colorful icon.

PR #781, opened 2025-04-11
Fix reload last pipeline

Edouard2laire:
This fix a bug introduced by https://github.com/brainstorm-tools/brainstorm3/commit/d0aac8311fcbd70bf99eece9cd2e49a84be5e144

in line 1410, 
    FileNames = GetFilesList(Reports{iStart(1),3}, 0);

Report is 
image

and without the fix, iStart is [ 0, 1, 0, 0] instead of [2] so  iStart(1) is using the first entry which is the login of the version instead of the first called process.

### DIFF
--- a/toolbox/process/bst_report.m
+++ b/toolbox/process/bst_report.m
@@ -1407,8 +1407,8 @@ function Recall(target)
         return
     end
     % Get the start and process entries
-    iStart   = strcmpi(Reports(:,1), 'start');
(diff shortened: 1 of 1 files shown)


rcassani:
@Edouard2laire, I added directly in `master` with other fix: 22572529

rcassani:
Thanks for reporting this issue

PR #785, opened 2025-04-22
Improve iEEG editing features (`Group electrodes` and `Add/Remove contacts`)

chinmaychinara91:
This PR adds some new features that will enable iEEG **(currently focused only on SEEG)** editing as under:
- [x] **Group electrodes (SEEG only)**
   - **Testing:**
     - Option-1: Select electrodes in panel iEEG > right click on the electrodes > `Group Electrodes` 
     - Option-2: Select electrodes in panel iEEG > press `Ctrl+G`
Screenshot 2025-04-21 174535
   - This feature will be more useful in case of automated detection using GARDEL where there could be some bad detections of contacts being grouped into different electrodes even though they should be part of just one.
   
- [x] **Add contacts (SEEG only)**
   - **Testing:**
     - Option-1: Open 3DViz from SEEG channel > Add IsoSurface > Select electrode in panel iEEG > Turn on `Select Surface Point` > Select a location on the IsoSurface > click `Contacts` (menu) > `Add contact` in iEEG panel
     - Option-2: Open 3DViz of SEEG channel > Add IsoSurface > Select electrode in panel iEEG > Turn on `Select Surface Point` > Select a location on the IsoSurface > Right click (`Channels > Add contact`)
     - Option-3: Open MRI Viewer from SEEG channel > Select electrode in panel iEEG > Set crosshair at a location in MRI Viewer > click `Contacts` (menu) > `Add contact` in iEEG panel
     - Option-4: Open MRI Viewer from SEEG channel > Select electrode in panel iEEG > Set crosshair at a location in MRI Viewer > Right click (`Electrodes> Add contact`)  
     
- [x] **Remove contacts (SEEG only)**
   - **Testing:**
     - Option-1: Select contact(s) in panel iEEG > press `Delete` or `Backspace`
     - Option-2: Select contact(s) in panel iEEG > click `Contacts` (menu) > `Remove selected contacts`
     - Option-3: Select the contact in 3D figure >  press `Delete` or `Backspace`
     - Option-4: Select the contact in 3D figure > right click (`Channels > Remove contact`)

- Panel iEEG
Screenshot 2025-04-21 184412
- 3D figure
Screenshot 2025-04-21 184513
 - MRI Viewer
Screenshot 2025-04-21 184557

- **Addresses # 11 and # 12 in #732** 
>Delete and add contacts while maintaining the ordering

>Regrouping contacts correctly for wrong grouping of contacts for a electrode

Some points for discussion @tmedani @rcassani:
1. I have not added the remove contacts feature from MRI viewer.  Do you think it is worth having something like a drop-down list approach we have for `Set electrode position` ?
Screenshot 2025-04-21 175358

### DIFF
--- a/toolbox/gui/figure_3d.m
+++ b/toolbox/gui/figure_3d.m
@@ -1601,7 +1601,9 @@ function DisplayFigurePopup(hFig)
     else
         TfFile = [];
     end
-
+    % Check if IsoSurface exists in the figure
+    isIsoSurf = any(~cellfun(@isempty, regexp({TessInfo.SurfaceFile}, 'tess_isosurface', 'match')));
+    
     % Create popup menu
     jPopup = java_create('javax.swing.JPopupMenu');
     
@@ -1816,7 +1818,6 @@ function DisplayFigurePopup(hFig)
             jMenuChannels.addSeparator();
             gui_component('MenuItem', jMenuChannels, [], 'Configure display', IconLoader.ICON_CHANNEL, [], @(h,ev)SetElectrodesConfig(hFig));
             % For iEEG: Add/Remove contact(s) (TODO: support for ECoG)
-            isIsoSurf = any(~cellfun(@isempty, regexp({TessInfo.SurfaceFile}, 'tess_isosurface', 'match')));
             sSelElec = panel_ieeg('GetSelectedElectrodes');
             if ~isempty(sSelElec) && strcmpi(sSelElec(end).Type, 'SEEG')
                 if isIsoSurf
@@ -2022,7 +2023,6 @@ function DisplayFigurePopup(hFig)
         % ==== MENU: TOGGLE BETWEEN CENTROID/SURFACE POINT SELECTION ====
         if gui_brainstorm('isTabVisible', 'iEEG')
             isSelectingCoordinates = getappdata(hFig, 'isSelectingCoordinates');
-            isIsoSurf = any(~cellfun(@isempty, regexp({TessInfo.SurfaceFile}, 'tess_isosurface', 'match')));
             if isIsoSurf && isSelectingCoordinates
                 jItem = gui_component('checkboxmenuitem', jPopup, [], 'Select surface centroid', [], [], @(h,ev)panel_coordinates('SetCentroidSelection', ev.getSource.isSelected()));
                 isSelectingCentroid = getappdata(hFig, 'isSelectingCentroid');
--- a/toolbox/gui/figure_3d.m
+++ b/toolbox/gui/figure_3d.m
@@ -1807,14 +1807,6 @@ function DisplayFigurePopup(hFig)
             end
         % 3DElectrodes
         elseif ~isempty(hElectrodeGrid)
-            % For iEEG: Remove contact(s) (TODO: support for ECoG)
-            if gui_brainstorm('isTabVisible', 'iEEG')
-                if ~isempty(SelChan) && ~isempty(Modality) && ismember(Modality, {'SEEG'})
-                    jItem = gui_component('MenuItem', jMenuChannels, [], 'Remove selected contact(s)', IconLoader.ICON_MINUS, [], @(h,ev)panel_ieeg('RemoveContactHelper'));
-                    jItem.setAccelerator(KeyStroke.getKeyStroke(KeyEvent.VK_DELETE, 0));
-                    jMenuChannels.addSeparator();
-                end
-            end
             % Menu "View sensor labels"
             isLabels = ~isempty(GlobalData.DataSet(iDS).Figure(iFig).Handles.hSensorLabels);
             jItem = gui_component('CheckBoxMenuItem', jMenuChannels, [], 'Display labels', IconLoader.ICON_CHANNEL_LABEL, [], @(h,ev)ViewSensors(hFig, [], ~isLabels));
@@ -1823,6 +1815,18 @@ function DisplayFigurePopup(hFig)
             % Configure 3D electrode display
             jMenuChannels.addSeparator();
             gui_component('MenuItem', jMenuChannels, [], 'Configure display', IconLoader.ICON_CHANNEL, [], @(h,ev)SetElectrodesConfig(hFig));
+            % For iEEG: Add/Remove contact(s) (TODO: support for ECoG)
+            isIsoSurf = any(~cellfun(@isempty, regexp({TessInfo.SurfaceFile}, 'tess_isosurface', 'match')));
+            sSelElec = panel_ieeg('GetSelectedElectrodes');
+            if ~isempty(sSelElec) && strcmpi(sSelElec(end).Type, 'SEEG')
+                if isIsoSurf
+                    gui_component('MenuItem', jMenuChannels, [], 'Add contact', IconLoader.ICON_PLUS, [], @(h,ev)panel_ieeg('AddContact'));
+                end
+                if ~isempty(SelChan)
+                    jItem = gui_component('MenuItem', jMenuChannels, [], 'Remove selected contact(s)', IconLoader.ICON_MINUS, [], @(h,ev)panel_ieeg('RemoveContactHelper'));
+                    jItem.setAccelerator(KeyStroke.getKeyStroke(KeyEvent.VK_DELETE, 0));
+                end
+            end
         % Other figures
         else
             % Menu "View sensors"
--- a/toolbox/gui/figure_3d.m
+++ b/toolbox/gui/figure_3d.m
@@ -1293,7 +1293,7 @@ function FigureKeyPressedCallback(hFig, keyEvent)
                     % For iEEG: Remove contact(s) (TODO: support for ECoG)
                     if gui_brainstorm('isTabVisible', 'iEEG')
                         if ~isempty(SelChan) && ~isempty(FigureId.Modality) && ismember(FigureId.Modality, {'SEEG'})
-                            panel_ieeg('RemoveContact');
+                            panel_ieeg('RemoveContactHelper');
                         end
                     end
                 % ESCAPE: RESET SELECTION
@@ -1810,7 +1810,7 @@ function DisplayFigurePopup(hFig)
             % For iEEG: Remove contact(s) (TODO: support for ECoG)
             if gui_brainstorm('isTabVisible', 'iEEG')
                 if ~isempty(SelChan) && ~isempty(Modality) && ismember(Modality, {'SEEG'})
-                    jItem = gui_component('MenuItem', jMenuChannels, [], 'Remove selected contact(s)', IconLoader.ICON_MINUS, [], @(h,ev)panel_ieeg('RemoveContact'));
+                    jItem = gui_component('MenuItem', jMenuChannels, [], 'Remove selected contact(s)', IconLoader.ICON_MINUS, [], @(h,ev)panel_ieeg('RemoveContactHelper'));
                     jItem.setAccelerator(KeyStroke.getKeyStroke(KeyEvent.VK_DELETE, 0));
                     jMenuChannels.addSeparator();
                 end
--- a/toolbox/gui/figure_3d.m
+++ b/toolbox/gui/figure_3d.m
@@ -1264,8 +1264,9 @@ function FigureKeyPressedCallback(hFig, keyEvent)
                             view_timefreq(TfInfo.FileName, 'SingleSensor', SelChan{1});
                         end
                     end
-                % DELETE: SET CHANNELS AS BAD
+                % DELETE
                 case {'delete', 'backspace'}
+                    % Set channels as bad
                     isMulti2dLayout = (isfield(GlobalData.DataSet(iDS).Figure(iFig).Handles, 'hLines') && (length(GlobalData.DataSet(iDS).Figure(iFig).Handles.hLines) >= 2));
                     if ~isAlignFig && ~isempty(SelChan) && ~isSensorsOnly && ~isempty(GlobalData.DataSet(iDS).DataFile) && ...
                             (length(GlobalData.DataSet(iDS).Figure(iFig).SelectedChannels) ~= length(iSelChan)) && ~isMulti2dLayout
@@ -1282,6 +1283,12 @@ function FigureKeyPressedCallback(hFig, keyEvent)
                         panel_channel_editor('UpdateChannelFlag', GlobalData.DataSet(iDS).DataFile, newChannelFlag);
                         % Reset selection
                         bst_figures('SetSelectedRows', []);
+                    end
+                    % For iEEG: Remove contact(s) (TODO: support for ECoG)
+                    if gui_brainstorm('isTabVisible', 'iEEG')
+                        if ~isempty(SelChan) && ~isempty(FigureId.Modality) && ismember(FigureId.Modality, {'SEEG'})
+                            panel_ieeg('RemoveContact');
+                        end
                     end
                 % ESCAPE: RESET SELECTION
                 case 'escape'
@@ -1794,6 +1801,14 @@ function DisplayFigurePopup(hFig)
             end
         % 3DElectrodes
         elseif ~isempty(hElectrodeGrid)
+            % For iEEG: Remove contact(s) (TODO: support for ECoG)
+            if gui_brainstorm('isTabVisible', 'iEEG')
+                if ~isempty(SelChan) && ~isempty(Modality) && ismember(Modality, {'SEEG'})
+                    jItem = gui_component('MenuItem', jMenuChannels, [], 'Remove selected contact(s)', IconLoader.ICON_MINUS, [], @(h,ev)panel_ieeg('RemoveContact'));
+                    jItem.setAccelerator(KeyStroke.getKeyStroke(KeyEvent.VK_DELETE, 0));
+                    jMenuChannels.addSeparator();
+                end
+            end
             % Menu "View sensor labels"
             isLabels = ~isempty(GlobalData.DataSet(iDS).Figure(iFig).Handles.hSensorLabels);
             jItem = gui_component('CheckBoxMenuItem', jMenuChannels, [], 'Display labels', IconLoader.ICON_CHANNEL_LABEL, [], @(h,ev)ViewSensors(hFig, [], ~isLabels));
--- a/toolbox/gui/figure_3d.m
+++ b/toolbox/gui/figure_3d.m
@@ -1225,6 +1225,15 @@ function FigureKeyPressedCallback(hFig, keyEvent)
                     if ismember('control', keyEvent.Modifier)
                         bst_figures('ViewResults', hFig); 
                     end
+                    % For iEEG: Add contact
+                    if gui_brainstorm('isTabVisible', 'iEEG')
+                        TessInfo = getappdata(hFig, 'Surface');
+                        isIsoSurf = any(~cellfun(@isempty, regexp({TessInfo.SurfaceFile}, 'tess_isosurface', 'match'))); 
+                        sSelElec = panel_ieeg('GetSelectedElectrodes');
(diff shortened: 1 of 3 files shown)


rcassani (changes requested):
@chinmaychinara91, further testing was needed
- There was trouble with the vector orientation of the contact coordinates, that changed whether the centroid was computed or not.
- Merge a context menu does not make sense, that is the only option like that. This kind of options should be consistent with what we already have. E.g merging events, merging scouts.

I will commit the changes directly
-

toolbox/gui/panel_ieeg.m
```diff
@@ -1391,9 +1461,80 @@ function AddElectrode()
...
+    % Get contacts for this electrode
+    sContacts = GetContacts(sSelElec.Name);
+    % Add new channel data for the contact
+    sChannel = db_template('channeldesc');
+    sChannel.Name = sprintf('%s%d', sSelElec(end).Name, size(sContacts, 2)+1);
+    sChannel.Type = 'SEEG';
+    sChannel.Group = sSelElec(end).Name;       
+    Channels = [Channels(1:iChan(end))'; sChannel; Channels(iChan(end)+1:end)'];
+    iChan(end+1) = iChan(end)+1;
+    Channels(iChan(end)).Loc = XYZ';
```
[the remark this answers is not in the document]
chinmaychinara91: My apologies. Should have checked it better.

toolbox/gui/panel_ieeg.m
```diff
@@ -1391,9 +1461,80 @@ function AddElectrode()
...
+        for iFig = iFigall(iDSall == iDS)
+            GlobalData.DataSet(iDS).Figure(iFig).SelectedChannels = [GlobalData.DataSet(iDS).Figure(iFig).SelectedChannels, iChan];
+        end
+    end
+    % Update intraelectrode structure in channel
+    sContactsLoc = [GlobalData.DataSet(iDSall(1)).Channel(iChan).Loc];
+    % Sort contacts (distance from origin)
+    sContactsLocSorted = GetSortedContacts(sContactsLoc);
+    % Update electrode contact number
+    sSelElec(end).ContactNumber = size(sContactsLocSorted, 2);
```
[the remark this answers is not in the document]
chinmaychinara91: Did consider `length` first but `size` makes more sense here. If number of contacts is < 3 then length gives wrong count as it will always give 3.

toolbox/gui/panel_ieeg.m
```diff
@@ -1391,9 +1461,80 @@ function AddElectrode()
...
+    end
+    % Get channel data
+    Channels = GlobalData.DataSet(iDSall(1)).Channel;
+    iChan = find(strcmpi({Channels.Group}, sSelElec(end).Name));
+    if isempty(iChan)
+        java_dialog('warning', ['Set the tip and skull entry for electrode "' sSelElec(end).Name '"'], 'Add contact');
+        return;
+    end
+    % Get crosshair location from figure
+    XYZ = GetCrosshairLoc('scs');
```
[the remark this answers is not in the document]
chinmaychinara91: My apologies. Should have checked it better. The root cause was what you pointed above the vector orientation of the contact coordinates.

chinmaychinara91:
> There was trouble with the vector orientation of the contact coordinates, that changed whether the centroid was computed or not.

Thank you for this. My apologies for not having done a better root cause analysis. 
  
> Merge a context menu does not make sense, that is the only option like that. This kind of options should be consistent with what we already have. E.g merging events, merging scouts.

I see what you mean now. One other option I thought of was to add a merge icon on the top list of icons in the iEEG panel but to avoid crowding there I went with context menu.

toolbox/gui/panel_ieeg.m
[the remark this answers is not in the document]
chinmaychinara91: > The naming of the variable sContactsLocSorted is misleading, it suggest that it is a struct array with sContacts (sContactsLocSorted).
> 
> Please use meaningful variable names, naming convention is explicit in the Scripting tutorial:
> https://neuroimage.usc.edu/brainstorm/Tutorials/Scripting#Naming_conventions

Apologies for the confusion. Will be careful.

toolbox/gui/panel_ieeg.m
```diff
@@ -1391,9 +1461,80 @@ function AddElectrode()
...
+        return;
+    end
+    % Get contacts for this electrode
+    sContacts = GetContacts(sSelElec.Name);
+    % Add new channel data for the contact
+    sChannel = db_template('channeldesc');
+    sChannel.Name = sprintf('%s%d', sSelElec(end).Name, size(sContacts, 2)+1);
+    sChannel.Type = 'SEEG';
+    sChannel.Group = sSelElec(end).Name;       
+    Channels = [Channels(1:iChan(end))'; sChannel; Channels(iChan(end)+1:end)'];
```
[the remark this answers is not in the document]
rcassani: This Matlab fundamentals:

```
a = rand(1, 10);
b = rand(1,5);
c = 1;

d = [a, c, b];    % Easy to read
e = [a'; c; b']'; % Not easy, nor clean, nor simple

isequal(d, e)
```

```
ans =

  logical

   1
```

toolbox/gui/panel_ieeg.m
```diff
@@ -1482,6 +1637,186 @@ function RemoveElectrode()
...
+    global GlobalData;
+    % Get selected electrode
+    [sSelElec, iSelElec, iDSall, iFigall] = GetSelectedElectrodes();
+    if isempty(iDSall)
+        return;
+    end
+    % Check if this is an new implantation folder
+    ChannelFile = GlobalData.DataSet(iDSall(1)).ChannelFile;
+    [~, folderName] = bst_fileparts(bst_fileparts(ChannelFile));
+    isImplantation = ~isempty(strfind(folderName, 'Implantation'));
```
rcassani, line 1652: Should not then only allow `AddContact()` in the Implantation study? 
(Do not make the change, this is a question)

toolbox/gui/panel_ieeg.m
```diff
@@ -1482,6 +1637,186 @@ function RemoveElectrode()
...
+    end
+    sSelCont = GetSelectedContacts();
+    % If every contact of that electrode is selected, remove the whole electrode
+    if numel(sSelCont) == sSelElec(end).ContactNumber
+        RemoveElectrode();
+    else
+        % Otherwise just remove the highlighted contacts
+        RemoveContact();
+    end
+end
```
rcassani, line 1769: Why aren't these checks inside `RemoveContact`?

toolbox/gui/panel_ieeg.m
```diff
@@ -2557,9 +2893,12 @@ function SetDisplayMode(DisplayMode)
 
 
 %% ===== ALIGN CONTACTS =====
-function Channels = AlignContacts(iDS, iFig, Method, sElectrodes, Channels, isUpdate, isProjectEcog)
+function Channels = AlignContacts(iDS, iFig, Method, sElectrodes, Channels, isUpdate, isProjectEcog, sContactsLoc)
```
rcassani, line 2897: The `auto` options for this process does not align the locations, it just adds the locations to the channel.

Why is that option in here? 

Moreover, that `auto` option only works if `sElectrodes` is just one electrode. This is not the case of all the other options.

Please add code where it belongs.

toolbox/gui/figure_mri.m
```diff
@@ -625,6 +625,13 @@ function FigureKeyPress_Callback(hFig, keyEvent)
                     if ismember('control', keyEvent.Modifier)
                         SetElectrodePosition(hFig);
                     end
+                    % For iEEG: Add contact (TODO: support for ECoG)
```
rcassani, line 628: Is it going to be really added the ECoG case? 
Or it is just not supported?
Does it make sense to add add/remove contacts for ECOG?

toolbox/gui/figure_mri.m
rcassani, line 1: Modification in this files contain checks that doe not belong in here, they belong to the `AddContact` function in `panel_ieeg`

toolbox/gui/figure_3d.m
rcassani, line 1: @chinmaychinara91  Modify this file to do the check where it is necessary. See [97adc71](https://github.com/brainstorm-tools/brainstorm3/pull/785/commits/97adc716512d5d4a887a6f9ed490f39fd9630747) as example

Please do not do checks that are not necessary here.

toolbox/gui/panel_ieeg.m
```diff
@@ -948,8 +985,29 @@ function SetSelectedContacts(iSelCont)
     SetMriCrosshair(sContacts);
 end
 
+%% ===== ELECTRODE LIST PANEL POP-UP MENU =====
```
rcassani, line 988: I do not think we have popup menus in lists.

Why don't we use the same approach as in clusters, events and scouts, and have the action in a button or menu?


The button has been added already.

rcassani (changes requested):
@chinmaychinara91, I made some changes directly in the branch.

Please check the commits and understand why the changes were made.
There are few changes that are requested. Read carefully the comments.

In general, it looks good.
- There were some bugs
- There are questions to define (e.g., popup menu)

I think adding and removing contacts should be only allowed in the Implantation study. What do you think @tmedani?
We can go even further not allowing to add recordings to the Implantation study.

toolbox/gui/panel_ieeg.m
```diff
@@ -1329,10 +1392,13 @@ function SetElectrodeVisible(isVisible)
     GlobalData.DataSet(iDSall(1)).isChannelModified = 1;
 end
 
-
 %% ===== ADD ELECTRODE =====
-function AddElectrode()
+function AddElectrode(elecName)
```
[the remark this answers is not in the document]
rcassani: Done in [5da2a11](https://github.com/brainstorm-tools/brainstorm3/pull/785/commits/5da2a1170bee9c4f86ee8edee7b8a64e1c525df2)

toolbox/gui/panel_ieeg.m
chinmaychinara91: Agreed. This makes sense.

toolbox/gui/panel_ieeg.m
chinmaychinara91: Wanted to keep the `RemoveContact` feature only about removing contacts. Since there is a case where removing all contacts is equivalent to removing the whole electrode `RemoveElectrode` so wanted to keep it separated from the main remove contact function. Hence the helper function.

toolbox/gui/figure_mri.m
chinmaychinara91: > Is it going to be really added the ECoG case?
Or it is just not supported?
Does it make sense to add add/remove contacts for ECOG?

What I meant by "support for ECoG" in the comments was "adding" the feature for ECoG. Sorry for not being more clear. 

The way Brainstorm defines creation of ECoG electrodes, I don't think we need this feature. 

But if we ever go in the route of adding automated detection of ECoG (some workshop participants were asking about it and there are some open-source literature and tools available that do it which I have not exhaustively tested) then there could be a potential use case there where having add/remove contacts as post processing bad detections could come in handy. This is the only use case that I can see now as a future prospect of adding these features for ECoG. 

@tmedani what do you think ?

toolbox/gui/figure_mri.m
chinmaychinara91: Got you. Have noted them.

toolbox/gui/figure_3d.m
chinmaychinara91: Noted. Will do the modifications.

toolbox/gui/panel_ieeg.m
chinmaychinara91: Thanks for adding the button.

> I do not think we have popup menus in lists.
  Why don't we use the same approach as in clusters, events and scouts, and have the action in a button or menu?

That is correct. We do not have popup menus in lists. I added it as it just came naturally to me to have that option as an alternative to having the button at the top to avoid crowding. Unless we decide to go for popup for lists in the future, to maintain consistency, I agree button or menu should be the way.

Going forward I think it would be better to cluster features in the top `MenuBar` of the iEEG panel into a menu or exploring other parts of the panel to add them. The IEEG panel is getting crowded at the top. 

Even now after adding the merge button the `Contacts` menu is clipped from the top. For now, the `TB_DIM` scaling can be adjusted for proper fitting. 
Screenshot 2025-04-28 152455

toolbox/gui/panel_ieeg.m
chinmaychinara91: > Why is that option in here?
> Please add code where it belongs. If it does not align contacts, then it should no be here.

Yes `auto` just adds the locations to the channels. I was in a debate myself when I added this but went ahead with it instead because it very well handles all the needed checks to make sure the channels are updated properly. 
Apologies for not being more critical about this. I will make the necessary changes.

toolbox/gui/figure_3d.m
chinmaychinara91: Done in 93098b2.

toolbox/gui/panel_ieeg.m
rcassani: Did you consider that an electrode can exist without contacts?
It is like this when the electrode is created, it has no contacts

toolbox/gui/panel_ieeg.m
rcassani: If you were "in a debate myself", you could have contact me.

toolbox/gui/figure_mri.m
rcassani: > What I meant by "support for ECoG" in the comments was "adding" the feature for ECoG. Sorry for not being clear.

I understand that, my question is if that feature is going to be added for real, as it says `TODO`. 

> The way Brainstorm defines creation of ECoG electrodes, I don't think we need this feature.

Then why did you add it as a `TODO` feature? 

It seems to me that support for adding/removing ECOG contacts goes beyond of this PR.
Please do not expend time on it.

toolbox/gui/panel_ieeg.m
```diff
@@ -1391,32 +1464,115 @@ function AddElectrode()
...
+%% ===== ADD CONTACT =====
+function AddContact()
+    global GlobalData;
+    % Get electrodes
+    [sSelElec, iSelElec, iDSall, iFigall] = GetSelectedElectrodes();
+    if isempty(sSelElec)
+        java_dialog('warning', 'No electrode selected.', 'Add contact');
+        return
+    end
+    % If multiple electrodes were selected, update selection to just the electrode whose contacts are visible in iEEG panel
```
rcassani, line 1476: Can you make this comment more concise?

"If multiple electrodes were selected, update selection to just the electrode whose contacts are visible in iEEG panel"

"Use last selected electrode if multiple electrodes are selected"

toolbox/gui/panel_ieeg.m
```diff
@@ -1482,6 +1638,207 @@ function RemoveElectrode()
     UpdateFigures();
 end
 
+%% ===== REMOVE CONTACT =====
+function RemoveContact()
+    global GlobalData;
+    % Get selected electrode
+    [sSelElec, iSelElec, iDSall, iFigall] = GetSelectedElectrodes();
```
rcassani, line 1645: As with `AddContact`, just keep the last selected electrode.
This will:
- Update the GUI to show only one electrode selected
- Make clearer the code by changing `sSelElec(end)` by `sSelElec` 
- Avoid updating all electrodes even when only one was modified:
  ```
  % Set the changed electrode properties
  SetElectrodes(iSelElec, sSelElec);
  ```

rcassani (changes requested):
Getting there.
Please make the requested changes.

toolbox/gui/panel_ieeg.m
chinmaychinara91: My bad. Should have checked with you.

toolbox/gui/figure_mri.m
chinmaychinara91: Yes adding support for ECoG goes beyond this PR and I added a `TODO` to just keep a check for that. Will remove it as you suggested below.

toolbox/gui/figure_3d.m
```diff
@@ -1299,11 +1294,9 @@ function FigureKeyPressedCallback(hFig, keyEvent)
                         % Reset selection
                         bst_figures('SetSelectedRows', []);
                     end
-                    % For iEEG: Remove contact(s) (TODO: support for ECoG)
+                    % For iEEG: Remove contacts (TODO: support for ECoG)
```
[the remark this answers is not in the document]
chinmaychinara91: Check f6588e4

toolbox/gui/figure_3d.m
```diff
@@ -1264,8 +1274,9 @@ function FigureKeyPressedCallback(hFig, keyEvent)
                             view_timefreq(TfInfo.FileName, 'SingleSensor', SelChan{1});
                         end
                     end
-                % DELETE: SET CHANNELS AS BAD
+                % DELETE
```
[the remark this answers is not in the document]
chinmaychinara91: Check 64e911a

toolbox/gui/panel_ieeg.m
```diff
@@ -109,10 +110,11 @@
                 % Contacts list
                 jListCont = java_create('org.brainstorm.list.BstClusterList');
                 jListCont.setBackground(Color(.9,.9,.9));
-                jListCont.setLayoutOrientation(jListCont.HORIZONTAL_WRAP);
+                jListCont.setLayoutOrientation(jListCont.VERTICAL);
```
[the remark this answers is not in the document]
chinmaychinara91: Yes I did. Thank you for this. Makes sense to not have warping here.

toolbox/gui/panel_ieeg.m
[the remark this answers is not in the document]
chinmaychinara91: Check 79cf3e9

toolbox/gui/panel_ieeg.m
chinmaychinara91: Check fd5e3e6

toolbox/gui/panel_ieeg.m
rcassani: This behaviour is not consistent. The menu entry says "Remove SEEG contacts".
If the user manually selects all the contacts and in an electrode remove them instead of deleting the electrode, maybe they aimed to remove the contacts, and not the electrode.

toolbox/gui/panel_ieeg.m
```diff
@@ -1482,6 +1614,216 @@ function RemoveElectrode()
...
+    end
+    sSelCont = GetSelectedContacts();
+    % If every contact of that electrode is selected, remove the whole electrode
+    if numel(sSelCont) == sSelElec.ContactNumber
+        RemoveElectrode();
+    else
+        % Otherwise just remove the highlighted contacts
+        RemoveContact();
+    end
+end
```
rcassani, line 1760: (From the discussion in: https://github.com/brainstorm-tools/brainstorm3/pull/785#discussion_r2064469142)

This behaviour is not consistent. The menu entry says "Remove SEEG contacts".
If the user manually selects all the contacts and in an electrode remove them instead of deleting the electrode, maybe they aimed to remove the contacts, and not the electrode.

- Remove the `RemoveContactHelper` and the calls to it.
- Add the following checks in `RemoveContact` if not there yet
   - No electrode selected
   - Multiple selected electrodes
   - Electrode type (must be SEEG)
- Allow to remove all the contacts from an electrode without removing the electrode. As it if were created new

rcassani (changes requested):

toolbox/gui/panel_ieeg.m
chinmaychinara91: Check d8dbfd6.

toolbox/gui/panel_ieeg.m
chinmaychinara91: Check d8dbfd6.

PR #787, opened 2025-04-29
Load brainentropy instead of install

Edouard2laire:
Currently, every time we try to do a localization, a pop-up asks if we want to update brainentropy. This removes the unnecessary pop-up.


rcassani:
Does it ask because the local version is outdated? 
Or because the current trouble with the server?
https://neuroimage.usc.edu/forums/t/brainstorm-script-keeps-reinstalling-cat12/51252/8

PR #790, opened 2025-05-06
`RemoveElectrode`: Loop on electrodes only if it is an implantation

chinmaychinara91:
Restructure the `RemoveElectrode` function to loop on electrodes to delete from channel file only when we do an implantation.

### DIFF
--- a/toolbox/gui/panel_ieeg.m
+++ b/toolbox/gui/panel_ieeg.m
@@ -1552,10 +1552,10 @@ function RemoveElectrode(sSelElec)
     end
     % Loop on datasets
     for iDS = unique(iDSall)
-        % Loop on electrodes to delete
-        for iElec = 1:length(sSelElec)
-            % If new implantation scheme: delete all the contacts for this electrode
-            if isImplantation
+        % If new implantation scheme: delete all the contacts for this electrode
+        if isImplantation
+            % Loop on electrodes to delete
+            for iElec = 1:length(sSelElec)
                 % Get contacts for this electrode
                 iChan = find(strcmpi({GlobalData.DataSet(iDS).Channel.Group}, sSelElec(iElec).Name));
                 if isempty(iChan)


rcassani (changes requested):
This is not correct:

1. The check for implantation should be outside of the `Loop on datasets`. If id does notes not:
  a. If the folder name does not contain `Implantation`, the electrodes will be deleted, but not in the figure

2. The check should not only be outside of the `Loop on datasets`, but before. 
  If the folder name does not contain `Implantation`. Why do we want the user to confirm a deletion that will not happen?

toolbox/gui/panel_ieeg.m
```diff
@@ -1552,10 +1552,10 @@ function RemoveElectrode(sSelElec)
     end
     % Loop on datasets
     for iDS = unique(iDSall)
-        % Loop on electrodes to delete
-        for iElec = 1:length(sSelElec)
-            % If new implantation scheme: delete all the contacts for this electrode
-            if isImplantation
+        % If new implantation scheme: delete all the contacts for this electrode
```
rcassani, line 1555: Inaccurate comment: 
- It does not need to be a `new implantation scheme` 
  BTW, what is a `scheme`? Should not be a `Study` in Brainstorm jargon?
- `Delete all the contacts for this electrode` 
  What is `this electrode`, it deletes the contacts for all the selected electrodes
    
Unnecessary comment:
It just explains the IF structure

chinmaychinara91: > It does not need to be a new implantation scheme
> BTW, what is a scheme? Should not be a Study in Brainstorm jargon?
Yes it should be study

Agreed the comments should be updated

chinmaychinara91:
> 1. The check for implantation should be outside of the `Loop on datasets`. If id does notes not:
>    a. If the folder name does not contain `Implantation`, the electrodes will be deleted, but not in the figure
> 2. The check should not only be outside of the `Loop on datasets`, but before.
>    If the folder name does not contain `Implantation`. Why do we want the user to confirm a deletion that will not happen?

Agreed.

chinmaychinara91:
I am closing this PR.

I will have a new PR that better address all of the above along with (as we discussed) having `Add/Remove electrodes and contacts` defined only for **Implantation** as they are all related.

Issue #793, opened 2025-05-12
EEG points from Polhemus file not converted from CTF coordinates to SCS


Moo-Marc:
Bug found by Ellie in our lab.

First, when loading a CTF dataset with .pos file that includes EEG coordinates, they are not saved in Bst.  I changed this part of the code recently, but this was already the case as it only keeps the .Headpoints, while the EEG positions are saved in .Channels:
https://github.com/brainstorm-tools/brainstorm3/blame/b85498b157a4db15d01e66966ad14872b40eb43f/toolbox/io/in_fopen_ctf.m#L190

Second, when adding EEG positions, it looks like import_channel no longer applies the CTF to SCS transformation, though it should do it here, unless the check is incorrect in this case:
https://github.com/brainstorm-tools/brainstorm3/blob/b85498b157a4db15d01e66966ad14872b40eb43f/toolbox/io/import_channel.m#L549

A bit more testing and debugging needed.

rcassani:
Hi @Moo-Marc, can you share an simple CTF dataset with MEG and EEG to test and address those two points?

rcassani:
PR #797 aims to solve it. 
@Moo-Marc, can you take a look?

PR #797, opened 2025-05-21
Bugfix: Correct conversion of EEG points from Polhemus file on reviewing CTF data

rcassani:
Solve the two issues reported in #793

1. First, when loading a CTF dataset with .pos file that includes EEG coordinates, they are not saved in Bst.

On importing CTF data, it checks for position of EEG channels in the `.pos` file if any, and add these locations in `'Native'` coordinates, they are later converted in SCS. Solved in commit: 026342f1858fb02435319265048f811b1e33d6ba

2. When adding EEG positions from the `.pos` file, `import_channel` does not applies the (Native) CTF to SCS transformation

This was a bug, that came from the transformation from `RawPoints=>Native`, as an temporal `SCS` field was kept in `ChannelMat`. Solved in commit: a61d9ea97dc37ce2c913fe285dbbba9bb3045742

3. Fixes a bug in temporarily renaming fiducials. This bug did not have any consequence.
Solved in commit: 16b9db4eda71a453ee2ae8cc9179a58fb99dd607

## Results
On provided data:

1. On importing CTF data MEG+EEG with a `.pos` file with EEG points

| Before    | After |
| -------- | ------- |
|   |   |

2. On adding EEG positions from a `.pos` file with EEG points

image

### DIFF
--- a/toolbox/io/in_fopen_ctf.m
+++ b/toolbox/io/in_fopen_ctf.m
@@ -184,10 +184,21 @@ end
 
 %% ===== READ POLHEMUS FILE =====
 if ~isempty(pos_file)
-    % Read file
+    % Read file, returns in CTF native
     HeadMat = in_channel_pos(pos_file);
     % Copy head points
     ChannelMat.HeadPoints = HeadMat.HeadPoints;
+    % Copy channel positions for EEG
+    iHeadEegChannels = channel_find(HeadMat.Channel, 'EEG');
+    for ix = 1 : length(iHeadEegChannels)
+        HeadEegChannelName = HeadMat.Channel(iHeadEegChannels(ix)).Name;
+        iChannelEegChannel = channel_find(ChannelMat.Channel, HeadEegChannelName);
+        if ~isempty(iChannelEegChannel) && length(iChannelEegChannel) == 1 && ...
+           strcmpi(ChannelMat.Channel(iChannelEegChannel).Type, 'EEG') && isempty(ChannelMat.Channel(iChannelEegChannel).Loc)
+            ChannelMat.Channel(iChannelEegChannel).Loc = HeadMat.Channel(iHeadEegChannels(ix)).Loc;
+        end
+    end
+
     isAlign = true;
     % Warn if unsure about coordinate system. Should now be converted to "Native" CTF coil-based
     % when HPI points are present when loading the pos file.


toolbox/io/in_channel_pos.m
```diff
@@ -153,6 +153,8 @@
                 error('Unexpected transformation(s).')
             end
             ChannelMat.TransfEegLabels{iTrans} = 'RawPoints=>Native';
+            % Remove misleading orientation fields
+            ChannelMat = rmfield(ChannelMat, 'SCS');
         elseif numel(iCardinal) >= 3
```
Moo-Marc, line 158: I saw that in old data (~2012-2013), we had a few cases where there were no HPI, but the points named like "anat fids" were most likely the HPI coils.  I presume this early on when digitize_panel was created.  So I think we should change this fringe case to at least have a warning, or better, to ask the user to confirm if these are coils or anat, and not assume anything.

Moo-Marc (approved):
Good catch for misleading .SCS when still in "native" coordinates. Looks good to me.
But I added a comment about another related issue.  Thanks!

toolbox/io/in_channel_pos.m
rcassani: ~~As it is now, those points will not be detected as `HPI-` locations, thus it will lead to the error and display of the warning `'BST> Warning: Unable to ensure head points are in "native" coordinates.'`.~~

EDIT: Since those `anat fids` are not identified as `HPI-` locations the `RawPoints=>Native` will not carried out. Also, since those `anat fid` points will not have transformation 'Native=>Brainstorm/CTF' will not be performed either, so the the warning shown is: 
"Warning: Unable to confirm coordinate system of head points. Assuming "Native" CTF head-coil-based system."

---

~~We can restrict coping the locations for headpoints and channels IFF the `'RawPoints=>Native'` transformation is present:~~

~~What about that solution?~~

EDIT: Crossed text is not correct. Locations need to be copied regardless their coordinate system (Native or Branstorm/CTF). If a second alignment is needed or not (Brainstorm/CTF case) this is decided in `in_fopen_ctf.m` after the EEG locations are copied

```
    elseif ismember('Native=>Brainstorm/CTF', HeadMat.TransfMegLabels)
        disp('BST> Warning: head point coordinates appear to already be in SCS, presumably because of missing digitized head coils.');
        isAlign = false;
```

PR #800, opened 2025-05-22
Add/Remove electrodes and contacts IFF `Implantation`

chinmaychinara91:
Continuing from the discussion in https://github.com/brainstorm-tools/brainstorm3/pull/790, this PR addresses allowing Add/Remove electrode and contacts IFF it is a Brainstorm Implantation.

### DIFF
--- a/toolbox/gui/panel_ieeg.m
+++ b/toolbox/gui/panel_ieeg.m
@@ -1385,9 +1385,7 @@ function AddElectrode(newLabel)
         return;
     end
     % Proceed only if this is an implantation folder
-    ChannelFile = GlobalData.DataSet(iDS(1)).ChannelFile;
-    [~, folderName] = bst_fileparts(bst_fileparts(ChannelFile));
-    if ~strcmpi(folderName, 'implantation')
+    if ~isImplantationFolder(iDS)
         return;
     end
     % Parse inputs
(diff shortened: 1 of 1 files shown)


chinmaychinara91:
The way it is now, the Add/Remove electrodes and contacts GUI elements are still active even when opening a non-implantation channel. Two proposals:
1. Disable them for non-implantation channels completely.
2. Keep them active, but when users click on them, give a pop up message saying this feature is only available while doing Brainstorm Implantation.

@rcassani @tmedani what do you think about these proposals or if you could suggest something better ?

toolbox/gui/panel_ieeg.m
```diff
@@ -3285,9 +3301,9 @@ function SetElectrodeLoc(iLoc, jButton)
     jButton.setForeground(java.awt.Color(0, 0.8, 0));
     % Get the contact for this electrode
     iChan = find(strcmpi({GlobalData.DataSet(iDS(1)).Channel.Group}, sSelElec.Name));
-    % Check if this is an new implantation folder
-    [fPath, folderName] = bst_fileparts(bst_fileparts(GlobalData.DataSet(iDS(1)).ChannelFile));
-    isImplantation = ~isempty(strfind(folderName, 'Implantation'));
+    % Check if this is an implantation folder
+    [~, folderName] = bst_fileparts(bst_fileparts(GlobalData.DataSet(iDS(1)).ChannelFile));
+    isImplantation = strcmpi(folderName, 'implantation');
```
rcassani: Why not to make a function of this `isImplantationFolder()`

This is text book example of when to use functions, these lines are copied-pasted at least 6 times.

rcassani: Do the check for folder name starting with `'Implantation'`.

In this sense, the user could duplicate the folder and do experiments in the copies

rcassani (changes requested):
Copy-pasting code is not encourage, use a function when needed

toolbox/gui/panel_ieeg.m
chinmaychinara91: Check 2cd1bfc

chinmaychinara91: Check 2cd1bfc

Issue #803, opened 2025-06-15
Progress bar is relatively slow


Edouard2laire:
Hello, 

I was recently trying to update MEM code to improve performance and realize that a significant amount of time was spent updating the progress bar. This is especially true for large data, for which ```bst_progress('inc', 1); ``` would be call a large amount of time.  

Here is a very small program showing the issue: 

```matlab

profile on -historysize 10000000000
main(1e4);
profile viewer

function main(n)

    bst_progress('start', 'Test progress bar... ' , 'Test progress bar... ', 1, n);

    for i = 1:n

        M  = randn(3,3);
        
        bst_progress('inc', 1); 

    end
    bst_progress('stop');

end

```

The program run in 51 seconds, with 50 seconds spent in bst_progress. In the MEM case, it was not that extrem as the lopp was doing something but still bst_progress represented ~25% of the runtime. 

In theory, this can be solved by limiting the amount of time bst_progress is called : 

```matlab

function main2(n)

    bst_progress('start', 'Test progress bar... ' , 'Test progress bar... ', 1, 100);
    freq_update_progress_bar = round(n / 100);

    for i = 1:n

        M  = randn(3,3);
        if mod(i, freq_update_progress_bar) == 0
            bst_progress('inc', 1); 
        end
    end
    bst_progress('stop');
end


```

In this case, main2 run in 0.644s. 
However, this trick does not work for par-for loops: 

```matlab

function main3(n)

    bst_progress('start', 'Test progress bar... ' , 'Test progress bar... ', 1, n);
    
    q = parallel.pool.DataQueue;
    afterEach(q, @(x) bst_progress('inc', 1));

    for i = 1:n

        M  = randn(3,3);
        send(q, 1); 

    end
    bst_progress('stop');
end

```
Main3 ran in 74 seconds : 

<img width="1415" alt="Image" src="https://github.com/user-attachments/assets/890e7d09-968e-4356-bf0c-1eafe5b0324a" />

Few observation: 

<img width="1415" alt="Image" src="https://github.com/user-attachments/assets/b033f24c-f63a-4789-b70a-7f1abb906874" />

- it seems that getting information from java is taking some times. Maybe we can save a copy in matlab so we don't have to call getValue and getMaximum from the jProgressBar
- setValue seems also costly so maybe we can update the bar only very time a new % is reach. 

Edouard

PR #804, opened 2025-06-15
improve progress bar performance

Edouard2laire:
This PR implements some ideas from #803 to improve the performance of the progress bar.

After this update, the benchmark proposed in #803 : 

+ Main runs in 2s instead of 50s: 

image

+ Main3 runs in 5s instead of 70s: 
image


rcassani:
@Edouard2laire, can you test this on your Mac?
The performance is also improved on Linux though it is not as drastic, (`main1` from 2s to 1s)

Edouard2laire:
tested, it seems to be working. the time is now around .5 seconds.

PR #806, opened 2025-06-18
Corr with flat signals

rcassani:
This PR updates the code in `bst_corrn` to allow to return `NaN` if correlation is computed with flat signals. This is inline with the Matlab function `corr`

It also updates `bst_connectiity` to be able to save `NaN` in the connectivity matrix.

This addresses the issue reported in #774

### DIFF
--- a/toolbox/connectivity/bst_connectivity.m
+++ b/toolbox/connectivity/bst_connectivity.m
@@ -1078,9 +1078,8 @@ for iFile = 1 : length(FilesA)
             bst_report('Error', OPTIONS.ProcessName, [], ['Invalid method "' OPTIONS.Method '".']);
             CleanExit; return;
     end
-    % Replace any NaN values with zeros
+    % Replace Inf values with zeros
     if isnumeric(R)
-        R(isnan(R)) = 0;
         R(isinf(R)) = 0;
     end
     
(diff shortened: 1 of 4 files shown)


rcassani:
Hi @Edouard2laire, while it was functional, the proposed solution in #774 led to an inconsistent behaviour, i.e., the user is asking for connectivity between N signals but only a subset of those is returned. It was not indicated why they were removed, or which ones were removed. The fact that connectivity matrices can change of size depending on the data to be analyzed, was confusing.

So going deeper on rationale, the core issue was that `bst_corrn` returned `1` at the correlation between two flat signals, and it returned `0` for the correlation of a signal and a flat signal. So this was updated, now it returns `NaN` in those cases and that is shown in the different connectivity representations:

image
`s2` and `s4` were flat signals and correlation between all pairs was computed.

What do you think of this solution?

Edouard2laire:
yup. i love it. This is was i was actually trying to achieve without success :)  

Thanks for implementing it.

Edouard2laire:
Is this solution working for all Connecticut measure or only correlation ?

rcassani:
> Is this solution working for all connectivity measure or only correlation ?

It depends on the connectivity metric.

This update allows to save `NaN`s in the connectivity matrix. So if a metric returns `NaN`s there values are respected, and will they will appear as grayed out in the image figure and their respective links will not be shown in the graph figure.

PR #807, opened 2025-06-19
Load map without std field

Edouard2laire:
When testing the correlation fix in #806, i encounted this error: 

image

This PR is here to fix it :)


rcassani:
Do you know how this result file was created? 
`Std` should be there, it is part of the template of fields to save (`db_template('resultsmat')`)

PR #813, opened 2025-07-03
Import events BIDS: ask if the offset should be added

Edouard2laire:
In some case, we don't want to apply the offset; notably if the file is cut (extract time) before importing the event file.

### DIFF
--- a/toolbox/io/in_events_bids.m
+++ b/toolbox/io/in_events_bids.m
@@ -89,9 +89,32 @@ for iEvt = 1:length(uniqueEvt)
     events(iEvt).label  = uniqueEvt{iEvt};
     events(iEvt).epochs = ones(1, length(iMrk));
     events(iEvt).times  = [onsets{:}];
-    % Convert latencies from the first sample in the file to the real timing 
-    % See specs: https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/05-task-events.html
-    events(iEvt).times = events(iEvt).times + sFile.prop.times(1);
+
+
+    % ===== TIME OFFSET =====
+    isAddOffset = 1;
+    if isInteractive
+        % Convert latencies from the first sample in the file to the real timing 
+        % See specs: https://bids-specification.readthedocs.io/en/stable/04-modality-specific-files/05-task-events.html
+
+        if (sFile.prop.times(1) ~= 0)
+            res = java_dialog('question', ['The raw data file starts at ' num2str(sFile.prop.times(1)) ' sec.' 10 10 ...
+                                          'Is this offset already added to these events?' 10 10],...
+                                          'Import events', [], {'Yes', 'Add Offset','Cancel'},'Yes');
+            if isempty(res) || strcmpi(res, 'Cancel')
+                bst_progress('stop');
+                return;
+            elseif strcmpi(res, 'Yes')
+                isAddOffset = 0;
+            end
+        end
+    end
+
+    % Add a column for time
+    if isAddOffset
+        events(iEvt).times = events(iEvt).times + sFile.prop.times(1);
+    end
+  
     % Extended events if durations are defined for all the markers
     if all(~cellfun(@isempty, durations)) && all(~cellfun(@(c)isequal(c,0), durations))
         events(iEvt).times(2,:) = events(iEvt).times + [durations{:}];


rcassani:
This can get confusing for the cases where the data does not originally start in 0 s.

Just an example at hand is `sub-01_ses-meg_task-facerecognition_run-01_meg` in [DS000117](https://openneuro.org/datasets/ds000117/versions/1.1.0). 

As per the BIDS description on the [`Onset`](https://bids-specification.readthedocs.io/en/stable/modality-specific-files/task-events.html#stimuli-directory)
> Onset (in seconds) of the event, measured from the beginning of the acquisition of the first data point stored in the corresponding task data file.

In Brainstorm, the events occurrences are saved as "absolute time", not relative to the beginning of the recording. That is why the events from `_events.tsv` are imported with respect to the time of the first sample. 

How are you cutting the recording files?

## Alternatives
1. Import the original data from the BIDS dataset; import the `_events.tsv` file; then cut the raw file

2. Cut the original data at `X s`; import the `_events.tsv` file; add the offset of `X s` with the process (or GUI entry) to add time offset to events.

Edouard2laire:
> it is explicit in the BIDS specification that they are relative to the first time sample in the recording file. So, these event onsets are not wrong, they apply to the original recording file. What is wrong is changing the first time sample of the recording file without changing the event onsets.


Then out_events_bids is wrong as it's exporting the onset relative to t = 0, and not relative to the beginning of the file. 


> The correct way would be to make a copy of that `_events.tsv` file and apply the same modifications that are done when the sync and extract time are performed in the recording file. A new recording file with a new event file. Moreover, if the new sync+cut data is to be shared as BIDS, this new recording file and its new event file will follow the BIDS specification about event onset being relative to the first time sample.


I agree with the argument. I think the main point of Shahla was, let's say you have a File A, and an event file (A.tsv) created by a neurologist: you import borth and everything works well, you can then apply some pre-processing including some cutting...   But now lets say, the neurologist come back to the original recording A and find some more event of interest and send you and updated version of the first tsv.  You can only import the event in the raw data before the preprocessing has been done and everything following needs to be recomputed. 

This PR is mainly here to be able to re-import the tsv at any point and avoid un-nescessary computation. The tsv is still valid in the context of BIDS because it is linked to file A, that was not cut. 


> The change proposed in this PR, results in an ambiguous behaviour, as in the shared example (`sub-01_ses-meg_task-facerecognition_run-01_meg` in [DS000117](https://openneuro.org/datasets/ds000117/versions/1.1.0)). In that file, the recordings start at `226 s`, and the first event in the `_event.tsv` has as onset at `24.207 s`, which means it happens at time `250.207 s`. There is no reason to ask the user if they want or not add the time offset, the onsets are relative to the first time sample, `250.207` is the correct event time.

I don't think this is very ambiguous. It is just making the process a bit more generic. The same question is actually asked for in_events_array. Note that in this case, adding the offset is the default behavior as expected by BIDS standards, whereas its not in the case of in_events_array. 




Regards, 
Edouard

rcassani:
> Then out_events_bids is wrong as it's exporting the onset relative to t = 0, and not relative to the beginning of the file. 

This is true, and we will be correct ASAP.
Exporting events from Brainstorm as `_events.tsv` and reimporting them in Brainstorm must give the same events times.
FYI @yashvakilna 

> This PR is mainly here to be able to re-import the tsv at any point and avoid un-nescessary computation. The tsv is still valid in the context of BIDS because it is linked to file A, that was not cut.

I agree with you, the event file is valid one, and it is valid for file A. But it is not for the modified file A, which is the intended file for which the events will be imported. 

> I don't think this is very ambiguous. It is just making the process a bit more generic.
With the example data that I provided, let say the original `.fif` is imported, then its respective `_events.tsv`. The prompt will be there, when it is not needed.

Also, the proposed change would apply IFF the original BIDS started at 0s. 
E.g., An original recording starting at 50s, with a first event onset is at 200s, so the event occurrence is 250s. The first 50s of the original file are cut, so it now starts at 100s, (first event onset is still 200s), by NO applying the offset it the event occurrence will be 200s, which is wrong. Applying the offset, occurrence is 300s also wrong as that event file is not for the cut version. So way to do would be to import them with the offset, and the apply the same cut to the event onset (reduce the event occurrence by 50s), now it is ok at 250s.
 
The question of offset is asked in `in_events_array` because there is not specification that say if the times (or samples) are relative to the t=0 or to the first sample time. This is not the case with BIDS events files, it is explicit in their specification that they are relative to the first sample time.

Edouard2laire:
Thank you 😊

PR #814, opened 2025-07-04
Improve plugin unload when .brainstorm is a symbolic link

Edouard2laire:
On my lab computer /home/edelaire/.brainstorm is a symbolic link to /NAS/home/edelaire/.brainstorm and because of that, I could not unload plugin.  This PR fixes that.

### DIFF
--- a/toolbox/core/bst_plugin.m
+++ b/toolbox/core/bst_plugin.m
@@ -2383,15 +2383,19 @@ function [isOk, errMsg, PlugDesc] = Unload(PlugDesc, isVerbose)
     % === UNLOAD PLUGIN ===
     % Do not modify path in compiled mode
     if ~bst_iscompiled()
-        matlabPath = str_split(path, pathsep);
         % Remove plugin folder and subfolders from path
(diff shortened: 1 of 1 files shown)


rcassani:
There should not be a problem installing, loading, unloading and unistalling plugins.
When `addpath` and `rmpath` are used, symbolic links are resolved, and the absolute paths are used.

Did you happen to move `.brainstorm` from `/home/edelaire/.brainstorm` and then create the symbolic link? All of this when the plugin path was in the Matlab path?

If not, can you point me how to replicate the issue that this PR aims to solve?

Edouard2laire:
The issue is not addpath / rmpath, it is the condition ' ismember(allSubFolders{i}, matlabPath)' 

For example, when unloading mcxlab, then
allSubFolders{2} is '/home/edelaire/.brainstorm/plugins/mcxlab-cuda/mcxlab' but
matlabPath{1} is '/NAS/home/edelaire/.brainstorm/plugins/mcxlab-cuda/mcxlab'

They are the same folder, and matlab recognize it when calling rmpath but its not called since ismember(allSubFolders{i}, matlabPath) returns 0

PR #816, opened 2025-07-14
Integrate GARDEL's automatic contact localization in Brainstorm

chinmaychinara91:
This PR aims to automatically detect SEEG contacts in Brainstorm using GARDEL tool's automatic detection pipeline.

About GARDEL: [link](https://gitlab-dynamap.timone.univ-amu.fr/public_shared_tools/gardel/-/wikis/home)

Test data: [link](https://drive.google.com/drive/folders/1Kw9hbKVwh47EmBbiH1Z8DhrHHzeKYavt?usp=sharing)

Tutorial (WIP): ~https://neuroimage.usc.edu/brainstorm/Tutorials/AutoContactLocalization~ https://neuroimage.usc.edu/brainstorm/Tutorials/AutoContactLocGardel

***
**Steps to test:**
1. Right click subject > **Import MRI**. Set fiducials and save.
2. Right click subject > **Import CT**, coregister (SPM), reslices, and skull strip (SPM).
3. Right click on **CT_spm_reslice_masked_spm > CT segmentation > Generate threshold mesh from CT**
4. Right click subject > **SEEG/ECOG implantation**
5. Choose **CT+MRI+IsoSurf** (can choose any but this option gives a better visualization as both 3D and 2D open up). 
6. In **Panel iEEG**, at the bottom click **Auto > Gardel**. ~magic~ Gardel's `elec_auto_segmentation` function is called, and the live automatic implantation can be seen.
7. Close figures and click **Save** in the prompts to save the channel data.

@tmedani @samuel.medinavillalon@gmail.com

EDIT: Add clarification on step 2

### DIFF
--- a/toolbox/gui/panel_ieeg.m
+++ b/toolbox/gui/panel_ieeg.m
@@ -51,8 +51,8 @@ function bstPanelNew = CreatePanel() %#ok<DEFNU>
         jToolbarE.setPreferredSize(TB_DIM);
         jToolbarE.setOpaque(0);
         % Add/remove
-        gui_component('ToolbarButton', jToolbarE,[],[], {IconLoader.ICON_PLUS, TB_DIM}, 'Add new electrode', @(h,ev)bst_call(@AddElectrode));
-        gui_component('ToolbarButton', jToolbarE,[],[], {IconLoader.ICON_MINUS, TB_DIM}, 'Remove selected electrodes', @(h,ev)bst_call(@RemoveElectrode));
+        gui_component('ToolbarButton', jToolbarE,[],[], {IconLoader.ICON_SEEG_DEPTH_PLUS, TB_DIM}, 'Add new electrode', @(h,ev)bst_call(@AddElectrode));
+        gui_component('ToolbarButton', jToolbarE,[],[], {IconLoader.ICON_SEEG_DEPTH_MINUS, TB_DIM}, 'Remove selected electrodes', @(h,ev)bst_call(@RemoveElectrode));
         gui_component('ToolbarButton', jToolbarE,[],[], {IconLoader.ICON_FUSION, TB_DIM}, 'Merge selected electrodes', @(h,ev)bst_call(@MergeElectrodes));
         % Set color
         jToolbarE.addSeparator();
@@ -79,8 +79,8 @@ function bstPanelNew = CreatePanel() %#ok<DEFNU>
         % Button "Select surface centroid"
         jButtonCentroid = gui_component('ToolbarToggle', jToolbarC, [], '', IconLoader.ICON_GOOD, 'Select surface centroid', @(h,ev)panel_coordinates('SetCentroidSelection', ev.getSource.isSelected()));
         % Buttons "Add" and "Remove" contacts
-        gui_component('ToolbarButton', jToolbarC,[],[], {IconLoader.ICON_EVT_OCCUR_ADD, TB_DIM}, 'Add SEEG contact', @(h,ev)bst_call(@AddContact));
-        gui_component('ToolbarButton', jToolbarC,[],[], {IconLoader.ICON_EVT_OCCUR_DEL, TB_DIM}, 'Remove SEEG contacts', @(h,ev)bst_call(@RemoveContact));
+        gui_component('ToolbarButton', jToolbarC,[],[], {IconLoader.ICON_SEEG_SPHERE_PLUS, TB_DIM}, 'Add SEEG contact', @(h,ev)bst_call(@AddContact));
+        gui_component('ToolbarButton', jToolbarC,[],[], {IconLoader.ICON_SEEG_SPHERE_MINUS, TB_DIM}, 'Remove SEEG contacts', @(h,ev)bst_call(@RemoveContact));
 
     % ===== PANEL MAIN =====
     jPanelMain = gui_component('Panel');


toolbox/core/bst_plugin.m
```diff
@@ -175,6 +175,20 @@
...
+    PlugDesc(end).Category       = 'Anatomy';
+    PlugDesc(end).URLzip         = 'https://gitlab-dynamap.timone.univ-amu.fr/public_shared_tools/gardel/-/archive/GARDEL_Brainstorm/gardel-GARDEL_Brainstorm.zip';
+    PlugDesc(end).URLinfo        = 'https://gitlab-dynamap.timone.univ-amu.fr/public_shared_tools/gardel/-/wikis/home';
+    PlugDesc(end).TestFile       = 'code/GARDEL.m';
+    PlugDesc(end).ReadmeFile     = 'README.md';
+    PlugDesc(end).CompiledStatus = 2;
+    PlugDesc(end).RequiredPlugs  = {'spm12'};
+    PlugDesc(end).LoadFolders    = {'*'};
+    PlugDesc(end).InstalledFcn   = {};
+    PlugDesc(end).DeleteFiles    = {'.gitignore'};
```
rcassani: Among the plugin files, there is a NIfTI file `T1.nii` that accounts for around half of the GARDEL plugin. It seems this file is not required for the automatic segmentation, can you confirm?

toolbox/core/bst_plugin.m
```diff
@@ -175,6 +175,20 @@
     PlugDesc(end).LoadFolders    = {'ct2mrireg'};
     PlugDesc(end).DeleteFiles    = {'fmri_analysis', 'for_clio', 'mixed_atlas', 'process_script', 'reg_prepost', 'visualize_channels', '.gitignore', 'README.md'};
     
+    % === ANATOMY: GARDEL ===
+    PlugDesc(end+1)              = GetStruct('gardel');
+    PlugDesc(end).Version        = 'latest';
+    PlugDesc(end).Category       = 'Anatomy';
```
rcassani: Would not be better to be in the `SEEG` category?
Yes, GARDEL operates in anatomy files. But its results is used in the functional side, and it is not for any anatomy, but for anatomy of subjects with SEEG electrodes.

toolbox/gui/panel_ieeg.m
```diff
@@ -3751,4 +3880,30 @@ function ExportChannelFile(isAtlas)
...
+        % Base case: a single letter.
+        labelText = char('A' + labelNum - 1);
+    else
+        % Recursively generate multi-letter labels
+        labelNum  = labelNum - 1;
+        remainder = mod(labelNum, nChars);
+        quotient  = floor(labelNum / nChars);
+        labelText = [NumToText(quotient) char('A' + remainder)];
+    end
+end
```
rcassani, line 3911: Pretty nice use of recursion and char indexing, thus it is fast 😍

rcassani (commented):
This PR integrates GARDEL's `elec_auto_segmentation` function to be called from the iEEG panel for automatic detection of SEEG electrodes and their contacts.

Some questions:

1. One of discussion topics with Samuel was the option of using the original CT volume as input to GARDEL, so the electrode
localization would have higher resolution. However in the way this PR implements the GARDEL integration, that does not
seem possible because skull striping is required, and this (either SPM or BrainSuite) require both MRI and CT volumes to
have the same size. 
  
    Are we going to still offer the option of electrode detection with the original CT volume?

2. Can we avoid the manually generating the threshold mesh from CT?
   As it is done in GARDEL: https://gitlab-dynamap.timone.univ-amu.fr/public_shared_tools/gardel/-/commit/9a4f455944a7d3d20650bbe8ec0406171cd44131#2ad2b46c6a33e85f2727201fa4c21dfa13400328

3. Can we ask GARDEL's developers to add a check for the Parallel Computing Toolbox? This to make the tool available a users without this Matlab.
    The check can be done with:
    ```
    % 'distrib_computing_toolbox' former name of the 'Parallel Computing Toolbox'
    license('test', 'distrib_computing_toolbox')
    ```
    
4. The Auto > GARDEL option works in source code and compiled Brainstorm. 
    However, SPM coregistration does not, that would be important to fix (not in this PR though). Otherwise a requirement is that the MRI and CT volumes are already coregistered.

5. GUI in iEEG panel can be improved. I'll work on it.

PS, @chinmaychinara91, I already made some changes on the PR. Let me know if you have any question about them.

chinmaychinara91:
Thank you for the review @rcassani.

> 1. One of discussion topics with Samuel was the option of using the original CT volume as input to GARDEL, so the electrode
>    localization would have higher resolution. However in the way this PR implements the GARDEL integration, that does not
>    seem possible because skull striping is required, and this (either SPM or BrainSuite) require both MRI and CT volumes to
>    have the same size.
>    Are we going to still offer the option of electrode detection with the original CT volume?

As we had discussed earlier as a group, there are two ways to have GARDEL in Brainstorm:
1. Use the GARDEL's elec_auto_segmentation function to be called from the iEEG panel 
2. Open GARDEL from Brainstorm, do the work there and bring back things to Brainstorm.

This PR covers the 1st point. I discussed with @tmedani and having the 2nd one is not a priority right now. Once we have the 1st point covered through this PR and roll it out with proper documentation then will aim for the 2nd point.

Now, the main reason why we needed to have the original CT was not because of its higher resolution but for the 2nd point above. The original CT is required there because GARDEL works in raw CT space. 

But as you rightly said, working in the CT space gives higher resolution so definitely the detection is more accurate. I have tested this and to do that user needs to: import the CT first, set the fiducials in the CT, import MRI, coregister MRI to CT and then run skull stripping and generating isosurf from that. What we can do is explain both the ways with pros and cons in the tutorial page. Let me know if you have some other suggestions.

> 2. Can we avoid the manually generating the threshold mesh from CT?
>    As it is done in GARDEL: https://gitlab-dynamap.timone.univ-amu.fr/public_shared_tools/gardel/-/commit/9a4f455944a7d3d20650bbe8ec0406171cd44131#2ad2b46c6a33e85f2727201fa4c21dfa13400328

Do you mean to do have something like GARDEL to automatically get a good threshold initial estimate ?
`CT_thr =quantile(single(CT(:)),0.75) + 2.5*iqr(single(CT(:)));`
 
We can avoid it yes by just automatically passing the best guessed isoValue either by above or like we do currently (mean of min and max level CT intensity) but it is not always the case that isoValue will work for the automatic segmentation. Even in GARDEL a GUI pops up to adjust the initial estimated isoValue before actually doing the electrode segmentation. I have observed it in some cases where there could be some artefacts and it might need some additional tuning of the threshold. Therefore I feel it would be good to manually generate this thresholded mesh to do the fine tuning.

> 3. Can we ask GARDEL's developers to add a check for the Parallel Computing Toolbox? This to make the tool available a users without this Matlab.
>    The check can be done with:
>    ```
>    % 'distrib_computing_toolbox' former name of the 'Parallel Computing Toolbox'
>    license('test', 'distrib_computing_toolbox')
>    ```
Sure I can check with them. But I have tested it and it works even without the `Parallel Computing Toolbox`.

> 4. The Auto > GARDEL option works in source code and compiled Brainstorm.
>    ~However, SPM coregistration does not, that would be important to fix (not in this PR though). Otherwise a requirement is that the MRI and CT volumes are already coregistered.~ SOLVED in [28450f1](https://github.com/brainstorm-tools/brainstorm3/commit/28450f19d9a7d12a0e750d3f438c3b7fc681a9c0)

Thanks for this

> 5. GUI in iEEG panel can be improved. I'll work on it.

Ok. Thank you.

> 6. Documentation is needed.

I am working on it. Link to it added in the main description. Will let you know once it is ready.

> 
> PS, @chinmaychinara91, I already made some changes on the PR. Let me know if you have any question about them.

They look good. Thank you.

toolbox/core/bst_plugin.m
chinmaychinara91: Agreed

toolbox/core/bst_plugin.m
chinmaychinara91: That is correct.

rcassani: Please address it
Check other files and folders that are not needed

toolbox/core/bst_plugin.m
rcassani: Please make this change

chinmaychinara91: Check [a9c0d0b](https://github.com/brainstorm-tools/brainstorm3/pull/816/commits/a9c0d0b4213af8a26931f1740d33bb69a234eefd)

toolbox/core/bst_plugin.m
chinmaychinara91: It seems most of the other files were not required for the automatic segmentation 😅. Check [37a3d85](https://github.com/brainstorm-tools/brainstorm3/pull/816/commits/37a3d8508b0ab8b3eb0c7071156bc0b175bdfb57).

chinmaychinara91:
> GUI in iEEG panel can be improved. I'll work on it.

@rcassani are you planning to do this in this PR. Checking before I start taking screenshots for the tutorial.

rcassani:
@chinmaychinara91, yes UI update will be done in this PR

PR #819, opened 2025-07-21
Thresholding of isosurface without creating a new one

chinmaychinara91:
This PR addresses point # 4 in PR https://github.com/brainstorm-tools/brainstorm3/pull/732).

>Thresholding of isosurface without creating a new one
Creating of a new file was supposed to be temporary (added in https://github.com/brainstorm-tools/brainstorm3/pull/689)
NEW FILE IS STILL CREATED. SEE WAYS TO UPDATE ONLY IN THE FIGURE AND SAVE ON EXIT


toolbox/anatomy/tess_isosurface.m
rcassani, line 1: Since we can now edit the isosurface in the surface panel.
 
- Would not be better than `tess_isosurface` always creates a new file?

Moreover, as if `tess_isosurface` updates the isosurface file, it will update the last one in the case that there are more than one isosurfaces. This is quite arbitrary.

toolbox/gui/panel_surface.m
```diff
@@ -2784,4 +2822,56 @@ function ApplyDefaultDisplay() %#ok<DEFNU>
...
+                    s.SulciMap    = GlobalData.Surface(iSurf).SulciMap;
+                    newIsoValue   = GlobalData.Surface(iSurf).Comment;
+                    % Get parameters from original file (not the loaded)
+                    [ctFile, oldIsoValue, isoRange] = GetIsosurfaceParams(SurfaceFile);
+                    sIsoSrfOrg = load(file_fullpath(SurfaceFile), 'Comment', 'History');
+                    s.Comment   = strrep(sIsoSrfOrg.Comment, num2str(oldIsoValue), newIsoValue);
+                    s.History   = sIsoSrfOrg.History;
+                    % Update History
+                    s = bst_history('add', s, 'threshold_ct', ...
+                                sprintf('Thresholded CT: %s threshold = %s minVal = %d maxVal = %d', ctFile, newIsoValue, isoRange));
```
rcassani, line 2875: Here, it is appending. Rewriting the file history does not seem appropriated

toolbox/anatomy/tess_isosurface.m
rcassani: @chinmaychinara91, this questions is not outdated.
(not sure why it appears like that)

chinmaychinara91:
> @chinmaychinara91, I refactored to:
> 
> Fix the bug that modifications were not shown in all figures
> Store modified surface in GlobalData.Surface
> Save is triggered on unloading the surface
> Prepare SaveModifications for future uses

Thanks for these. I will test all out thoroughly and get back.

> Bonus: Is there a reason the load the isosurf with gray color? It may be better to load it with a color that helps with the contrast with the MRI slices

There was no such preference for gray color. Your proposal sounds like a good idea. Couple of things:
1. The color should not be yellow (because the crosshair color while doing surface selection is yellow)
2. The color should be something contrasting the electrodes rendered as well.
What do you think ?

toolbox/anatomy/tess_isosurface.m
chinmaychinara91: > Since we can now edit the isosurface in the surface panel.
> 
> Would not be better than tess_isosurface always creates a new file?

Did not understand the question here. 

> Moreover, as if tess_isosurface updates the isosurface file, it will update the last one in the case that there are more than one isosurfaces. This is quite arbitrary.

That is not the case. It updates the correct file but brings it to the last when `db_add_surface` is called. Right click on subject > click `Reload` after and you can see the surfaces being reordered correctly in the tree.

toolbox/gui/panel_surface.m
chinmaychinara91: Sounds good.

chinmaychinara91:
1. Tested with different combination of figures. Looks good.
2. One scenario to looks at: Open Isosurface (loads isosurafce with the isovalue in memory), use slider to change to new isovalue, bring slider back to initial isovalue, close figure, the Save prompt comes up which should not as the isovalue has not changed.

toolbox/anatomy/tess_isosurface.m
rcassani: With nothing loaded in memory. 

When the function `tess_isosurface()` is called from the CT segmentation, it either creates a new isosurface file if it there is none, or it updates the last isosuface file that exist for the Subject that holds the CT. 

This behaviour was there to avoid having multiple isosurface files. But now we can edit an isosurface by opening it and changing the iso threshold in the surface panel. So, the question:

Would not be better than `tess_isosurface` always creates a new file (when called from the CT segmentation popup menu)?

chinmaychinara91: Thank you for the clarification. Yes definitely it is better than `tess_isosurface` always creating a new file.

rcassani:
@chinmaychinara91, for the isosurface color, what about a gray-like with some hue [190,190,220] (RGB)

toolbox/anatomy/tess_isosurface.m
rcassani: Done in [1e007f4](https://github.com/brainstorm-tools/brainstorm3/pull/819/commits/1e007f49c19e717b15ed92737c05aad302e5dcf9)

rcassani:
> One scenario to looks at: Open Isosurface (loads isosurafce with the isovalue in memory), use slider to change to new isovalue, bring slider back to initial isovalue, close figure, the Save prompt comes up which should not as the isovalue has not changed.

Solved in [3a6ca19](https://github.com/brainstorm-tools/brainstorm3/pull/819/commits/3a6ca19f62df4be53df7bfb2473ce1e5821a8ceb)

PR #822, opened 2025-07-22
Smooth Surface files

tmedani:
Add option to smooth surface data (triangular mesh)

### DIFF
--- a/toolbox/anatomy/tess_smooth_select.m
+++ b/toolbox/anatomy/tess_smooth_select.m
@@ -43,17 +43,17 @@ if (nargin < 2) || isempty(Method)
 end
 
 MultipleFiles = [];
-% USAGE: tess_smooth_wrapper(TessMat, ...)
+% USAGE: tess_smooth_select(TessMat, ...)
 if isstruct(TessFile)
     TessMat  = TessFile;
     TessFile = [];
-% USAGE: tess_smooth_wrapper({TessFile1, TessFile2}, ...)
+% USAGE: tess_smooth_select({TessFile1, TessFile2}, ...)
 elseif iscell(TessFile)
     if (length(TessFile) > 1)
         MultipleFiles = TessFile;
     end
     TessFile = TessFile{1};
-% USAGE: tess_smooth_wrapper(TessFile, ...)
+% USAGE: tess_smooth_select(TessFile, ...)
 elseif ischar(TessFile)
     TessFile = file_short(TessFile);
     TessMat = [];
@@ -139,7 +139,7 @@ end
 %% ===== PROCESS MULTIPLE FILES =====
 if ~isempty(MultipleFiles)
     for i = 1:length(MultipleFiles)
-        NewTessFile = tess_smooth_wrapper(MultipleFiles{i}, Method, Params);
+        NewTessFile = tess_smooth_select(MultipleFiles{i}, Method, Params);
     end
     return
 end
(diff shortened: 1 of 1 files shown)


tmedani:
Adds a new context‑menu item for surface files in the Anatomy tab that lets users smooth a mesh. Users can set the smoothing method and factor (alpha) and number of iterations (iter)

rcassani:
@tmedani, do we need this smoothing approach in addition to what we have in [`tess_smooth.m`](https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/anatomy/tess_smooth.m)?

tmedani:
@rcassani  while we can use [tess_smooth.m](https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/anatomy/tess_smooth.m), I believe it would be valuable to add the `sms `method as an alternative since it has other smoothing algo (other than the basic laplacian).
Both we and other users may need it, also since sms is part of iso2mesh and we have it in many places already implemented.

To keep the interface consistent, we could display a message box asking the user whether to use the standard Brainstorm function or the sms methods — similar to the existing popup for "Less vertices". This would give users flexibility.

@rcassani what do you think?

rcassani:
@tmedani, I'll prepare this and let you know when ready

tmedani:
Great. 
I have it around. Let me push the updates, and you can review. [by end of the day]

toolbox/anatomy/tess_smooth_select.m
```diff
@@ -0,0 +1,220 @@
...
+        ['<HTML><B> iso2mesh / Laplacianhc:</B><BR>' ...
+        '&nbsp;&nbsp;&nbsp;| - Curvature-aware smoothing that preserves mesh shape.<BR>' ...
+        '&nbsp;&nbsp;&nbsp;| - Calls the <CODE>sms</CODE> function with selectable methods.'], ...
+        ['<HTML><B> iso2mesh / Laplacian:</B><BR>' ...
+        '&nbsp;&nbsp;&nbsp;| - Curvature-aware smoothing that preserves mesh shape.<BR>' ...
+        '&nbsp;&nbsp;&nbsp;| - Calls the <CODE>sms</CODE> function with selectable methods.'], ...
+        ['<HTML><B> iso2mesh / Lowpass:</B><BR>' ...
+        '&nbsp;&nbsp;&nbsp;| - Curvature-aware smoothing that preserves mesh shape.<BR>' ...
+        '&nbsp;&nbsp;&nbsp;| - Calls the <CODE>sms</CODE> function with selectable methods.'], ...
+    };
```
rcassani: @tmedani, please update this information window to show the difference among methods

toolbox/anatomy/tess_smooth_select.m
```diff
@@ -0,0 +1,220 @@
...
+if (nargin < 3) || isempty(Params)
+    Params = [];
+end
+if (nargin < 2) || isempty(Method)
+    Method = [];
+    Params = [];
+end
+
+MultipleFiles = [];
+% USAGE: tess_smooth_wrapper(TessMat, ...)
```
tmedani: @rcassani what is here "tess_smooth_wrapper" ?

rcassani: It does not exist. It was the tentative name for `tess_smooth_select()`

toolbox/tree/tree_callbacks.m
rcassani, line 1: This will be added directly on `master`

toolbox/anatomy/tess_smooth_select.m
rcassani: Fixed in [460e617](https://github.com/brainstorm-tools/brainstorm3/pull/822/commits/460e617378a6c95e8dd458ab35869b11df456abb)

toolbox/tree/tree_callbacks.m
rcassani: Added in 6012207c

PR #823, opened 2025-07-24
Update tess_smooth_select.m

tmedani:
Fix: Normalize alpha mapping between Brainstorm and iso2mesh (as suggested by @Edouard2laire )


rcassani:
I understand the reasoning behind this change.

The reason why it was not changed was to be transparent in the arguments to be used in `sms()`.
with this now the user needs to be aware that `1-alpha` happens before calling `sms()`

tmedani:
Indeed, a bit confusing, however, even in the original `sms` function, the value is reverted back. 
So the value inserted in the BST will be the final value used by the final smoothing method. 
Check this line "  https://github.com/fangq/iso2mesh/blob/722bbc37721d84ef90a95a1bf8dd0646133fd57e/smoothsurf.m#L57C1-L57C8

tmedani:
Thanks :)

PR #824, opened 2025-07-24
Import BIDS fixes

Moo-Marc:
Minimal fixes for IntendedFor field of MEG coordsystem.json file, when it contains and array of MRIs instead of just 1, or it's saved as a BIDS URI. 
Also remove extraneous warning for "missing" (not BIDS standard) 'group' column in channels.tsv files.

### DIFF
--- a/toolbox/process/functions/process_import_bids.m
+++ b/toolbox/process/functions/process_import_bids.m
@@ -755,15 +755,10 @@ function [RawFiles, Messages, OrigFiles] = ImportBidsDataset(BidsDir, OPTIONS)
                                 if iscell(sCoordsystem.IntendedFor)
                                     sCoordsystem.IntendedFor = sCoordsystem.IntendedFor{1};
                                 end
-                                % Resolve BIDS URI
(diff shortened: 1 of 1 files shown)


toolbox/process/functions/process_import_bids.m
```diff
@@ -1328,17 +1323,21 @@
...
+% For now, this function is limited to the current BIDS dataset.
+% A warn is shown if such a dataset name is used in the URI.
+% See https://bids-specification.readthedocs.io/en/stable/common-principles.html#bids-uri
+function [OutFile, Msg] = ResolveBidsUri(InFile, BidsDir)
     Msg = '';
-    if ~strncmp(InFile, 'bids:', 5)
-        % Nothing to do.
-        OutFile = InFile;
-        return;
+    if ~strncmp(InFile, 'bids:', 5) || strncmp(InFile, 'bids::', 6)
```
Moo-Marc, line 1337: @rcassani This additional check is a bug: it will generate paths like /main/path/bids::relative/path

PR #826, opened 2025-07-28
[Matlab2024b] remove usage of legend with nargout == 2

Edouard2laire:
Hello, 

Starting from Matlab2024b, matlab is issuing a warning when using the function legend with 2 outpou, as this will be removed in future matlab release. 

Since the outputs were not used it was easy to remove them from the code. 

see https://www.mathworks.com/help/matlab/ref/legend.html

### DIFF
--- a/toolbox/gui/figure_spectrum.m
+++ b/toolbox/gui/figure_spectrum.m
@@ -1613,15 +1613,15 @@ function PlotHandles = PlotAxesButterfly(hAxes, PlotHandles, TfInfo, TsInfo, X, 
     % Plotting the names of the channels
     if ~isempty(LinesLabels) && TsInfo.ShowLegend && ((length(LinesLabels) > 1) || ~isempty(LinesLabels{1}))
         if (length(LinesLabels) == 1) && (length(PlotHandles.hLines) > 1)
-            [hLegend, hLegendObjects] = legend(PlotHandles.hLines(1), strrep(LinesLabels{1}, '_', '-'));
+            legend(PlotHandles.hLines(1), strrep(LinesLabels{1}, '_', '-'));
         elseif (length(PlotHandles.hLines) == length(LinesLabels))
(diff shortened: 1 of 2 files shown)


rcassani:
@Edouard2laire, thanks for the heads-up

I've been using older versions of Matlab (mostly 2020b and 2022b) for sake of verifying backwards compatibility, but it seems, I'll need to run as well the last one, once in a while

Edouard2laire:
If you reorganize the if structure, you can probably remove the condition on isJSDesktop

brainstorm.m
```diff
@@ -98,8 +98,12 @@
         bst_webread('https://github.com/brainstorm-tools/bst-java/raw/master/brainstorm/dist/brainstorm.jar', BstJar);
     end
     % Add Brainstorm JARs to classpath
-    javaaddpath(fullfile(BrainstormHomeDir, 'java', 'RiverLayout.jar'));
-    javaaddpath(BstJar);
+    if usejava('jvm')
```
rcassani: I think we can use better `javachk` and print the error message it returns

rcassani: Done in [80c3b58](https://github.com/brainstorm-tools/brainstorm3/pull/826/commits/80c3b58a6e09a0d834fa9a523ae95c1a17c4e7c7)

rcassani:
@Edouard2laire, this looks good to be merged

Edouard2laire:
Tested on matlab 2025a, with and without jvm. Worked well. 

Question: is java also required in server mode ? I thought it would not since the GUI is not displayed

rcassani:
> Question: is java also required in server mode ? I thought it would not since the GUI is not displayed

There are few cases where we need the JVM for non-GUI stuff. E.g., `sun.misc.BASE64{De,En}coder` for IO `.gii` files and generate the reports, and parsing XML files.

PR #827, opened 2025-07-30
Refine Tetra Fem Mesh

tmedani:
Summary: 
Introduce an interactive GUI that enables users to refine FEM meshes interactively from the GUI. This option would be accessible via the right-click context menu of an FEM mesh node in the database tree > Refine FEM mesh.

Motivation:
This feature is critical for optimizing FEM computations, particularly in cases where localized accuracy is needed (e.g., near electrodes, complex cortical folds, or any user ROI). Targeted mesh refinement helps reduce overall computational load while maintaining precision where it matters most.


tmedani:
@rcassani 
I added a few comments where I need your input and improvement for the code.

PR #829, opened 2025-07-30
Better detect BOLD and NIRS to avoid sign flip

Edouard2laire:
Edit: we should update https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/process/functions/process_source_atlas.m#L196-L201 too


Edouard2laire:
ok, it's ready for review :)

PR #833, opened 2025-09-04
[Java] Remove java from the bandpass process figure

Edouard2laire:
The objective of this PR is to remove the java componnent (that will be removed from Matlab in a coming release) from the panel that displays the filter information. 

The PR has also as objective to see how you would like to proceed with the coming depreciation of some java components. Do we only kleep the matlab version or do we keep both ? 

This was also an occasion for me to try copilot in vs code. it seems to be doing ok. 
The functions used here (uifigure and uilabels were introduced in Matlab2016b)

Before: 

image

After: 

image

Note: if we want to keep more backward compatibility, then uicontrol can be used to display the text (see 
https://github.com/brainstorm-tools/brainstorm3/commit/041f70412cd5c4156cb438f7264c7d46bfaa24bd). However, the information cannot be easily displayed in BOLD: 

image

### DIFF
--- a/toolbox/process/functions/process_bandpass.m
+++ b/toolbox/process/functions/process_bandpass.m
@@ -320,40 +320,40 @@ function DisplaySpec(sfreq)
     strFilter2 = [strFilter2 'Transient (full): <B>' num2str(FiltSpec.order / 2 / sfreq, '%1.3f') ' s</B><BR>'];
     strFilter2 = [strFilter2 'Transient (99% energy): <B>' num2str(FiltSpec.transient, '%1.3f') ' s</B><BR>'];
     strFilter2 = [strFilter2 'Sampling frequency: <B>', num2str(sfreq), ' Hz</B><BR>'];
-    strFilter2 = [strFilter2 'Frequency resolution: <B>' num2str(dF, '%1.3f') ' Hz</B><BR>'];
+    strFilter2 = [strFilter2 'Frequency resolution: <B>' num2str(dF, '%1.5f') ' Hz</B><BR>'];
     
     hFig = HFilterDisplay(Hf,Freqs,Ht,t,FiltSpec.transient,strFilter1,strFilter2,XFreqLim) ; 
 
 end
 
 function hFig = HFilterDisplay(Hf,Freqs,Ht,t,transient,strFilter1,strFilter2,XFreqLim)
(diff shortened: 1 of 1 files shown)


tmedani:
Thanks @Edouard2laire for testing this and your feedback on Copilot :)

I think it is something that we need to do. 

Have you noticed any time difference while using the uifigure versus the Brainstorm version?

Edouard2laire:
> Have you noticed any time difference while using the uifigure versus the Brainstorm version?

I have not measured the time. But I have not felt any measurable difference when using it in comparison to the Java version. 

I added the use of uigridlayout so MATLAB is handling all the resizing automatically and it looks quite fine to me: 
image

rcassani:
Maybe an impertinent questions, buy why do we need `uifigure` for this figure?
This figure does not have interaction nor controls. Would not be simpler a vanilla figure?

Edouard2laire:
i added a benchmark here: https://github.com/Edouard2laire/brainstorm3/commit/4e622ab06c7f2f0dce8108db4e51a23a4b578322


On my mac, matlab 2023b, i have the following results: 
```
Java: 0.14 s
HTLM: 0.13 s
```


so no real difference.

chinmaychinara91:
> so indeed, j HTML might be slightly slower but i dont think it's enough to be annoying

With `timeit` indeed I see a similar observation in Windows as well. Not much of a difference.

Another thing I tried is just measuring using simple `tic` and `toc` (without the 5 sec pause):
```
tic();
HFilterDisplayHtml(Hf,Freqs,Ht,t,FiltSpec.transient,strFilter1,strFilter2,XFreqLim);
t_html = toc();

tic();
HFilterDisplayJava(Hf,Freqs,Ht,t,FiltSpec.transient,strFilter1,strFilter2,XFreqLim);
t_java = toc();

fprintf('Time: \nJava: %.2f s\nHTLM: %.2f s\n', t_java, t_html)
```
the time taken for loading the filter response figure for the first run is as under (tested on MATLAB 2023b and 2024b). 
```
Time: 
Java: 0.55 s
HTLM: 2.84 s
```
The subsequent runs gave almost comparable times though like the ones you have shown above.

While `timeit` is a better way to measure because it does multiple runs of the same function but just wanted to put this observation out there.

PR #838, opened 2025-09-08
[bst_plugin usage] use bst_websave when bst_webread fail

Edouard2laire:
For some reason, on my computer bst_webread fail to recover the plugins stats, but bst_websave works. 

image

Getting close to the 1000th download of nirstorm :)


rcassani:
`timeout` is now an optional argument (with default of 30s) for `bst_webread.m`
The only place, a quicker answer (5 s) is needed is `bst_check_internet()`

Closed by 7867fd44

PR #815, opened 2025-07-11
[Head model] Add the possibility to compute head model for NIRS

Edouard2laire:
Once PR https://github.com/Nirstorm/nirstorm/pull/266 is merged, we should be able to compute head model for nirs directly from the GUI, also allowing using to create head model containing both EEG and NIRS

PR #839, opened 2025-09-09
Display sensitivity map for NIRS

Edouard2laire:
This is the continuation of https://github.com/brainstorm-tools/brainstorm3/pull/815, allowing to display the nirs sensitivity directly from the GUI.

Todo: 
- [x] Display each wavelength separately 
- [x] Add possibility to threshold? -- can be done already using surface button
- [x] Correctly display the source and detector instead of the midpoint 

Optional: 
- [x] Same scale for all sensors (using maximum: custom mode)

Future work: 

- Display channels on top of the sum channel 
- Export a video like this 

https://github.com/user-attachments/assets/dbc56b8b-cfbb-4616-ac56-6dec4ab42c2a

### DIFF
--- a/toolbox/gui/view_leadfield_sensitivity.m
+++ b/toolbox/gui/view_leadfield_sensitivity.m
@@ -278,34 +278,32 @@ panel_surface('SetSizeThreshold', hFig, 1, 1);
                     set (hLabel, 'Visible', 'on');
                 end
             case 'h'
-
-                strHelp = [  ...
-                            '<TR><TD><B>Left arrow</B></TD><TD>Previous target channel (red color)</TD></TR>' ...
-                            '<TR><TD><B>Right arrow</B></TD><TD>Next target channel (red color)</TD></TR>'];
-                if ~isMeg && ~isNIRS
-                    strHelp = [  ...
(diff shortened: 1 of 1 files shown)


toolbox/gui/view_leadfield_sensitivity.m
```diff
@@ -168,7 +185,11 @@
 % Update display
 UpdateLeadfield();
 % Reset thresholds
-panel_surface('SetDataThreshold', hFig, 1, 0);
+if isNirs
+    panel_surface('SetDataThreshold', hFig, 1, 1/100);
```
rcassani, line 189: @Edouard2laire, what about using `0` as with all the other sensitivity plots?

rcassani (approved):
Everything looks go to be merged

toolbox/gui/view_leadfield_sensitivity.m
Edouard2laire: i prefer 1% so it's showing something consistent as when we do NIRS > extract sensitivity from head model; it also results to a visualization that is much closer to what is shown in all the other nirs software.

User were very confused with our display before we put the threshold for display

rcassani: Sounds good to me

Issue #808, opened 2025-06-23
Loading of long duration data after source localization


Edouard2laire:
Hello, 

**Context:** 

Last year, we merged  https://github.com/brainstorm-tools/brainstorm3/pull/733, allowing us to save wMEM map as factors. ImageGridAmp is now a cell array, and c corresponds to the product of the matrix contained in this cell array.  So if ImageGridAmp = { A, B, C} then the full spatio-temporal map can be obtained as A * B * C. 

** Issue:** 
We recently started to apply wMEM on very long data (~ 8 hours, at 10 Hz) and were able to save it nicely in brainstorm database, thanks to the factorization method: ImageGridAmp contains 3 matrix of respective dimension 15002 x 7810, 7810 x 73682 , 73682 x 302001. 

The issue is that when trying to load, or apply any process on the map, then Brainstorm is computing the product, creating an array of dimension 15 002 x 302 001 which is very huge and doesn't fit in memory. 

**Solution: **

Solving the issue for the process might not be feasible, and I am ok writing a specific process for long data but I think we can solve the visualization issue. I am not sure yet how to implement it in brainstorm and would like some help. 

The main issue is that when we want to visualize the map, at let say t = 300s, we first compute the entire spatio-temporal map, and then select t = 300. But if we do the selection of the time first, and then multiply the matrices, then there is no issue. 

Similarly, if we want the time-course of a specific number of vertices, we can first select the vertices in the first matrix, and then do the multiplication leading to a much smaller memory impact. 

Here is a small function that applies this principle: 

```matlab

function [time, ImageGridAmp ] =  getDataMEM(sData, vertex, TimeWindow)
    
    ImageGridAmp = sData.ImageGridAmp;
    if ~iscell(ImageGridAmp)
        ImageGridAmp = {ImageGridAmp};
    end

    % create spatial mapping and reduce the spatial dimension 
    mapping = sparse(1:length(vertex), vertex, 1, length(vertex), size(ImageGridAmp{1},1));
    ImageGridAmp{1}  = mapping * ImageGridAmp{1};

    % create the temporal mapping and reduce the temporal dimension 
    iTime = panel_time('GetTimeIndices', sData.Time, TimeWindow);
    time  = sData.Time(iTime);

    mapping = sparse(iTime,  1:length(iTime), 1, size(ImageGridAmp{end},2), length(iTime));
    ImageGridAmp{end} = ImageGridAmp{end} * mapping;

    % Compute the full matrix
[...]

PR #841, opened 2025-09-15
[memory] keep the sources as cell array for as long a possible (reduce memory)

Edouard2laire:
Hello, 

The goal of this PR is to start implementing the idea described in https://github.com/brainstorm-tools/brainstorm3/issues/808

Since we gained a lot in memory usage by not storing the matrix plainly (nVertex x nTime) but as factors, it can be useful to not regenerate the entire matrix upon loading, but to only expand part of the matrix necessary for the task at hand. 

For example, to display a specific time point, the last matrix can be reduced before doing the multiplication,  greatly reducing the time spent and memory usage. 

This PR is critical when dealing with long nirs data (lets say whole night nirs) while impacting very little short-duration recording. 

Edit:  Maybe a better name could be found for bst_multiply_cells, but I don't have any idea right now.

### DIFF
--- a/toolbox/math/bst_multiply_cellmat.m
+++ b/toolbox/math/bst_multiply_cellmat.m
@@ -56,10 +56,13 @@ end
 
 % Choose direction of association for multiplication
 if nMat == 3
+    % Cost is the number of scalar multiplications
     cost1 = matSizes{1}(1) * matSizes{1}(2) * matSizes{2}(2) + matSizes{1}(1) * matSizes{2}(2) * matSizes{3}(2);  % (A*B)*C
     cost2 = matSizes{2}(2) * matSizes{2}(1) * matSizes{3}(2) + matSizes{1}(1) * matSizes{2}(1) * matSizes{3}(2);  % A*(B*C)
     fromRight = cost2 < cost1;
 else
+    % Heuristic rule
+    % Note, the optimal association approach can be obtained computing the cost recursively
     fromRight = size(X{end}, 2) < size(X{1}, 1);
 end
 


toolbox/core/bst_memory.m
```diff
@@ -2189,17 +2189,39 @@ function LoadResultsMatrix(iDS, iResult)
...
+                usefulInnerIx       = any(ResultsValues{im}, 1);
+                ResultsValues{im}   = ResultsValues{im}(:, usefulInnerIx);
+                ResultsValues{im+1} = ResultsValues{im+1}(usefulInnerIx, :);
             end
+            if ~isempty(iRows)
+                ResultsValues{1} = ResultsValues{1}(iRows, :);
+            end
+            % Compute full results
+            ResultsValues = double(bst_multiply_cellmat(ResultsValues));
+            % Std should be empty if ImageGridAmp is a cell
```
Edouard2laire, line 2219: should we still keep an assert here to get an error if std is not empty?

toolbox/math/bst_multiply_cellmat.m
```diff
@@ -0,0 +1,91 @@
...
+        fromLeft = 1;
+    elseif isOkR
+        fromLeft = 0;
+    else
+        error('Matrices cannot be multiplied in any direction')
+    end
+end
+
+% Indices for multiplication
+ixs = [1:nMat];
```
Edouard2laire: i am wondering if for readablility, if it might not be better to just keep the if/else condition ? it a nice refactor but make the code quite hard to read as it's not obvious on how the multiplication order is done.

toolbox/core/bst_memory.m
rcassani: Yes, the assert is still there (L2210)

toolbox/math/bst_multiply_cellmat.m
rcassani: Do you mean to have an if/else for Left and Right?

Would it help to add a comment inside this `if ~fromLeft`?
Something like "Reverse indices, so multiplications happens from right to left"

Edouard2laire: The code is not doing what we actually want; what we meant in the original code by multiply from left or right is the change between computing A x B x C as (A x B) x C or A x (B x C). Both give the same results; but one might be more efficient depending on the matrix sizes;


Here what is computed is either A x B x C or C x B x A which will lead to different results.

rcassani: Got it, I oversaw the previous multiplication code, and interpreted as really changing the order of the matrices, rather than associative property.

Edouard2laire:
tested. it seems to be working :)

Edouard2laire:
> @Edouard2laire, for associating 3 matrices, now the exact cost is computed,

oh great. thanks :) 

When oppening a map; this is the cost that appears: 
    Cost 1: 11653743944 - cost 2: 473855972337,   cost2/cost1: 40.661265

rcassani:
For the matrix sizes provided above, cost2 < cost1, what are the matrices for the map you are testing?

Also, the costs are computed with the full matrices, however their size could change by indexing the inner dimensions below, thus the costs should be re-computed

Edouard2laire:
i added a print in the function and I think the current code is working fine: 
 
```
fprintf('Mult: (%d,%d) x (%d,%d) x (%d,%d), C1: %d -- C2:%d, fromRight: %d \n', matSizes{1}(1), ...
                                                                            matSizes{1}(2), ...
                                                                            matSizes{2}(1), ...
                                                                            matSizes{2}(2), ...
                                                                            matSizes{3}(1), ...
                                                                            matSizes{3}(2), ...
                                                                            cost1, cost2, fromRight);
  
```


Here is the result: 
```
Mult: (15002,4041) x (4041,10684) x (10684,1), C1: 647857289456 -- C2:103797126, fromRight: 1 
Mult: (15002,4041) x (4041,10684) x (10684,873), C1: 787622642352 -- C2:90614890998, fromRight: 1 
Mult: (15002,4041) x (4041,10684) x (10684,873), C1: 787622642352 -- C2:90614890998, fromRight: 1 
Mult: (15002,4041) x (4041,10684) x (10684,873), C1: 787622642352 -- C2:90614890998, fromRight: 1 
Mult: (15002,4041) x (4041,10684) x (10684,873), C1: 787622642352 -- C2:90614890998, fromRight: 1 
Mult: (15002,4041) x (4041,10684) x (10684,873), C1: 787622642352 -- C2:90614890998, fromRight: 1 
Mult: (15002,4041) x (4041,10684) x (10684,873), C1: 787622642352 -- C2:90614890998, fromRight: 1 
Mult: (15002,4041) x (4041,10684) x (10684,873), C1: 787622642352 -- C2:90614890998, fromRight: 1 
Mult: (15002,4041) x (4041,10684) x (10684,873), C1: 787622642352 -- C2:90614890998, fromRight: 1 
Mult: (15002,4041) x (4041,10684) x (10684,873), C1: 787622642352 -- C2:90614890998, fromRight: 1 
Mult: (15002,4041) x (4041,10684) x (10684,873), C1: 787622642352 -- C2:90614890998, fromRight: 1 
Mult: (15002,4041) x (4041,10684) x (10684,873), C1: 787622642352 -- C2:90614890998, fromRight: 1 
Mult: (15002,4041) x (4041,10684) x (10684,873), C1: 787622642352 -- C2:90614890998, fromRight: 1 
Mult: (15002,4041) x (4041,10684) x (10684,425), C1: 715816589488 -- C2:44113778550, fromRight: 1 
Mult: (73,4041) x (4041,10684) x (10684,10901), C1: 11653743944 -- C2:473855972337, fromRight: 0 
Mult: (15002,4041) x (4041,10684) x (10684,1), C1: 647857289456 -- C2:103797126, fromRight: 1 
Mult: (15002,4041) x (4041,10684) x (10684,1), C1: 647857289456 -- C2:103797126, fromRight: 1 
Mult: (15002,4041) x (4041,10684) x (10684,1), C1: 647857289456 -- C2:103797126, fromRight: 1 
Mult: (15002,4041) x (4041,10684) x (10684,1), C1: 647857289456 -- C2:103797126, fromRight: 1 
Mult: (15002,4041) x (4041,10684) x (10684,1), C1: 647857289456 -- C2:103797126, fromRight: 1 
Mult: (15002,4041) x (4041,10684) x (10684,1), C1: 647857289456 -- C2:103797126, fromRight: 1 
Mult: (15002,4041) x (4041,10684) x (10684,1), C1: 647857289456 -- C2:103797126, fromRight: 1 
```
>>

PR #844, opened 2025-09-19
[IO] Export fiducials

Edouard2laire:
This adds the option to export fiducials from an MRI. This will be useful for the BIDS export. 

I am adding it to a separate PR so it's easier (faster?) to review. Note that this doesn't quite work. It works if I don't convert the coordinate to voxel, but with the conversion, it doesn't. 

What's strange is that the coordinate in the JSON file doesn't match what is in the MRI after importation..


rcassani:
There maybe an overlap with the current development being done by @Moo-Marc

@Moo-Marc, can you check this?

rcassani:
- Fiducials in the `sMri.SCS` as in voxels, not `SCS` so no need to convert
- Also in Brainstorm fiducials as 1-indexed, but BIDS expects 0-indexed, there the difference of `1`

Edouard2laire:
Actually this doesnt make any sense.... 

In process_import_bids,  we do 

```matlab
                    json = bst_jsondecode(jsonFile);
                    sFid = GetFiducials(json, 'voxel');
```

but sFid doesnt seems to be used anywhere...

Edouard2laire:
So i found the first break of matlab2025: 


image



inputing the fiducials is having some issue; and interacting with the buttons is kind of tricky

Moo-Marc:
I forgot it's in a branch already (WIP):
https://github.com/Moo-Marc/brainstorm3/blob/Coordinates/toolbox/io/bst_save_coregistration.m#L91-L199
Note in particular the part about .InitTransf 'reorient', which you may not have.

Again, the idea is just for you to compare the part for exporting MRI fids, so that what we add works for both your and my use cases, and we may catch bugs. 
Thanks!

Edouard2laire:
ok. so this should be ready for review. 


I dont really see thepoint of inverting the init transformation when exporting the fiducials here as we are also exporting the MRI. if a transformation was applied to the MRI during the brainstorm importation, this will also be present in the exported MRI and is not inverted before the export is being done (at least from what i understand)

rcassani:
@Moo-Marc, @Edouard2laire, since it is for fiducials in the *exported* MRI. It makes sense to add the generation of this JSON file inside `export_mri()`

```
export_mri( BstMriFile, OutputMriFile, ExportFiducials)
```

With `ExportFiducials` = `'bids'` to generate the JSON file

toolbox/io/export_mri.m
```diff
@@ -155,6 +171,45 @@ function export_mri( BstMriFile, OutputMriFile )
...
+function export_fiducials_bids(sMri, OutputJsonFile)
+% Export_mri_fiducials: Export a MRI fiducials to a json file.
+% Coordinates are exported in RAS voxel, using a 0-based indexing
+    output = struct();
+    output.CoordinateUnits = 'voxel';
+    % Fiduciala (mm): Subject coordinate system
+    SCS_fieldsname = {'NAS', 'RPA', 'LPA'};
+    for iFields = 1:length(SCS_fieldsname)
+        if isfield(sMri.SCS, SCS_fieldsname{iFields}) && ~isempty(sMri.SCS.(SCS_fieldsname{iFields}))
+            output.FiducialsCoordinates.(SCS_fieldsname{iFields}) = sMri.SCS.(SCS_fieldsname{iFields});
```
rcassani: Should not be `AnatomicalLandmarkCoordinates` instead of `FiducialsCoordinates`?

Edouard2laire: getFiducial function seems to accept both but, yes, it seems the BIDS specification mention only AnatomicalLandmarkCoordinates. https://bids-specification.readthedocs.io/en/stable/modality-specific-files/magnetic-resonance-imaging-data.html#anatomical-landmarks


Fixed

Moo-Marc: For MRI, yes it should be Anat...  That's the only field defined for MRI in BIDS, and it's present in all other modalities as well.  In addition, for EEG and NIRS, BIDS defines FiducialsCoordinates.  For MEG it's HeadCoilCoordinates instead of FiducialsCoordinates.  But again, what I use for coregistration, and the only one for MRI is Anat.  (edit: I now see it's been changed.)

Edouard2laire: Thx. So for my next PR, if i understand well what you wrote in the code, the exported fiducials for recording (EEG, MEG, fNIRS) should come from the MRI and not what is saved in the channel mat ? (eg. sChannel.SCS.NAS)

Moo-Marc: That's a somewhat complex question.

There's some flexibility in BIDS. While it's not explicit on this, I believe it's best to ensure the points are exactly the same in both modalities.  Note that this is not how Brainstorm does it: one can mark anat fids on the MRI that approximately match the points that were digitized, but if we calculate their exact distances, they wouldn't match.  Yet they're "good enough" and we equate the coordinate systems computed from each set of these points.  That's valid, but I feel having the exact same points on both sides will be clearer.

But even with the choice of using the exact same points on both sides, BIDS still would allow us to decide what points to use.  I usually proceed in two steps: after doing the alignment in Brainstorm, I replace the MRI fiducials by the digitized fiducials.  I think this functionality is still not merged, and it probably would only work for CTF MEG, but it will be under Import > ChannelFile > Adjust coordinate systems.  That function corrects other files accordingly: MRI, surfaces, and MEG channel files.  After that, the MRI fids and the digitized fids will match exactly.  However, if a participant has multiple sessions, they would only match for one.

Again, here BIDS allows different solutions.  One is session-labeled fiducials on the MRI side, but I really don't like that as it's not fully defined/standardized.  Instead, I re-use the same MRI fiducials and convert them to each MEG session's raw coordinates. 

In summary, that's why I convert from the MRI fiducials: to ensure we use the exact same points on both sides, and so that it works with multiple MEG sessions with the same MRI.

Moo-Marc: Furthermore, the channel file .SCS is not at all standardized in Brainstorm.  Often, it's empty, depending on the modality and MEG system for example.  I added some code for CTF MEG to update these from the digitized fiducials, otherwise it would not reflect the current alignment after any coregistration (automatic or manual).  Another thing that's not standardized is the naming of the coordinate transformations that are saved in the channel file when importing.  So that could also make it a little more difficult to figure out the proper transformation to convert the MRI fids to other modalities' raw coordinates.  Again, it might be a bit easier if you're exporting, but my application was to export back to the original raw files.

Issue #820, opened 2025-07-21
Brainstorm and matlab2025


Edouard2laire:
Hello, 

i started using MATLAB 2025. So far, I have not encountered any issue, but some warnings that might need some attention: 

```
> In javacomponent (line 112)
In figure_mri>CreateFigure (line 183)
In figure_mri (line 49)
In bst_figures>CreateFigure (line 177)
In bst_figures (line 59)
In view_mri (line 201)
In tree_callbacks>SurfaceCheckAlignment_Callback (line 3312)
In tree_callbacks>@(h,ev)SurfaceCheckAlignment_Callback(bstNodes(1)) (line 1234) 
Warning: You are using a limited version of JAVACOMPONENT. JAVACOMPONENT will be removed in a future release. For more information see [UI Alternatives for MATLAB Apps](https://www.mathworks.com/products/matlab/app-designer/java-swing-alternatives.html)
on mathworks.com.
```

and 

```
Warning: An error occurred while drawing the scene: Error while processing frame: Error: Could not load TeX fonts. Falling back to uninterpreted text. 
> In defaulterrorcallback (line 12)
In bst_progress (line 160)
In view_surface (line 179)
In tree_callbacks (line 206)
In bst_call (line 28)
In panel_protocols>CreatePanel/protocolTreeClicked_Callback (line 125)
In bst_call (line 28)
In panel_protocols>@(h,ev)bst_call(@protocolTreeClicked_Callback,h,ev) (line 75) 

```

I'll update if I find more. It seems the JAVACOMPONENT might be removed in the next release. see; https://www.mathworks.com/products/matlab/app-designer/java-swing-alternatives.html

tmedani:
Hi @Edouard2laire 
Is this for the newest version, 2025b, or 2025a?
Is there any warning or error on the terminal?

I just downloaded 2025b, and I don't have this issue. [we have other issues to check now, but not this one]

Edouard2laire:
it was Matlab2025a. There was the warning about the javacomponent but that's all.  Maybe it's because of my multiple-screen setup. 

> [we have other issues to check now, but not this one]

No problem; let me know if i can help :)

Edouard2laire:
It seems now confirmed that Matlab2026a won't include Java jvm by default : https://www.mathworks.com/help/releases/R2026a/matlab/release-notes.html#mw_760b743a-b1dc-4c98-bc22-0fb7a8d23320 




[...]

PR #846, opened 2025-09-30
Supress  `javacomponent` warnings >=R2025a

chinmaychinara91:
Based on the issue described in https://github.com/brainstorm-tools/brainstorm3/issues/820, this PR aims to suppress those messages.

Screenshot 2025-09-29 174455

MATLAB 2025a onwards the MATLAB desktop and graphics system uses JavaScript and HTML instead of Java (check blog https://blogs.mathworks.com/matlab/2025/05/16/whats-with-all-the-big-changes-in-r2025a/). The warning now falls under the category `MATLAB:ui:javacomponent:BridgeForWebFigures` as compared to `MATLAB:ui:javacomponent:FunctionToBeRemoved` prior to R2025a.

### DIFF
--- a/toolbox/anatomy/mri_editMask.m
+++ b/toolbox/anatomy/mri_editMask.m
@@ -161,8 +161,12 @@ function figureEditMask_OpeningFcn(hObject, eventdata, handles, varargin)
     
     % === Replace Matlab slider by Java slider ===
     % Disable the Java-related warnings after 2019b
-    if (bst_get('MatlabVersion') >= 907)
+    % MATLAB >= 2019b and MATLAB <= 2024b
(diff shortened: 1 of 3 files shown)


Edouard2laire:
LGTM ! i guess we need to start thinking about how brainstorm will handle the transition away from Java.

rcassani:
@chinmaychinara91, this PR just hide the warnings, we need to fix the issue from root.

There is this PR #833 that handles changes for the band pass filter GUI (which in the list of things to review)

tmedani:
> @tmedani, do you want this warnings hidden for your next event?

@rcassani Ill request an earlier version from Mathworks, to avoid any unexpected issue

PR #847, opened 2025-10-06
restaure progress bar after installing fieldtrip

Edouard2laire:
If you try to load a file while fieldtrip is not installed, the progress bar is wrong after the installation and stay as 'installing fieldtrip' during the file loading.

### DIFF
--- a/toolbox/io/in_data_edf_ft.m
+++ b/toolbox/io/in_data_edf_ft.m
@@ -31,7 +31,15 @@ if ~exist('edf2fieldtrip', 'file')
     if ~isInstalled
         error(errMsg);
     end
+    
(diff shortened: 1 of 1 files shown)


rcassani:
@Edouard2laire, thanks for catching this.

The bar issue would apply to all the other plugins that are automatically installed during I/O or other tasks. As such, the get/set `pBarParams` will be better on the plug in install. Committed 8ef94285

rcassani:
@Edouard2laire, `in_data_fieldtrip` is now modified to accept FieldTrip loaded structs, the comment was moved out of that struct to avoid ambiguity on accepting a FieldTrip struct plus a custom field.

PR #848, opened 2025-10-07
Resection labeling in Brainstorm

chinmaychinara91:
This PR integrates the resection-labeling from https://github.com/ajoshiusc/auto_resection_mask/tree/brainstorm-plugin as a Brainstorm plugin

[More details to be updated]

### DIFF
--- a/toolbox/core/bst_plugin.m
+++ b/toolbox/core/bst_plugin.m
@@ -202,6 +202,26 @@ function PlugDesc = GetSupported(SelPlug, UserDefVerbose)
     PlugDesc(end).ReadmeFile     = 'README.md';
     PlugDesc(end).LoadFolders    = {'*'};
     PlugDesc(end).TestFile       = 'process_nmp_fetch_maps.m';
+    
+    % === ANATOMY: RESECTION LABELING ===
+    PlugDesc(end+1)              = GetStruct('resection-labeling');
+    PlugDesc(end).Version        = 'latest';
+    PlugDesc(end).Category       = 'Anatomy';
+    PlugDesc(end).AutoUpdate     = 1;
+    switch(OsType)
+        case 'linux64'
+            PlugDesc(end).URLzip   = 'https://neuroimage.usc.edu/bst/getupdate.php?d=bst_resection_labeling_linux.zip';
+            PlugDesc(end).TestFile = 'resection_labeling';
+        case {'mac64', 'mac64arm'}
+            PlugDesc(end).URLzip   = 'https://neuroimage.usc.edu/bst/getupdate.php?d=bst_resection_labeling_mac.zip';
+            PlugDesc(end).TestFile = 'resection_labeling';
+        case 'win64'
+            PlugDesc(end).URLzip   = 'https://neuroimage.usc.edu/bst/getupdate.php?d=bst_resection_labeling_win.zip';
+            PlugDesc(end).TestFile = 'resection_labeling.exe';
+    end
+    PlugDesc(end).URLinfo        = 'https://github.com/ajoshiusc/auto_resection_mask/tree/brainstorm-plugin';
+    PlugDesc(end).CompiledStatus = 1;
+    PlugDesc(end).LoadFolders    = {'bin'};
 
     % === ANATOMY: ROAST ===
     PlugDesc(end+1)              = GetStruct('roast');
--- a/toolbox/process/functions/process_resection_labeling.m
+++ b/toolbox/process/functions/process_resection_labeling.m
@@ -1,0 +1,221 @@
+function varargout = process_resection_labeling( varargin )
+% PROCESS_RESECTION_LABELING: Deliniate surgical resection mask using pre- and post-op MRIs.
+%
+% USAGE:              OutputFiles = process_resection_labeling('Run',     sProcess, sInputs)
+%         [ResecMaskFile, errMsg] = process_resection_labeling('Compute', iSubject, MriFilePreOp, MriFilePostOp)
+%         [ResecMaskFile, errMsg] = process_resection_labeling('Compute', iSubject, sMriPreOp,    sMriPostOp)
+
+% @=============================================================================
+% This function is part of the Brainstorm software:
+% https://neuroimage.usc.edu/brainstorm
+% 
+% Copyright (c) University of Southern California & McGill University
+% This software is distributed under the terms of the GNU General Public License
+% as published by the Free Software Foundation. Further details on the GPLv3
+% license can be found at http://www.gnu.org/copyleft/gpl.html.
+% 
+% FOR RESEARCH PURPOSES ONLY. THE SOFTWARE IS PROVIDED "AS IS," AND THE
+% UNIVERSITY OF SOUTHERN CALIFORNIA AND ITS COLLABORATORS DO NOT MAKE ANY
+% WARRANTY, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO WARRANTIES OF
+% MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE, NOR DO THEY ASSUME ANY
+% LIABILITY OR RESPONSIBILITY FOR THE USE OF THIS SOFTWARE.
+%
+% For more information type "brainstorm license" at command prompt.
+% =============================================================================@
+%
+% Authors: Chinmay Chinara, 2025
+%          Anand A. Joshi, 2025
+
+eval(macro_method);
+end
+
+
+%% ===== GET DESCRIPTION =====
+function sProcess = GetDescription() %#ok<DEFNU>
+    % Description the process
+    sProcess.Comment     = 'Resection labeling';
+    sProcess.Category    = 'Custom';
+    sProcess.SubGroup    = {'Import', 'Import anatomy'};
+    sProcess.Index       = 42;
+    sProcess.Description = 'https://neuroimage.usc.edu/brainstorm/Tutorials/SegBrainSuite#Resection_labeling';
+    % Definition of the input accepted by this process
+    sProcess.InputTypes  = {'import'};
+    sProcess.OutputTypes = {'import'};
+    sProcess.nInputs     = 1;
+    sProcess.nMinFiles   = 0;
+    % File selection options
+    SelectOptions = {...
+        '', ...                            % Filename
+        '', ...                            % FileFormat
+        'open', ...                        % Dialog type: {open,save}
+        'Import MRI...', ...               % Window title
+        'ImportAnat', ...                  % LastUsedDir: {ImportData,ImportChannel,ImportAnat,ExportChannel,ExportData,ExportAnat,ExportProtocol,ExportImage,ExportScript}
+        'single', ...                      % Selection mode: {single,multiple}
+        'files', ...                       % Selection mode: {files,dirs,files_and_dirs}
+        bst_get('FileFilters', 'mri'), ... % Get all the available file formats
+        'MriIn'};                          % DefaultFormats: {ChannelIn,DataIn,DipolesIn,EventsIn,MriIn,NoiseCovIn,ResultsIn,SspIn,SurfaceIn,TimefreqIn
+    % Option: Subject name
+    sProcess.options.subjectname.Comment = 'Subject name:';
+    sProcess.options.subjectname.Type    = 'subjectname';
+    sProcess.options.subjectname.Value   = '';
+    % Option: Pre-op MRI file
+    sProcess.options.preopmrifile.Comment = 'Pre-op MRI filename:';
+    sProcess.options.preopmrifile.Type    = 'filename';
+    sProcess.options.preopmrifile.Value   = SelectOptions;
+    % Option: Post-op MRI file
+    sProcess.options.postopmrifile.Comment = 'Post-op MRI filename:';
+    sProcess.options.postopmrifile.Type    = 'filename';
+    sProcess.options.postopmrifile.Value   = SelectOptions;
+end
+
+
+%% ===== FORMAT COMMENT =====
+function Comment = FormatComment(sProcess) %#ok<DEFNU>
+    Comment = sProcess.Comment;
+end
+
+
+%% ===== RUN =====
+function OutputFiles = Run(sProcess, sInputs) %#ok<DEFNU>
+    OutputFiles = {};
+    % Get subject name
+    SubjectName = file_standardize(sProcess.options.subjectname.Value);
+    if isempty(SubjectName)
+        bst_report('Error', sProcess, [], 'Subject name is empty.');
+        return;
+    end
+    % Get pre-op MRI filename to import
+    MriFilePreOp = sProcess.options.preopmrifile.Value{1};
+    if (length(sProcess.options.preopmrifile.Value) < 2) || isempty(sProcess.options.preopmrifile.Value{2})
+        FileFormatPreOp = 'All';
+    else
+        FileFormatPreOp = sProcess.options.preopmrifile.Value{2};
+    end
+    if isempty(MriFilePreOp)
+        bst_report('Error', sProcess, [], 'Pre-op MRI file not selected.');
+        return
+    end
+    % Get post-op MRI filename to import
+    MriFilePostOp = sProcess.options.postopmrifile.Value{1};
+    if (length(sProcess.options.postopmrifile.Value) < 2) || isempty(sProcess.options.postopmrifile.Value{2})
+        FileFormatPostOp = 'All';
+    else
+        FileFormatPostOp = sProcess.options.postopmrifile.Value{2};
+    end
+    if isempty(MriFilePostOp)
+        bst_report('Error', sProcess, [], 'Post-op MRI file not selected.');
+        return
+    end
+
+    % ===== GET/CREATE SUBJECT =====
+    % Get subject 
+    [sSubject, iSubject] = bst_get('Subject', SubjectName);
+    % Create subject is it does not exist yet
+    if isempty(sSubject)
+        [sSubject, iSubject] = db_add_subject(SubjectName);
+    end
+    if isempty(iSubject)
+        bst_report('Error', sProcess, [], ['Cannot create subject "' SubjectName '".']);
+        return
+    end
+    % The subject can't be using the default anatomy
+    if (iSubject ~= 0) && sSubject.UseDefaultAnat
+        bst_report('Error', sProcess, [], ['Subject "' SubjectName '" is using the default anatomy (read-only).']);
+        return
+    end
+    % Proceed importing the MRI files only if no anatomy defined for subject
+    if isempty(sSubject.Anatomy)
+        % Import pre-op MRI volume
+        DbMriFilePreOp  = import_mri(iSubject, MriFilePreOp, FileFormatPreOp, 0, 0, 'mri_preop');      
+        % Import post-op MRI volume
+        DbMriFilePostOp = import_mri(iSubject, MriFilePostOp, FileFormatPostOp, 0, 0, 'mri_postop');
+    else
+        bst_report('Error', sProcess, [], ['Subject "' SubjectName '" has anatomy defined. Select an empty subject.']);
+        return
+    end
+
+    % Call processing function
+    [isOk, errMsg] = Compute(iSubject, DbMriFilePreOp, DbMriFilePostOp);
+    % Handling errors
+    if ~isOk
+        bst_report('Error', sProcess, [], errMsg);
+    elseif ~isempty(errMsg)
+        bst_report('Warning', sProcess, [], errMsg);
+    end
+    % Return an empty structure
+    OutputFiles = {'import'};
+end
+
+
+%% ===== COMPUTE RESECTION-LABELING =====
+function [ResecMaskFile, errMsg] = Compute(iSubject, MriFilePreOp, MriFilePostOp)
+    ResecMaskFile = [];
+    errMsg = '';
+
+    % Load pre- and post-op MRI
+    if isstruct(MriFilePreOp)
+        sMriPreOp  = MriFilePreOp;
+        sMriPostOp = MriFilePostOp;
+    elseif ischar(MriFilePreOp)
+        sMriPreOp  = in_mri_bst(MriFilePreOp);
+        sMriPostOp = in_mri_bst(MriFilePostOp);
+    else
+        errMsg = 'Invalid call.';
+        return;
+    end
+    disp(['RESEC_LABEL> pre-op MRI:  ' sMriPreOp.Comment]);
+    disp(['RESEC_LABEL> post-op MRI: ' sMriPostOp.Comment]);
+
+    % Install/load resection-labeling plugin
+    [isOk, errInstall, PlugDesc] = bst_plugin('Install', 'resection-labeling');
+    if ~isOk
+        errMsg = [errMsg, errInstall];
+        return;
+    end
+
+    % === SAVE BOTH MRI AS NIfTI ===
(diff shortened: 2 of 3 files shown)


toolbox/process/functions/process_resection_identification.m
```diff
@@ -0,0 +1,220 @@
...
+    elseif ~isempty(errMsg)
+        bst_report('Warning', sProcess, [], errMsg);
+    end
+    % Return an empty structure
+    OutputFiles = {'import'};
+end
+
+
+%% ===== COMPUTE RESECTION-LABELING =====
+function [ResecMaskFile, errMsg] = Compute(iSubject, MriFilePreOp, MriFilePostOp)
```
rcassani: `iSubject` is not needed, it can obtained from one of the MRI inputs.

Indeed, getting `iSubject` from both MRI will help to tests for:
- Same Subject
- No DefaultAnatomy

toolbox/process/functions/process_resection_labeling.m
```diff
@@ -0,0 +1,220 @@
...
+    if ~isOk
+        errMsg = [errMsg, errInstall];
+        return;
+    end
+
+    % === SAVE BOTH MRI AS NIfTI ===
+    bst_progress('start', 'Resection labeling', 'Exporting pre- and post-op MRI...');
+    % Create temporary folder
+    TmpDir = bst_get('BrainstormTmpDir', 0, 'resection-labeling');
+    % Save pre-op MRI in .nii.gz format (resection-labeling plugin expects compressed .nii)
```
rcassani: In the plugin, `nibabel.load()` is used, to read the image files.

Can we update that code, so we do not need to gzip the volume here, as it will be ungzipped there? 

This will save time and storage.

toolbox/tree/tree_callbacks.m
```diff
@@ -1150,6 +1150,12 @@
                         if (length(iAnatomy) == 1) && iSubject ~=0 && ~isPet
                             gui_component('MenuItem', jPopup, [], 'SEEG/ECOG implantation', IconLoader.ICON_SEEG_DEPTH, [], @(h,ev)bst_call(@panel_ieeg, 'CreateImplantation', filenameRelative));
                         end
+                        % === RESECTION LABELING ===
+                        if (length(iAnatomy) == 2) && ~isCt && ~isPet  % pre and post implant MRIs
+                            MriFilePreOp  = sSubject.Anatomy(iAnatomy(1)).FileName;
+                            MriFilePostOp = sSubject.Anatomy(iAnatomy(2)).FileName;
+                            gui_component('MenuItem', jPopup, [], 'Resection labeling', IconLoader.ICON_MASK, [], @(h,ev)bst_call(@process_resection_labeling, 'Compute', iSubject, MriFilePreOp, MriFilePostOp));
```
rcassani: Icon ` IconLoader.ICON_MASK` does not exist in `brainstorm.jar`, but the one in screenshot looks like `ICON_FEM` maybe we can use this one.

toolbox/tree/tree_callbacks.m
```diff
@@ -1150,6 +1150,12 @@
                         if (length(iAnatomy) == 1) && iSubject ~=0 && ~isPet
                             gui_component('MenuItem', jPopup, [], 'SEEG/ECOG implantation', IconLoader.ICON_SEEG_DEPTH, [], @(h,ev)bst_call(@panel_ieeg, 'CreateImplantation', filenameRelative));
                         end
+                        % === RESECTION LABELING ===
+                        if (length(iAnatomy) == 2) && ~isCt && ~isPet  % pre and post implant MRIs
+                            MriFilePreOp  = sSubject.Anatomy(iAnatomy(1)).FileName;
+                            MriFilePostOp = sSubject.Anatomy(iAnatomy(2)).FileName;
```
rcassani: We may want to reorder the selected MRIs, so the default one (in green) is always the first one. To avoid the confusing result if the user selected the postop MRI first.

toolbox/process/functions/process_resection_identification.m
```diff
@@ -0,0 +1,220 @@
...
+    if isempty(iSubject)
+        bst_report('Error', sProcess, [], ['Cannot create subject "' SubjectName '".']);
+        return
+    end
+    % The subject can't be using the default anatomy
+    if (iSubject ~= 0) && sSubject.UseDefaultAnat
+        bst_report('Error', sProcess, [], ['Subject "' SubjectName '" is using the default anatomy (read-only).']);
+        return
+    end
+    % Proceed importing the MRI files only if no anatomy defined for subject
```
rcassani: Do we want this limitation?

A way to give the user the option to do the same thing with GUI and with Processes would be to:

1. Modify `process_import_mri` to add an option `isAutoAdjust` (I'll handle this in `master`)

2. Then the pipeline would look:
    ```
    Process IMPORT MriFile1 as MRI1
    Process IMPORT MriFile1 as MRI2
    Process RESECT(MRI1, MR2)
    ```

toolbox/core/bst_plugin.m
```diff
@@ -202,6 +202,26 @@
...
+    
+    % === ANATOMY: RESECTION LABELING ===
+    PlugDesc(end+1)              = GetStruct('resection-labeling');
+    PlugDesc(end).Version        = 'latest';
+    PlugDesc(end).Category       = 'Anatomy';
+    PlugDesc(end).AutoUpdate     = 1;
+    switch(OsType)
+        case 'linux64'
+            PlugDesc(end).URLzip   = 'https://neuroimage.usc.edu/bst/getupdate.php?d=bst_resection_labeling_linux.zip';
+            PlugDesc(end).TestFile = 'resection_labeling';
```
rcassani: Need to add execute permission to `resection_labeling` (`linux64` and `mac64*`)

toolbox/process/functions/process_resection_identification.m
```diff
@@ -0,0 +1,220 @@
...
+    disp(['RESEC_LABEL> Computation completed in: ' num2str(toc) 's']);
+
+    % === SAVE OUTPUT RESECTION MASKS ===
+    % Post-op MRI surgical resection mask warped in pre-op space
+    ResecMaskNii  = bst_fullfile(TmpDir, 'preop.resection.mask.nii.gz');
+    bst_progress('text', 'Saving resection mask...');
+    % Reading volumes
+    ResecMaskFile = import_mri(iSubject, ResecMaskNii, 'Nifti1', 0, 0, 'resection_mask');
+    % Delete the temporary files
+    file_delete(TmpDir, 0, 1);
```
rcassani: Why not to delete the files without confirmation?
If we want the user to confirm, the it's better to add a `isInteractive` flag set to `1` when calling `Compute()` from the tree, and set to `0` id called from the process. The last option is needed so users can run scripts.

toolbox/tree/tree_callbacks.m
rcassani: Or do the resection against the Default MRI.

rcassani (changes requested):
It looks fine, there are some details to address:

1. Why is it called labeling? There is no labeling, but masking, or resection identification. 

2. Is there a plotting in calling `resection_labeling`?
I got the messages that some plotting was happening but the figures did not show up. If there is plotting, can we move it to the process code so it plots Brainstorm figures? 

3. See the other individual comments

toolbox/tree/tree_callbacks.m
chinmaychinara91: Yes it should be `ICON_FEM`. Apologies for the typo. Fixed in [a296c90](https://github.com/brainstorm-tools/brainstorm3/pull/848/commits/a296c90552cb8b7126635575ad35ad12bd0799a1)

toolbox/process/functions/process_resection_identification.m
chinmaychinara91: > Why not to delete the files without confirmation?

Yes it is supposed to be that. I had changed it for debugging purposes and forgot to change it back.
EDIT: Check https://github.com/brainstorm-tools/brainstorm3/pull/848/commits/8cf2ab0f69f27774a9f302b29a07ce5a6851c05d

toolbox/process/functions/process_resection_labeling.m
chinmaychinara91: > Can we update that code

Yes we can do that as well. Will work on changing it in the plugin code.
EDIT: Updated the plugin code to handle it. Associated PR: https://github.com/ajoshiusc/auto_resection_mask/pull/4

toolbox/tree/tree_callbacks.m
chinmaychinara91: These sound better. The default MRI can be the pre MRI and the other the post MRI.

toolbox/process/functions/process_resection_identification.m
chinmaychinara91: Your suggestion sounds much better. Definitely, now it's better to not have this limitation.

toolbox/core/bst_plugin.m
chinmaychinara91: Maybe we should have it directly embedded in the binary ? Let me check this.
EDIT: Handled in the binary. Associated PR: https://github.com/ajoshiusc/auto_resection_mask/pull/4

toolbox/process/functions/process_resection_identification.m
chinmaychinara91: Sounds good.

toolbox/process/functions/process_resection_identification.m
rcassani: The process `process_import_mri()` handles now CT and PET volumes. Check [b4b2da3](https://github.com/brainstorm-tools/brainstorm3/commit/b4b2da3dfe76ed1d5baf00f6af1f022903a3c8fd)

rcassani: So, the user can give either the `FileName` or the `Comment` of the PostOp MRI.
And the resection identification will happen against the Default MRI for that Subject

toolbox/tree/tree_callbacks.m
rcassani: The resection identification option will be shown then for all the non-default MRI volumes

toolbox/process/functions/process_resection_identification.m
chinmaychinara91: >So, the user can give either the FileName or the Comment of the PostOp MRI.
And the resection identification will happen against the Default MRI for that Subject

From the `process_resection_identification`, instead of asking for both the pre-op and post-op (as it is now) we just ask for post-op MRI (image below) as it is expected that the pre-MRI is already imported/set as the default MRI ? 
image
If pre-MRI is not present then we give an appropriate message to user that the pre-MRI needs to be imported first. Does this sound good ?

rcassani: What if PostOp MRI is already imported?
(This is the case as calling from the db tree)

chinmaychinara91: ~In that case, like we have the dropdown list for the Subject in the image above, we can have a dropdown list for choosing the post-op MRI from all the non-default MRI volumes~
EDIT: How about just defining a text field for post-op MRI name in the process tab, where user can just type in the name of the postop file from the db tree ? Check https://github.com/brainstorm-tools/brainstorm3/pull/848/commits/e2f3e4da930f50a66f5c0a36e557bce5dd859d66.
image

toolbox/tree/tree_callbacks.m
chinmaychinara91: > Or do the resection against the Default MRI.

Done. Check https://github.com/brainstorm-tools/brainstorm3/pull/848/commits/dfe3e5cfea63cca89c6f37421f637648326e4635.

toolbox/process/functions/process_resection_identification.m
chinmaychinara91: >Indeed, getting iSubject from both MRI will help to tests for:
Same Subject
No DefaultAnatomy

Check  https://github.com/brainstorm-tools/brainstorm3/pull/848/commits/53e160e4c8360e4f3eb35056147b33c5900e3e1c

toolbox/core/bst_plugin.m
```diff
@@ -202,6 +202,26 @@
...
+    PlugDesc(end).Version        = 'latest';
+    PlugDesc(end).Category       = 'Anatomy';
+    PlugDesc(end).AutoUpdate     = 1;
+    switch(OsType)
+        case 'linux64'
+            PlugDesc(end).URLzip   = 'https://neuroimage.usc.edu/bst/getupdate.php?d=bst_resection_identification_linux.zip';
+            PlugDesc(end).TestFile = 'resection_identification';
+        case {'mac64', 'mac64arm'}
+            PlugDesc(end).URLzip   = 'https://neuroimage.usc.edu/bst/getupdate.php?d=bst_resection_identification_mac.zip';
+            PlugDesc(end).TestFile = 'resection_identification';
```
rcassani: Is it the same binary for macOS on Intel?

toolbox/process/functions/process_resection_identification.m
```diff
@@ -0,0 +1,197 @@
...
+    % Get subject 
+    [sSubject, iSubject] = bst_get('Subject', SubjectName);
+    if isempty(sSubject)
+        bst_report('Error', sProcess, [], ['Subject "' SubjectName '" does not exist.']);
+        return
+    end
+    if isempty(sSubject.Anatomy)
+        bst_report('Error', sProcess, [], 'Pre-op MRI does not exist. Import it and define the fiducials.');
+        return
+    end
```
rcassani: This should check for `sSubject.iAnatomy`.
E.g., In a subject with only a CT, `isempty(sSubject.Anatomy)` is `false`

toolbox/process/functions/process_resection_identification.m
```diff
@@ -0,0 +1,197 @@
...
+        bst_report('Error', sProcess, [], ['Subject "' SubjectName '" is using the default anatomy (read-only).']);
+        return
+    end
+    % Get post-op MRI
+    PostOpMriName = sProcess.options.postopmriname.Value;
+    iPostOpMri    = find(cellfun(@(x) ~isempty(regexp(x, sprintf('^%s$', PostOpMriName), 'match')), {sSubject.Anatomy.Comment}));
+    if isempty(iPostOpMri)
+        bst_report('Error', sProcess, [], 'Post-op MRI is either missing or its name is incorrect.');
+        return
+    end
```
rcassani, line 93: Modify to search by Comment OR FileName

toolbox/process/functions/process_resection_identification.m
```diff
@@ -0,0 +1,197 @@
...
+    
+    % Call processing function
+    [isOk, errMsg] = Compute(sSubject.Anatomy(sSubject.iAnatomy).FileName, sSubject.Anatomy(iPostOpMri).FileName);
```
rcassani: Output arguments here do not match the ones defined in the `Compute()`

toolbox/process/functions/process_resection_identification.m
```diff
@@ -0,0 +1,197 @@
...
+        return
+    end
+    % The subject can't be using the default anatomy
+    if (iSubject ~= 0) && sSubject.UseDefaultAnat
+        bst_report('Error', sProcess, [], ['Subject "' SubjectName '" is using the default anatomy (read-only).']);
+        return
+    end
+    % Get post-op MRI
+    PostOpMriName = sProcess.options.postopmriname.Value;
+    iPostOpMri    = find(cellfun(@(x) ~isempty(regexp(x, sprintf('^%s$', PostOpMriName), 'match')), {sSubject.Anatomy.Comment}));
```
rcassani: Use `strcmp` it does support string vs cell

toolbox/process/functions/process_resection_identification.m
```diff
@@ -0,0 +1,197 @@
...
+    postOpNii = bst_fullfile(TmpDir, 'postop.nii');
+    out_mri_nii(in_mri_bst(MriFilePostOp), postOpNii);
+    
+    % === CALL RESECTION-IDENTIFICATION PIPELINE ===
+    bst_progress('text', 'Calling resection-identification...');
+    % Get resection-identification executable
+    ResecExe = bst_fullfile(PlugDesc.Path, PlugDesc.SubFolder, 'resection_identification');
+    if ispc
+        ResecExe = [ResecExe, '.bat'];
+    end
```
rcassani: We have already the executable name in `PlugDesc.TestFile`

toolbox/process/functions/process_resection_identification.m
```diff
@@ -0,0 +1,197 @@
...
+    strCall = ['"' ResecExe '"' ' ' '"' preOpNii '"' ' ' '"' postOpNii '"' ' ' '"' TmpDir '"'];
+    disp(['RESEC_ID > System call: ' strCall]);
+    tic;
+    status = system(strCall);
+    if (status ~= 0)
+        errMsg = 'Error during resection-identification, see logs in the command window.';
+        bst_progress('stop');
+        return;
+    end
+    disp(['RESEC_ID > Computation completed in: ' num2str(toc) 's']);
```
rcassani: Round to integer.
Space for units `100 s` not `100s`

toolbox/process/functions/process_resection_identification.m
```diff
@@ -0,0 +1,197 @@
...
+
+    % === SAVE OUTPUTS ===
+    bst_progress('text', 'Saving outputs...');
+    % Post-op MRI surgical resection mask warped in pre-op space
+    ResecMaskPreOpNii   = bst_fullfile(TmpDir, 'preop.resection.mask.nii.gz');
+    ResecMaskFilePreOp  = import_mri(iSubjectPreOp, ResecMaskPreOpNii,  'ALL-ATLAS', 0, 1, 'preop_resection_mask');
+    import_surfaces(iSubjectPreOp, ResecMaskFilePreOp,  'MRI-MASK', 0, [], [], 'preop_resection');
```
rcassani: If the preOp MRI does not have fiducial set, the surface and MRI will not be aligned.
As the surface is saved always in SCS.

A requirement for the process should be to have SCS in preOp

Errors are shown in the command line
```
IN_TESS> Warning: MRI is missing, or fiducials are not defined.
IN_TESS> Cannot convert surface to Brainstorm coordinate system.
```

toolbox/process/functions/process_resection_identification.m
```diff
@@ -0,0 +1,197 @@
...
+    % The subject(s) can't be using the default anatomy
+    if any([iSubjectPreOp, iSubjectPostOp] == 0)
+        bst_report('Error', sProcess, [], 'Default anatomy cannot be used (read-only).');
+        return
+    end
+    % Make sure both MRIs come from the same subject
+    if (iSubjectPreOp ~= iSubjectPostOp)
+        bst_report('Error', sProcess, [], 'Both MRIs are not from the same subject.');
+        return
+    end
```
rcassani: These checks are already made when calling from the GUI.
Move them to `Run()` to avoid reading the DB multiple times for the same information

toolbox/process/functions/process_resection_identification.m
rcassani: > EDIT: How about just defining a text field for post-op MRI name in the process tab, where user can just type in the name of the postop file from the db tree ?

This was suggested in https://github.com/brainstorm-tools/brainstorm3/pull/848#discussion_r2436952375

rcassani (commented):
Looks good

Few extra comments:
- The permissions to be executable does not seem to be properly set, at least in Linux. Test: download and unzip the `.zip` file

- There is still the legend about plotting, that one is confusing as the generated images are not shown, not even saved. Can you remove the generation of those images from the binaries?

```
nilearn/plotting/find_cuts.py:154: UserWarning: Could not determine cut coords: All voxels were masked by the thresholding. Returning the center of mass instead.
  warnings.warn(
Plotted preop MRI with resection mask for preop.nii
Plotted postop MRI with resection mask for postop.nii
RESEC_ID > Computation completed in: 1256.6037s
```

- About the menu entry.
  Does it fit better under MRI segmentation?

toolbox/core/bst_plugin.m
chinmaychinara91: Just got hands on an Intel Mac and it was not the case. Built a separate binary for Intel and it works well now. So now we have separate binaries each for Linux, Windows, Mac Intel, and Mac AS.

toolbox/process/functions/process_resection_identification.m
chinmaychinara91: Addressed in https://github.com/brainstorm-tools/brainstorm3/pull/848/commits/8020dad4fb5ba937119bf46f092b1a3a2d50acda

toolbox/process/functions/process_resection_identification.m
chinmaychinara91: My bad. Addressed in https://github.com/brainstorm-tools/brainstorm3/pull/848/commits/b48563fa483149d9db9eb6f4b137f53b2d765e49

toolbox/process/functions/process_resection_identification.m
chinmaychinara91: Addressed in https://github.com/brainstorm-tools/brainstorm3/pull/848/commits/5d7e80cc8260f8a7d56a855ef12e0a831c36e6ff

toolbox/process/functions/process_resection_identification.m
chinmaychinara91: Addressed in https://github.com/brainstorm-tools/brainstorm3/pull/848/commits/d4bc2bad2105d8c87a79a3a403141617648b3d53

toolbox/process/functions/process_resection_identification.m
chinmaychinara91: Addressed in https://github.com/brainstorm-tools/brainstorm3/pull/848/commits/200f9cd87941cdf52d73d088dde7c9fc7bd6981d

toolbox/process/functions/process_resection_identification.m
chinmaychinara91: Addressed in https://github.com/brainstorm-tools/brainstorm3/pull/848/commits/8020dad4fb5ba937119bf46f092b1a3a2d50acda

chinmaychinara91:
> The permissions to be executable does not seem to be properly set, at least in Linux. Test: download and unzip the .zip file

The permissions are set now. I set them directly in the server.

> There is still the legend about plotting, that one is confusing as the generated images are not shown, not even saved. Can you remove the generation of those images from the binaries?

Updated the binaries to remove the generation of the images.

>About the menu entry.
Does it fit better under MRI segmentation?

Definitely a better fit. Check https://github.com/brainstorm-tools/brainstorm3/pull/848/commits/98d7f198982635884e40e54f18f243287f2a25d5

toolbox/process/functions/process_resection_identification.m
chinmaychinara91: Check https://github.com/brainstorm-tools/brainstorm3/pull/848/commits/ab9f181ef7ad3596610495bf3573d6783a961e2f

toolbox/process/functions/process_resection_identification.m
chinmaychinara91: Removed them. 

Also,
- The same subject check, no need to move to `Run()` as all things happen there for the same subject.  
- The default anatomy check is also already handled in `Run()`.
- Check https://github.com/brainstorm-tools/brainstorm3/pull/848/commits/8489e74e1b7f540d26e2e54cc5a42b91025bb42d. This could be a case where, in the Process tab, the user might enter the pre-op MRI name in the post-op field.

chinmaychinara91:
Thank you @rcassani :smiley:

PR #851, opened 2025-10-17
Allow overlay on volume of choice

chinmaychinara91:
Based on our discussion in Slack, this PR allows user to choose on which volume they want to overlay rather than just overlaying on the default MRI.
image

If there is just one other volume other than the default MRI then it shows the option to overlay on default MRI.
image

@tmedani


chinmaychinara91:
@rcassani these changes look good.

PR #855, opened 2025-11-07
[process] Add possibility to use cluster_confirm when there is no cluster

Edouard2laire:
Hello, 

The goal of this PR is to be able to use process with cluster_confirm option even when no cluster are present in the file. (option of using cluster being unavailable). I also changed the color of the error to red, when using cluster, so it's a bit more visible.

A similar thing could probably be done for scout_confirm, but it's more rare to have no scout. :) 

**Before:**

image

**After:**

_Cluster confirm:_
image

_Cluster:_ 
image


rcassani:
Closed in 40773f8

PR #857, opened 2025-11-20
Recording acquisition date : save the hour and minute of the recording start

Edouard2laire:
Among the todo list of Brainstorm, there is something that is very interesting to me among the (Display (high-priority)(Part SEEG grant) list) :

> Display time in H:M:S instead of S > If there is t0 in H:M:S instead of S > As an option in Display configuration button>x-axis

This PR is here to start the process of making this possible. Indeed, for now, we only allow the user to save in brainstorm the date of the data acquisition, but we don't save any information on when the data acquisition started during that day. 

Here is what the new UI looks like: 

image

image

Few processes will then need to be updated, notably to load the date and time properly when loading / exporting file (i am notably thinking of snirf file and BIDS import/export). I suggest that we first start with this PR; and once merged, i will go and update where needed in the code; after including the display update. 

Edouard

### DIFF
--- a/toolbox/gui/figure_datetimepicker.m
+++ b/toolbox/gui/figure_datetimepicker.m
@@ -130,7 +130,7 @@ function dt = figure_datetimepicker(initialValue)
             error('Input must be a char or string.');
         end
     
-        str = char(str);
+        str = strtrim(char(str));
     
         % ---- Pattern 1: yyyy/MM/dd (DATE ONLY) ----
         pat_date_only = '^(\d{4})/(\d{2})/(\d{2})$';


Edouard2laire:
> * `figure_datetimepicker` should be able to accept text inputs with and without seconds.


The following calls now work: 

```
figure_datetimepicker()
figure_datetimepicker('2025/03/24')
figure_datetimepicker('2025/03/24 15:34')
figure_datetimepicker('2025/03/24 15:34:30')
```

Edouard2laire:
no, i tried to go full MATLAB directly. How important is it for you that we use Java here? 


> What about indicating this in the label? `Date (YYYY:MM:DD)`

Good idea. i will add it.

Edouard2laire:
Closing. Will be done in #859

PR #859, opened 2025-11-27
Recording acquisition date : save the hour and minute of the recording start

Edouard2laire:
This is #857 but without the new GUI. 

> 
> Among the todo list of Brainstorm, there is something that is very interesting to me among the (Display (high-priority)(Part SEEG grant) list) :
> 
> > Display time in H:M:S instead of S > If there is t0 in H:M:S instead of S > As an option in Display configuration button>x-axis
> 
> 
> This PR is here to start the process of making this possible. Indeed, for now, we only allow the user to save in brainstorm the date of the data acquisition, but we don't save any information on when the data acquisition started during that day. 
> 
> 
> Here is what the new UI looks like: 
> 
> image
> 
> image
> 
> 
> Few processes will then need to be updated, notably to load the date and time properly when loading / exporting file (i am notably thinking of snirf file and BIDS import/export). I suggest that we first start with this PR; and once merged, i will go and update where needed in the code; after including the display update. 
> 
> 
> 
> 
> Edouard
> 

Proposed GUI: 

image

### DIFF
--- a/toolbox/db/db_add_condition.m
+++ b/toolbox/db/db_add_condition.m
@@ -12,7 +12,9 @@ function iStudies = db_add_condition(SubjectName, ConditionName, isRefresh, Date
 %                       If empty or ommitted, asked to the user
 %     - isRefresh     : If 0, tree is not refreshed after adding condition
 %                       If 1, tree is refreshed
-%     - DateOfStudy   : String 'dd-MMM-yyyy', force Study entries created in the database to use this acquisition date
+%     - DateOfStudy   : String 'dd-MMM-yyyy' or 'yyyy-MM-ddTHH:mm:ss'
+%                       force Study entries created in the database to use this acquisition date
+%                       If empty `DateOfStudy = datetime();`
 % OUTPUT: 
 %     - iStudies : Indices of the studies that were created. 
 %                  Returns [] if an error occurs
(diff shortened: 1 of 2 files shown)


Edouard2laire:
Thanks for the update. 


Based on a quick search. There is three place that might break: 

- node_create_db_studies : l 282:                                     c = datevec(sStudy.DateOfStudy); -- https://github.com/brainstorm-tools/brainstorm3/pull/859/commits/5eb08ab220c8eff829c2509d031f3efd952afc7d
- process_noisecov: l286         studyDate = datenum(sStudy.DateOfStudy);
- process_export_bids: l1090:          acqTime = datenum(sStudy.DateOfStudy);
 - process_export_bids: l317 and l360:                     sessionId = datestr(dateOfStudy, 'yyyymmdd');

toolbox/gui/panel_record.m
```diff
@@ -2962,54 +2962,87 @@ function CopyRawToDatabase(DataFiles) %#ok<DEFNU>
...
+        end
+    end
+    % Try to generate datetime object from GUI or argin
+    try
+        newDatetime = datetime(newDatetime, 'InputFormat', inputFormat);
+    catch
+        error('Invalid date format. Input must be ''DD-MMM-YYYY'', ''YYYY-MM-DD hh:mm:ss'', ''YYYY-MM-DD hh:mm''');
     end
+    % New datetime string to convert to char
+    newDatetime.Format = 'yyyy-MM-dd''T''HH:mm:ss';
```
Edouard2laire, line 3038: do we want the new time format to be 'yyyy-MM-dd''T''HH:mm:ss' if the time of day is specified and yyyy-MM-dd' if the time is not specified  or keep yyyy-MM-dd''T''HH:mm:ss all the time ?

rcassani: I think the same format all the time (no pun intended), so we can have all timestamps in a unique format

PR #860, opened 2025-12-02
read video start from json

Edouard2laire:
Add the possibility to save the synchronization time to a json file

### DIFF
--- a/toolbox/io/import_video.m
+++ b/toolbox/io/import_video.m
@@ -123,7 +123,6 @@ for iFile = 1:length(VideoFiles)
     elseif file_exist(jsonFile)
         % Open the file 
         metadata = bst_jsondecode(jsonFile);
-
         if isfield(metadata, 'VideoStart') && ~isempty(metadata.VideoStart)
             VideoStart = metadata.VideoStart;
         end


toolbox/io/import_video.m
```diff
@@ -119,8 +121,16 @@
         end
         % Close the file
         fclose(fid);
+    elseif file_exist(jsonFile)
+        % Open the file 
+        json_text = fileread(jsonFile);
+        metdaData = jsondecode(json_text);
```
rcassani: Can be simplified as:
`metadata = bst_jsondecode(jsonFile);`

toolbox/io/import_video.m
```diff
@@ -119,8 +121,16 @@
...
+        % Open the file 
+        json_text = fileread(jsonFile);
+        metdaData = jsondecode(json_text);
+
+        if isfield(metdaData, 'VideoStart') && ~isempty(metdaData.VideoStart)
+            VideoStart = metdaData.VideoStart;
+        end
     end
-    
+
```
rcassani: Unnecessary change

toolbox/io/import_video.m
```diff
@@ -119,8 +121,16 @@
         end
         % Close the file
         fclose(fid);
+    elseif file_exist(jsonFile)
+        % Open the file 
+        json_text = fileread(jsonFile);
+        metdaData = jsondecode(json_text);
+
```
rcassani: Unnecessary empty line

toolbox/io/import_video.m
```diff
@@ -97,6 +97,8 @@
     [fPath, fBase, fExt] = bst_fileparts(VideoFiles{iFile});
     % Try to find an .smi file corresponding to the video and use it for the synchronization
     SmiFile = bst_fullfile(fPath, [fBase, '.smi']);
+    jsonFile = bst_fullfile(fPath, [fBase, '.json']);
+
```
rcassani: Unnecessary empty line

Edouard2laire:
> * Where is this type JSON file coming from?
> * Any documentation about the JSON file and the `VideoStart` as key? E.g. units?

It's made up by me. The thing is I have an EEG, NIRS, and a video signal. I import all of them and synchronize them. Once they are synchronized, I export them to be imported later and analyzed. 

I work well. Except for the video. Because if I then import the video again, it's no longer in sync with the recordings. (because we don't know when the video starts). The only solution is to export the video start that is saved in Brainstorm. I thought saving it in a json was making sense.

Edouard2laire:
> > Edit2: Maybe i should actually cut the video according to the synchronization so no need for the extra json
> 
> This makes sense, as this is what is happening the the EEG and NIRS recordings. right?

Yes. Ok I will do that then.

PR #864, opened 2025-12-08
BIDS import coregistration

Moo-Marc:
This correctly loads the coregistration for CTF MEG files as it is saved for the OMEGA database, i.e. matching AnatomicalLandmarkCoordinates in the MRI `_t1w.json` file (in 0-indexed voxel coordinates) and MEG `_coordsystem.json` file (in CTF head coordinates, in cm).  Brainstorm already dealt with the MRI correctly, but previously ignored the matching points on the MEG side, applying the usual "implied" coregistration: matching the MRI fiducials with digitized anatomical fiducials if present, else with digitized or MEG-measured head coils. 

Now if needed, a transformation will be added to the MEG, as when we do a manual coregistration. This is applied as usual to the channel positions etc, and saved in the channel file.  An additional history entry is added after the initial "import" entry to note that "Coregistration imported from BIDS files for this session". In OMEGA, we replace the MRI fiducials with the digitized anatomy points from the first session, in which case there will be no additional transformation needed when importing this first session.  The history entry will still be added so it is clear that the data is still aligned, with the mention "(no transformation needed)".  

Since the MEG `_coordsystem.json` file, and coregistration is defined once per BIDS session, empty-room noise recordings, which in principle cannot be coregistered, will still be treated the same and may get a manual transformation. This is inconsequential as channel coordinates in these noise recordings are meaningless.

While for now it only works on CTF MEG data, it was coded in a way that could be extended to other devices and modalities probably without too much work.  But it requires reviewing how Brainstorm deals with each, in particular the transformations that may be generated by Bst to go from "native" coordinates (which would be expected in BIDS files) to Brainstorm's SCS.  Probably many don't have any such transformation and could be added easily.

The code was tested with various types of coregistration, with or without the presence of digitization, multiple sessions, with a mix of sessions that are coregistered or not. Before/after coregistration figures were compared where I first saved coregistration performed in Brainstorm into the BIDS files with a custom function, and then loaded it back with this `process_import_bids`.  (My `bst_save_coregistration` function could also be added as a Bst process eventually, but for now it is still too specific to OMEGA.)

Independent small fixes in 2 other files: 

- cs_compute: logic, using the matrix after attempting to fix it, not before. (Yet also noted that the bad rotation matrix here probably can't be fixed anyway. So no immediate impact.)
- channel_apply_transf: was sometimes saving by mistake a function handle instead of a matrix. In other functions a matrix is expected.

### DIFF
--- a/toolbox/process/functions/process_import_bids.m
+++ b/toolbox/process/functions/process_import_bids.m
@@ -752,7 +752,7 @@ function [RawFiles, Messages, OrigFiles] = ImportBidsDataset(BidsDir, OPTIONS)
                             elseif ~isempty(coordsystemSpace)
                                 electrodesCoordSystem = coordsystemSpace;
                             end
-                            if isfield(sCoordsystem, 'AnatomicalLandmarkCoordinateSystem') %&& ~isempty(sCoordsystem.AnatomicalLandmarkCoordinateSystem)
+                            if isfield(sCoordsystem, 'AnatomicalLandmarkCoordinateSystem')
                                 landmarkCoordSystem = sCoordsystem.AnatomicalLandmarkCoordinateSystem;
                             end
                             % Coordinates can be linked to the scanner/world coordinates of a specific volume in the dataset


toolbox/process/functions/process_import_bids.m
```diff
@@ -733,21 +736,24 @@
...
+                            elseif isfield(sCoordsystem, 'EEGCoordinateSystem') && ~isempty(sCoordsystem.EEGCoordinateSystem)
+                                electrodesCoordSystem = sCoordsystem.EEGCoordinateSystem;
+                            elseif isfield(sCoordsystem, 'MEGCoordinateSystem') && ~isempty(sCoordsystem.MEGCoordinateSystem)
+                                electrodesCoordSystem = sCoordsystem.MEGCoordinateSystem;
+                            elseif isfield(sCoordsystem, 'NIRSCoordinateSystem') && ~isempty(sCoordsystem.NIRSCoordinateSystem)
+                                electrodesCoordSystem = sCoordsystem.NIRSCoordinateSystem;
+                            elseif ~isempty(coordsystemSpace)
+                                electrodesCoordSystem = coordsystemSpace;
+                            end
+                            if isfield(sCoordsystem, 'AnatomicalLandmarkCoordinateSystem') %&& ~isempty(sCoordsystem.AnatomicalLandmarkCoordinateSystem)
```
rcassani: Is the second condition something that was left out for debugging?

Moo-Marc: I don't remember why I disabled it.  But looks like it should work either way.

rcassani: Remove it in [c99eb18](https://github.com/brainstorm-tools/brainstorm3/pull/864/commits/c99eb182b44b4f45aa46999e3233189659db4da4)

PR #865, opened 2025-12-09
[datetime] Improve I/O support for time of acquisition

Edouard2laire:
This PR is here add support for datetime information from various files : 

** EDF** 

- Import/export date from the file
- Improve performance to find the max of the file (use larger block size than 1 seconds)

For the performance, exporting a 2-hour file (34 channels at 250Hz) reduced the time to find the max values from 24 seconds to 1 second. (nblock is decreasing from 7681 to 5)

** SNIRF** 
- modify the reader / writter to support datetime information (according to snirf standard)

** BIDS **
- Read from scan.tsv

 ** FIF **

Read from the file. Note, i tried on a file from https://openneuro.org/datasets/ds000117/versions/1.1.0  but i find the following acquisition date: '1941-56-26'T'09:04:09' i don't think it's right but it make sense based on the information in the file.  I guess the date is wrong but time is ok ? not sure.

### DIFF
--- a/toolbox/io/out_fopen_edf.m
+++ b/toolbox/io/out_fopen_edf.m
@@ -153,8 +153,8 @@ header.nrec = ceil(header.nrec / header.reclen);
             annot = sprintf('+%f', startTime);
             
             if (size(event.times,1) == 2)
-                duration = event.times(2,iEpc) - startTime;
-                annot    = [annot, sprintf('%c%f', char(21), duration)];
+                durationEvt = event.times(2,iEpc) - startTime;
+                annot    = [annot, sprintf('%c%f', char(21), durationEvt)];
             end
             
             annot = [annot, sprintf('%c%s%c%c', char(20), event.label, char(20), char(0))];


Edouard2laire:
ok. it should be ready for review. Note this doesnt attempt to fix the issue raised by Mathilde on the forum.

toolbox/io/in_fopen_edf.m
```diff
@@ -206,8 +206,19 @@
 % No info on bad channels
 sFile.channelflag = ones(hdr.nsignal,1);
 % Acquisition date
-sFile.acq_date = str_date(hdr.startdate);
+acq_date = datetime([ hdr.startdate, ' ', hdr.starttime], 'InputFormat','dd.MM.uu HH.mm.ss');
 
+% In the 'startdate', use 1985 as a clipping date in order to avoid the Y2K problem. 
+% So, the years 1985-1999 must be represented by yy=85-99 and the years 2000-2084 by yy=00-84. 
+% After 2084, yy must be 'yy' and only item 4 of this paragraph defines the date.
```
rcassani, line 213: Not sure what this sentence means:
"After 2084, yy must be 'yy' and only item 4 of this paragraph defines the date."

Edouard2laire: yeah its very obscure to me too. i guess its a problem for people in 2083 to deal with :)

PR #866, opened 2025-12-09
[datetime] Time-series display actual time

Edouard2laire:
This is the first POC for showing the actual time of the recording. 

image

Most of the other time are not changed. Should it be converted to time too ? 

The events for example: 
image

I think it's not necessary. what do you think ? 

When we display a raw signal; and an imported signal, do we want to change the x-axis to hours for both figure ? 
Example: 

image

### DIFF
--- a/toolbox/gui/figure_timeseries.m
+++ b/toolbox/gui/figure_timeseries.m
@@ -4371,7 +4371,7 @@ function UpdateLabelXAxis(iDS, iFig, display_mode)
             new_labels{iLabel} = char(previous_tick(iLabel));
         end
         
-        hAxes.XTickLabel = new_labels;
+        xticklabels(hAxes, new_labels); 
         TsInfo.XMode = 'time';
     end
     setappdata(hFig, 'TsInfo', TsInfo);


toolbox/gui/figure_timeseries.m
```diff
@@ -127,6 +127,10 @@ function CurrentTimeChangedCallback(iDS, iFig)
         % Move text cursor to current time frame
         set(DisplayHandles(1).hTextCursor, 'String', textCursor);
     end
+
+     SetTimeToHour(iDS, iFig);
```
Edouard2laire: %HACK

toolbox/gui/figure_timeseries.m
```diff
@@ -4315,6 +4326,58 @@ function SetResolution(iDS, iFig, newResX, newResY)
...
+        hAxes.XTickLabel = new_labels;
+        TsInfo.XMode = 'onset';
+    % Display x-label as actual time HH:MM:SS
+    else
+        start_file = datetime(GlobalData.DataSet(iDS).Measures.sFile.acq_date);
+        if isempty(start_file)
+            start_file = datetime('17:00:00', 'InputFormat','HH:mm:ss');
+        end
+    
+        previous_tick = start_file + duration(0, 0, hAxes.XTick);
```
rcassani: Check that `duration` is not used

toolbox/gui/figure_timeseries.m
```diff
@@ -4315,6 +4326,58 @@ function SetResolution(iDS, iFig, newResX, newResY)
...
+        previous_tick.Format = 'HH:mm:ss';
+        
+        new_labels = cell(length(previous_tick), 1);
+        for iLabel = 1:length(new_labels)
+            new_labels{iLabel} = char(previous_tick(iLabel));
+        end
+        
+        hAxes.XTickLabel = new_labels;
```
Edouard2laire:
```suggestion
        xticklabels(hAxes, new_labels); 
```

PR #870, opened 2025-12-14
Anat: Add context menu and function to rename FEM elements

tmedani:
Add option to rename FEM elements inside a surface.
Surface can be either a new primitive surface or an anatomy surface from the same Subject

### DIFF
--- a/toolbox/anatomy/fem_rename_elem.m
+++ b/toolbox/anatomy/fem_rename_elem.m
@@ -105,10 +105,26 @@ end
 % Get surface file
 iComment = find(strcmp(surfSelectComment, surfComments), 1);
 if iComment <= length(surfGeoComments)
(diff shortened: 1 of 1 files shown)


rcassani (commented):
@tmedani, these are my comments looking and using this new feature.
I'll work to address them. It would be nice to have you input in the one that is a question.

- [x] Rename function to indicate something that it does

- [x]  Use primitive generation, interactive, so it's in line

- [x]  Save the surface that was used to rename the FEM elements.
   Do not modify existent surfaces, save a copy

- [x]  Show surfaces in the same order as in the GUI

- [x]  Merge tissues in case that the label already exist

rcassani:
@tmedani, the changes are now done.
Test it, then we can do a final review before merge

PR #872, opened 2025-12-16
[align sensors] Select study when applying transformation to other study

Edouard2laire:
When doing the co-registration and brainstorm, ask me if i want to apply co-registration to other sensors in the database, i am always scared as it doesn't tell me what file will be modified. (especially when having multiple sessions within the same subject). 

So now brainstorm ask which file should be modified so the user can check.

### DIFF
--- a/toolbox/sensors/channel_align_manual.m
+++ b/toolbox/sensors/channel_align_manual.m
@@ -997,6 +997,7 @@ function CopyToOtherFolders(ChannelMatSrc, iStudySrc, Transf, iChannels)
     end
     % Apply transformation
     if ~isempty(ChannelFiles)
+        StudyNames = regexprep(StudyNames, '^@raw', '[RAW] ');
         isFileSelected = java_dialog('checkbox', 'Select the folders for the channel files to apply the same transformation:', ...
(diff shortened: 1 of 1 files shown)


Moo-Marc:
Good idea. To clarify though: it's not "other sensors", but "other studies" in Brainstorm language, i.e. "runs", "recordings", or "datasets".

rcassani:
Yes, I will change the dialog texts so they make sense the updated behaviour

Moo-Marc:
A few thoughts about this:

> if you export the run (snirf in this case); then reimport all the runs; then the automatic co-registration is only working for the one run with headpoints

I'm not familiar with snirf, but for CTF MEG, we have the head points file along side the MEG runs (as per BIDS), and it's imported for each recording and will be in each channel file.  Might it just be a case that the files need to be organized/named differently before importing?  Or are your head points really only saved in such a way that they're linked to a single run?  If so, is there a way to adjust the import function instead of having an extra step the user needs to do?

rcassani:
Would it make sense in your workflow to copy the headpoints (even before importing in Brainstorm) from the "seed" `.nsirf` file to the the other `.snirf` files that should shared those points?

PR #876, opened 2026-01-17
Add option for FEM mesh generation (supporting non-nested surfaces with cases that intersect )

tmedani:
Introduces a new 'iso2mesh-2026' method for FEM mesh generation from triangulated surfaces, supporting non-nested surfaces with cases that can intersect with manual tissue identification. Update method selection, descriptions, and user dialogs to include and explain the new option.


tmedani:
some examples: mesh with multiples tissues including intersections 
image

PR #880, opened 2026-02-02
IO: Add support and viz for actual time in raw and imported recordings

rcassani:
This PR has as goal to add the necessary elements to show time axis as actual time for raw and imported recordings

Tasks:

- [x] Change back `Study.DateOfStudy` to be dd-MMM-yyyy. Introduced in https://github.com/brainstorm-tools/brainstorm3/commit/c82f5364e1de59b533a2ac3364c0701f2224efd3
- [x] Add acquisition time as `t0` field in `sFile` and, field `T0` in `datamat`
- [x] Update format of `t0` to be: `'yyyy-MM-ddTHH:mm:ss.SSS'`
- [x] Save `t0` on importing raw: ✅EDF, ✅CTF, ✅SNIRF, ✅BST-DATA, ✅FIF, BIDS (reverted, just date)
- [x] Use `t0` when exporting data: ✅EDF, ✅SNIRF
- [x] Keep and update (if needed) `t0` for imported data (time, epochs or events)
- [x] Allow actual time display for raw and imported data
- [x] Update BST binary format to write and read the new `t0`
- [x] Import `t0` from `Study.DateOfStudy` if available. This is for raw data linked after https://github.com/brainstorm-tools/brainstorm3/commit/c82f5364e1de59b533a2ac3364c0701f2224efd3
- [x] Clean up

Co-authored-by: Edouard2laire <edouard.delaire@icloud.com>


rcassani:
@Edouard2laire, the tasks for this PR are now complete.
Can you take a look to it and how it behaves? 

To update your Protocol with a timestamp in `Study.DateOfStudy` just reload the entire Protocol, it will be automatically fixed to the having `Study.DateOfStudy` as `dd-MMM-yyyy` and `t0` as `yyyy-MM-ddTHH:mm:ss`. (You may want to test in copy of the Protocol, as there is not turning back).

rcassani:
@Edouard2laire, thanks, good catch

> is storing ms accuracy value in T0 important ? The users wont probably have access to such accuracy in their time of recording anyway.

While they may not have this resolution for the `t=0s`, it is necessary on the imported data as there the `t=0s` is liked to the event time, which has the resolution of one sample

toolbox/process/functions/process_import_bids.m
```diff
@@ -691,7 +691,7 @@
                                 end
                                 fDate = datetime('now');
                             end
-                            fDate.Format = 'yyyy-MM-dd''T''HH:mm:ss';
```
Edouard2laire, line 694: i am surprised. Shoudln't we then still import t0 ?

rcassani: This `tsvDates` are used to set the acquisition date for the Study. So, we are keeping only the date.

With this PR, the behaviour at importing a BIDS dataset is that the `T0` is obtained from the from the content of the recording files. 

One thing to consider in the subsequent PR with the processes for `T0`, would be to add a call to a process (after the raw files have been added) to overwrite the `T0` obtained from the recording file with the one in `_scans.tsv`. For example here: 
https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/process/functions/process_import_bids.m#L891

PR #881, opened 2026-02-08
Add directional sensitivity selection

tmedani:
Introduce leadfield sensitivity to be computed for all directions or per-axis (X/Y/Z).
 Add keyboard handling: when viewing the sum of channels (iChannel==0).
- Compute normLF differently based on iSensitivity (full vector norm or sum of absolute components for the chosen axis).
- Update legend text to reflect the selected sensitivity mode and adjust minor ordering of calls.

### DIFF
--- a/toolbox/gui/view_leadfield_sensitivity.m
+++ b/toolbox/gui/view_leadfield_sensitivity.m
@@ -481,7 +481,7 @@ panel_surface('SetSizeThreshold', hFig, 1, 1);
         elseif isAvgRef
             strTitle = [strTarget '  |  Average reference'];
         else
-            strTitle = [strTarget , sprintf('  |  Reference #%d/%d : %s (green)', iRef, length(Channels), Channels(iRef).Name)];
+            strTitle = [strTarget, sprintf('  |  Reference #%d/%d : %s (green)', iRef, length(Channels), Channels(iRef).Name)];
         end
         if (iChannel == 1) && (length(Channels) > 1)
             strTitle = [strTitle, 10 '[Press arrows for next/previous channel (or H for help)]'];


Edouard2laire:
For nirs, we already only save the normal componnent so the full vector norm is the same as selecting only the component normal the cortical surface).  But i guess if we constrain dipole to be normal to the cortical surface for EEG/MEG then it make sense to see the sensitivity when selecting this direction.

tmedani:
> For nirs, we already only save the normal componnent so the full vector norm is the same as selecting only the component normal the cortical surface). But i guess if we constrain dipole to be normal to the cortical surface for EEG/MEG then it make sense to see the sensitivity when selecting this direction.

I added this option, and it indeed turns out to be very useful.
I checked the visualization for MEG, and we can clearly see the expected MEG sensitivity pattern: MEG is much more sensitive to tangential sources than to radial sources.
image

toolbox/gui/view_leadfield_sensitivity.m
```diff
@@ -215,6 +223,16 @@ function KeyPress_Callback(hFig, keyEvent)
                         iChannel = length(Channels);
                     end
                 end
+                if (iSensitivity > 1)
+                    iChannel = 0;
+                end
+                if (iChannel == 0)
+                    iSensitivity = iSensitivity - 1;
+                    if (iSensitivity <= 0)
+                        iSensitivity = 0;
```
Edouard2laire:
```suggestion
                        iSensitivity = 1;
```

toolbox/gui/view_leadfield_sensitivity.m
```diff
@@ -224,8 +242,20 @@ function KeyPress_Callback(hFig, keyEvent)
...
                     iChannel = 0;   % Sum of all channels
                 end
+
+                if (iSensitivity > 1)
+                    iChannel = 0;
+                end
+                if (iChannel == 0)
+                    iSensitivity = iSensitivity + 1;
+                    if (iSensitivity > 5)
+                        iSensitivity = 0;
```
Edouard2laire:
```suggestion
                        iSensitivity = 1;
```

Edouard2laire:
Also, wouldn't it be better to assign a new keystroke for changing the direction (Maybe d?) so you could change direction for an individual channel. 


Also, what do you think of this sensitivity map? it's also MEG (BEM, openmeg model, with one layer model -- Brain with conductivity 0.33)

image


Overlapping spheres seems to be giving much better results: 

image

toolbox/gui/view_leadfield_sensitivity.m
```diff
@@ -204,6 +222,7 @@ function KeyPress_Callback(hFig, keyEvent)
         switch (keyEvent.Key)
             % === LEFT, RIGHT, UP, DOWN: CHANGE CHANNEL ===
             case 'leftarrow'
+                directionOfSensitivity = 1;
```
Edouard2laire:
```suggestion
```

i don't think we should reset this setting every time we switch sensors. Typically user will choose an orientation and then go over different sensors.

rcassani: Agree

tmedani: I was thinking the other way... 
I agree with this change so we can set the direction and apply it to all sensors. 
I'll remove the reset for all the keyboards.

tmedani:
> ok; i need to investigate; but i found something weird with nirs:

@Edouard2laire thanks for the fixe, 
I checked and indeed, the second reshape is not doing a correct job for the normal orientation.
Now it looks much better. 

This is ready to merge @rcassani

Edouard2laire (approved):

toolbox/gui/view_leadfield_sensitivity.m
rcassani: Solved in [906c860](https://github.com/brainstorm-tools/brainstorm3/pull/881/commits/906c86015fb4e60c9bd876eb14742cb3f10124ef)

PR #897, opened 2026-03-08
Multiple progress bar in Brainstorm

Edouard2laire:
Hello, 

3 years after posting https://github.com/brainstorm-tools/brainstorm3/issues/614, I finally came up with a simple, working design for having multiple progress bars in brainstorm without changing the progress bar API.
The idea is to have one window per progress bar and to keep track of which function invoked each progress bar, to be able to open/close/update the appropriate window. 

This should make the management of the progress bar much easier, as each function won't have to worry if a progress bar is already open or if the function they are calling might have a progress bar.  This PR only focuses on the change in the progress bar; we might then improve where it is used within brainstorm to provide better info to the user :) 

Here is a list of few example i used to test: 

**1st example:** 

```matlab
function main()
    bst_progress('start', 'External Loop', '', 0, 10);
    for iSubject = 1:10
        
        internal_loop();
        bst_progress('inc', 1);
    end
    bst_progress('stop');
end


function internal_loop()

    N = 20;    
    bst_progress('start', 'Internal Loop', '', 0, N);
    for i = 1:N
       pause(0.1)
       bst_progress('inc', 1);
    end

    %bst_progress('stop');

end

```

**Example 2**:   Note that the progress bar of 'internal_A' is closed automatically, as the function is not calling bst_progress('stop'); so there is no risk of ending up with a dozen of open progress bars :)

```matlab
function main()

    internal_A();
    internal_B()
end


function internal_A()

    N = 20;    
    bst_progress('start', 'Internal Loop A', '', 0, N);
    for i = 1:N
       pause(0.1)
       bst_progress('inc', 1);
    end

    %bst_progress('stop');

end



function internal_B()

    N = 20;    
    bst_progress('start', 'Internal Loop B', '', 0, N);
    for i = 1:N
       pause(0.1)
       bst_progress('inc', 1);
    end

    %bst_progress('stop');

end

```

**Exemple 3:  with parfor loops** 
```matlab

    N = 20;
    q = parallel.pool.DataQueue;
    afterEach(q, @(x) bst_progress('inc', 1));

    bst_progress('start', 'Internal Loop', '', 0, N);
    parfor i = 1:N
        pause(1+1.2*abs(randn()))
        send(q, i); 

        fprintf('%d \n', i)
    end

    bst_progress('stop');

```

**Example with multiple porgress bar within the same function:** 

Note: there can be a little risk, if     bst_progress('stop'); is not called inside the loop, but i guess that's fine. If we don't want that, then we can change  ``` ix            = find(strcmp(progress_list, caller_name), 1, 'last'); ``` to ```        ix            = find(strcmp(progress_list, caller_name), 1, 'first'); ```

```matlab

N = 20;    
bst_progress('start', 'Externl Loop', '', 0, N);
for i = 1:N
    bst_progress('start', 'Internal Loop B', '', 0, 10);
    for j = 1:10
        pause(0.1)
        bst_progress('inc', 1);
    end
    bst_progress('stop');
    bst_progress('inc', 1);
end
bst_progress('stop');

```

**Exemple when using MEM:** 
image


Edouard2laire:
Todo; 

- check import freesurfer. It create a dozen progress bar

Edouard2laire:
ok. this is ready for review.

I added a limit on the number of progress bar so it is safe to use.  Latter we can then optimize how we use the progress bar over the toolbox to improve the feedback given to the users.

toolbox/io/import_mri.m
```diff
@@ -555,7 +553,7 @@
 db_save();
 % Unload MRI (if a MRI with the same name was previously loaded)
 bst_memory('UnloadMri', BstMriFile);
-
+bst_progress('stop')
```
Edouard2laire:
```suggestion
bst_progress('stop');
```

Edouard2laire:
Any chance we could have this PR merge soon? We might discover a few places after the merge where the progress bar can be improved, for sure. But it will be easy to improve with user feedbacks.

PR #898, opened 2026-03-10
[IO] Load kernel data

Edouard2laire:
Hello, 

Those are minor fix to load nirs KERNEL data. Also remove the automatic computation of HbT as its not really usefull and can be confusing for users to see HbT even if its not present in their data. 

This PR also adds support for time-domain nirs (without importing additional moments for now as i dont know how to process them yet). 

Also fix unit for raw data (mV)


rcassani:
> The issue is also that it doesnt work

Oh, that was unexpected. Let me give it a try to fix otherwise we can remove it as suggested.
Please send me NIRS file to reproduce the images shown.

> Becauuse, we already compute HbT, inside nirstorm, i think this was never used.

That's true, so that montage does not have any effect in recordings that have already [HbT]

PR #908, opened 2026-04-08
Bugfix: `TPM.nii` lookup path while skull stripping

chinmaychinara91:
This PR addresses the skull-stripping issue reported on the forum::
https://neuroimage.usc.edu/forums/t/spm12-skull-stripping-tissue-probability-map-error/56326

The problem mainly affects the standalone Brainstorm build when running skull-stripping and SPM12 is not installed in the `plugins` folder. In the current implementation, calling `bst_get('SpmTpmAtlas', 'SPM')` restricts the search to the SPM plugin location. If `TPM.nii` is not found there, the function returns empty, which causes skull stripping to fail.

This change replaces that call with `bst_get('SpmTpmAtlas')`, which searches across all valid Brainstorm-accessible locations for `TPM.nii`. As a result, Brainstorm can resolve the template more robustly in standalone setups and avoid failures when the SPM plugin folder is not present.

I have verified the standalone build at my end.

### DIFF
--- a/toolbox/anatomy/mri_skullstrip.m
+++ b/toolbox/anatomy/mri_skullstrip.m
@@ -163,7 +163,7 @@ switch lower(Method)
         % Reset matlabbatch to start fresh
         clear matlabbatch;
         % Get the TPM atlas
-        TpmFile = bst_get('SpmTpmAtlas', 'SPM');
+        TpmFile = bst_get('SpmTpmAtlas');
         % Get the SPM tissue segments
         [~, TpmFiles] = mri_normalize_segment(sMriRef, TpmFile);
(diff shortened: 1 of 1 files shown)


rcassani:
@chinmaychinara91, this is now solved at https://github.com/brainstorm-tools/brainstorm3/commit/5a92c928c7301fe43e2e6272e9d2e3ae2d291ef8

The idea is not only to search for any available copy of `TPM.nii`, but to download it if needed (as we do for MNI normalization). For example the case of doing skull-stripping on a volume without MNI normalization first. The same goes for the call in `process_fem_mesh.m`

chinmaychinara91:
I see. Thanks for the fix @rcassani.

PR #911, opened 2026-04-21
[1/3] Fastgraph: Customize SPES (Nihon Kohden)

chinmaychinara91:
This PR is the first part required for [Fastgraph](https://pubmed.ncbi.nlm.nih.gov/32086098/). It involves customizing the raw data.

# Customize SPES (Nihon Kohden)
Linked to `process_customize_spes_nk` and it does the following:

- Finds Nihon Kohden "Stim Start" and "Stim Stop" events
- If provided, renames those stimulation block events
- Detects individual analog stimulation trigger pulses within each stimulation block
- If provided, adds a fixed time offset to stimulation trigger event
- If provided, splits detected trigger pulses into ODD and EVEN events, useful for alternating monophasic stimulation study (useful for studying Fastgraph plots - 3rd PR).

## The GUI
image

## Inputs
- **<ins>Update stimulation start/stop event label</ins>**: Gives an option to the user to rename the `Stim Start` and `Stim Stop` labels in the imported NK events if they want to. In this case, we are shortening the labels to `SB` and `SE` respectively. Came as a requirement from UTH as the channel names can get very long, thereby making the overall stimulation event label very long.
-  **<ins>Stimulation trigger channel</ins>**: The stimulation trigger channel in the recordings (here, `DC10`)
-  **<ins>Update stimulation trigger label</ins>**:  Gives an option to the user to rename the stimulation trigger event label (here `DC10`) in the raw NK events if they want to (here, change to `STIM`). Came from UTH, but maybe some other centers might have their own way of naming, so good to have this option.
- **<ins>Buffer time around stimulation block</ins>** (in s):  Required so that it contains the full stimulation segment plus some context before and after it (here, `2s`). Used as a buffer for the time window for detecting the stimulation trigger. 
- **<ins>Trigger time offset</ins>**: Explained in the GUI (see image above)
- **<ins>Add 'ODD' and 'EVEN' events</ins>**: If we check this, then it creates alternating monophasic stimulations from the original stimulation event. This is useful in downstream analysis as the odd and even pulses may produce different responses. We will be using this for Fastgraph plots (3rd PR).

## Steps to use
1. Load in the updated [example test protocol](https://drive.google.com/file/d/1FRylzuAFLCMs584ol9011KdNx3p5XGRE/view?usp=sharing) [ **NOTE**: This preprocessed protocol involves running CAT12 segmentation to get all the anatomical information, the implantation done using Brainstorm, and then loading in the raw data and adding the locations from the implantation to it. These steps will be part of the tutorial].

2. Drag the `Baseline > Link to raw file` to the `Process1` tab and click on `Run > Stimulation > Customize SPES (Nihon Kohden)`.

3. Fill in the values as seen in the GUI figure above. Click `Run`. The raw data gets updated as under:
Screenshot 2026-04-30 103101<br>

- `Stim Start` event label gets replaced by `SB` (short for stim begin - UTH follows this)
- `Stim Stop` event label gets replaced by `SE` (short for stim end - UTH follows this)
- `DC10` triggers were detected, and the event was renamed as `STIM`
- `-1ms` trigger offset is applied to the stimulation trigger events after studying the sample raw data.
- `ODD` and `EVEN` events was created from the `STIM` events above
- **NOTE**: `STIM`, `ODD` and `EVEN` event labels are appended with the stim site information (e.g. `T'3-T'4 4 #1` in the figure above where `T'3-T'4 4` means stimulation was delivered between contacts `T'3` and `T'4` at 4 milliamps and the `#1` means the first session. You can see in the image there is also a 2nd session which is `T'3-T'4 4 #2`. Analyzing each of these sessions separately is required during the downstream analysis, which will be clearer in the 3rd PR.

4. We can then just import the data by dragging the updated `Baseline > Link to raw file` to the `Process1` tab and click on `Run > Import > Import recordings > Import MEG/EEG: Events`. Set event name as `SB` and epoch time `-2000` to `32000`. For this sample data, we will be importing all `SB` events with an epoch of 34s (30s stimulation block with a buffer of 2s around it so that it contains the full stimulation segment plus some context before and after it). A sample below of what it looks like after importing.
image<br>

## Associated PRs
 1. https://github.com/brainstorm-tools/brainstorm3/pull/912
 2. #916 

@jcmosher @Nastaranlotfi @yashvakilna


Nastaranlotfi:
image

@rcassani , We need to import the stim blocks first, in order to determine which stim rates we have and stim lines. After the blocks are loaded, then we can detect the events. If we detect events first in the raw file, we get way too many extraneous events throughout the raw file. So restricting the event detection AFTER the stim block loading is strongly preferred.

rcassani:
Hi @Nastaranlotfi, I do not really understand what is shown the shared screenshot. Can you provide more context on what is exactly what cannot be done in the raw data, but that is possible to be done with in the imported data?

rcassani:
@chinmaychinara91, from yesterday's discussion doing the event labelling on the raw file is straight forward, right?

chinmaychinara91:
> @chinmaychinara91, from yesterday's discussion doing the event labelling on the raw file is straight forward, right?

@rcassani It is straight forward, but I observed something, let me explain in detail to make it clearer.

1. As per our discussion and your recommendation, in the raw file for the example test protocol provided above, the trigger line used is `DC10` and we can correctly get these events `Run > Events > Detect Analog triggers`.

Screenshot 2026-04-23 181709

2. We then create two copies of the events `DC10`, one called `ODD` that will contain the odd instances of `DC10` event (delete the even instances in the ODD group) and the other called `EVEN` that will contain the even instances of the `DC10` event (delete the odd instances in the EVEN group). Two approaches here:
 - If we do this ODD-EVEN creation in the raw itself and then import the data (`Run > Import > Import recordings > Import MEG/EEG events` with `Stim Start` as the event), the problem is that the ordering of imported ODD and EVEN per stimulation block is incorrect, as the ODD and EVEN are created considering the entire `DC10` for the whole recording in the raw and then imported. As you can see in the time series image below, the ordering of EVEN and ODD triggers should be reversed. 

Screenshot 2026-04-23 184023

- Now if we create the ODD and EVEN after importing the data, we are creating these events based on the local DC10 event per stimulation block. So this gives the correct ordering.

Screenshot 2026-04-23 185538
<br>
The screenshot shared by @Nastaranlotfi is for another proprietary data that UTH has, but I feel with the 2nd approach (i.e. creating ODD EVEN after importing the data) it should resolve all the issues.
<br><br>

Also, since we are creating this DC10 event in the raw itself, so we can remove the data import process 
(`process_import_data_event`) and detection of triggers (`process_evt_detect_analog`) out of the current Load SPES script, as the users can separately do that first. 

We can now call this routine something like `Preprocess SPES` rather than `Load SPES` as it will majorly involve renaming the Stim Starts and Stim Stops and creating ODD EVEN events. 
 
Hope it makes things clearer. Once we all come to a common understanding, I will then proceed with making the changes.

rcassani:
> Now if we create the ODD and EVEN after importing the data, we are creating these events based on the local DC10 event per stimulation block. So this gives the correct ordering.

@chinmaychinara91 on processing the raw file, you have complete control on how create the ODD and EVEN events. If it is needed to restart the ODD-EVEN sequence for each stimulation block:

```
LOAD the `sEvents` structure
CREATE Groups of instances of DC10 using the stimulation block
CREATE Events ODD and EVEN without instances
FOR each Group
   CREATE events ODD and EVEN according to the instance of DC10 in the block
   APPEND events ODD and EVEN   
```

> We then create two copies of the events DC10, one called ODD that will contain the odd instances of DC10 event (delete the even instances in the ODD group) and the other called EVEN that will contain the even instances of the DC10 event (delete the odd instances in the EVEN group). Two approaches here:

@chinmaychinara91, This was my suggestion as I did not know the ODD-EVEN was per stimulation block. You had all the information and skills to do it properly.

chinmaychinara91:
@rcassani thank you for the clarification. I will go ahead and make the changes accordingly.

chinmaychinara91:
Closing this PR and moving things to https://github.com/brainstorm-tools/brainstorm3/pull/913 for a clean history.

rcassani:
Ok, we continue on #913

Just a small note, this PR would have merged with squash, thus the commit history gets collapsed anyway, so to have a "clean history" isn't really a concern. Keeping the same feature in one PR is valuable as it keeps the full conversation, decisions, and code evolution in one place. Moving to new PRs makes that harder to trace later.

chinmaychinara91:
@rcassani Thank you for the clarification. That makes sense. I was thinking of keeping the final changes easier to review, but I understand your point, and that keeping the discussion/code evolution in one PR is better for traceability. I’ll keep this in mind for future PRs.

rcassani:
This PR is reopened, following the in-person discussion
FYI @tmedani

chinmaychinara91:
@rcassani, I have updated the description with details.

toolbox/process/functions/process_customize_spes_nk.m
```diff
@@ -0,0 +1,229 @@
...
+eval(macro_method);
+end
+
+%% ===== GET DESCRIPTION =====
+function sProcess = GetDescription() %#ok<DEFNU>
+% Description the process
+sProcess.Comment     = 'Customize SPES (Nihon Kohden)';
+sProcess.Category    = 'Custom';
+sProcess.SubGroup    = 'Stimulation';
+sProcess.Index       = 901;
```
rcassani: This process does nothing about simulation, it renames and create events.
SubGroup changed to `'Events'`

toolbox/process/functions/process_customize_spes_nk.m
```diff
@@ -131,11 +133,63 @@
...
+            sEvents = DataMat.Events;
+        end
+        if isempty(sEvents)
+            bst_report('Error', sProcess, sInput, 'This file does not contain any event.');
+            return
+        end
         % Locate Nihon Kohden stimulation block events
-        EventMat = in_bst_data(sInputs(iFile).FileName, 'F');
-        iStart = find(strncmp({EventMat.F.events.label}, 'Stim Start', 10));
-        iStop  = find(strncmp({EventMat.F.events.label}, 'Stim Stop', 9));
```
rcassani: Expecting that the events are in the good order is not a good practice.
Corrected in 3dc262c95961cf2bb4e679ee66f1056017e679b7

toolbox/process/functions/process_customize_spes_nk.m
```diff
@@ -0,0 +1,308 @@
...
+%          Chinmay Chinara, 2026
+%          Raymundo Cassani, 2026
+
+eval(macro_method);
+end
+
+%% ===== GET DESCRIPTION =====
+function sProcess = GetDescription() %#ok<DEFNU>
+    % Description the process
+    sProcess.Comment     = 'Customize SPES (Nihon Kohden)';
```
rcassani: @chinmaychinara91, verify that this is the name we want in the GUI

toolbox/process/functions/process_evt_detect_spes.m
```diff
@@ -0,0 +1,308 @@
...
+        end
+
+        % Save events
+        if isRaw
+            DataMat.F.events = sEvents;
+        else
+            DataMat.Events = sEvents;
+        end
+        % Add history entry
+        DataMat = bst_history('add', DataMat, 'NK', ['DESCRIPTION']);
```
rcassani: @chinmaychinara91, update the information that is saved in History to indicate what was done with the Events in the file

toolbox/process/functions/process_evt_detect_spes.m
rcassani, line 1: @chinmaychinara91, since it is a events process, follow the naming: `process_evt_...`

toolbox/gui/panel_record.m
```diff
@@ -171,6 +171,7 @@
         gui_component('MenuItem', jMenu, [], 'Group by time', IconLoader.ICON_FUSION, [], @(h,ev)CallProcessOnRaw('process_evt_grouptime'));
         gui_component('MenuItem', jMenu, [], 'Merge overlapping extended events', IconLoader.ICON_FUSION, [], @(h,ev)CallProcessOnRaw('process_evt_mergeext'));
         gui_component('MenuItem', jMenu, [], 'Add time offset', IconLoader.ICON_ARROW_RIGHT, [], @(h,ev)CallProcessOnRaw('process_evt_timeoffset'));
+        gui_component('MenuItem', jMenu, [], 'Customize SPES (NK)', IconLoader.ICON_EVT_TYPE_ADD, [], @(h,ev)CallProcessOnRaw('process_customize_spes_nk'));
```
rcassani: @chinmaychinara91, same as with the process menu. Verify that is the name we want in the GUI

rcassani (changes requested):
* The entire code was refactored to avoid unnecessary R/W the data file (+200 times)

* The process now works for imported and raw data. Execution in raw data is preferred.

* All processing is now done in the loaded event structure.

* Process can now be called with the raw data open with the option:
 `Record tab > Events > Customize SPES (NK)`
---

Before:

Image

After:

Image

toolbox/gui/panel_record.m
rcassani: In a second thought, not sure if we need it here.
The process is very specific to these files and protocol, it may not be of use to most users
@tmedani, what do you think?

toolbox/process/functions/process_customize_spes_nk.m
chinmaychinara91: Thanks for this. This is much better.

toolbox/process/functions/process_customize_spes_nk.m
chinmaychinara91: how about `Edit SPES events (Nihon Kohden)` ?

toolbox/process/functions/process_evt_detect_spes.m
chinmaychinara91: how about `process_evt_edit_spes_nk` ? Should I go ahead and make the change ?

toolbox/gui/panel_record.m
chinmaychinara91: This is a good addition I feel. It would still be nice to have it here, even though it is specific to a certain use case.

toolbox/process/functions/process_customize_spes_nk.m
rcassani: @chinmaychinara91, as per the devs discussion, these FAST graph process will have their own Process group: `FAST graphs`. See [56745b4](https://github.com/brainstorm-tools/brainstorm3/pull/911/commits/56745b4b21363cfb9452807279d0f5265dee07d8)

toolbox/gui/panel_record.m
rcassani: As per the devs discussion, this is now removed. Events menun in record tab should have only operations that are general rather than specific. See [9cbfdae](https://github.com/brainstorm-tools/brainstorm3/pull/911/commits/9cbfdae758bbea480697083521586ceeefb21c13)

toolbox/process/functions/process_evt_detect_spes.m
rcassani: Renamed to `process_evt_detect_spes`, [e0d8bbd](https://github.com/brainstorm-tools/brainstorm3/pull/911/commits/e0d8bbd7ed0e8e3cc9076f5a97844cc8b881c261)

toolbox/process/functions/process_customize_spes_nk.m
rcassani: Updated to `Detect single-pulse triggers in SPES` [d451308](https://github.com/brainstorm-tools/brainstorm3/pull/911/commits/d4513088d2b1151f036b2b38f2369b997274ef58)

toolbox/process/functions/process_evt_detect_spes.m
rcassani: Not in history, but in the report [570f6bb](https://github.com/brainstorm-tools/brainstorm3/pull/911/commits/570f6bb311906083c9c802e95b6d6555ee4eeb3d)

rcassani:
@chinmaychinara91, it looks good for me to be merged, check on your side

chinmaychinara91:
Looks good to me. It can be merged.

PR #912, opened 2026-04-21
[2/3] Fastgraph: Remove SPES artifacts using EMD

chinmaychinara91:
This PR is the 2nd part required for [Fastgraph](https://pubmed.ncbi.nlm.nih.gov/32086098/). It involves cleaning the data for any artifacts. 

**Prerequisite:** Follow steps in PR  https://github.com/brainstorm-tools/brainstorm3/pull/911

# Remove SPES artifacts
Linked to `process_remove_spes_artifacts`  and it involves removing SPES artifacts and slow drifts using the Empirical Mode Decomposition (EMD) based filtering.

## The GUI
image

## Inputs
- **<ins>Stimulation event</ins>**: The stimulation event around which to remove the artifacts (here, `STIM`)
- **<ins>EMD cutoff frequency</ins>** (in Hz): Self explanatory
- **<ins>Time to account for artifact</ins>** (in ms): The duration of the actual stimulation artifact itself, starting at the stimulation event. Used to set the main artifact span inside the interpolation window (here, `5` ms).
- **<ins>Time on either side of the artifact to spline with</ins>** (in ms): How much clean signal to take from before and after that artifact window so the spline has anchor points to interpolate across the bad segment (here, `3` ms).

## Steps to use
1. Follow steps as in PR #911 .
2. Drag the loaded stimulation blocks into the `Process1` tab. For e.g. `SB A13-A14 4 (#1)` .
3. Click on `Run > Stimulation > Remove SPES artifacts`. 
4. Fill in the values as seen in the GUI figure above. Click `Run`.

It creates new stimulation block e.g. `SB A13-A14 4 (#1)_emd`. The results can be seen as under.
Screenshot 2026-04-21 140035

## Associated PRs
1. https://github.com/brainstorm-tools/brainstorm3/pull/911
2. #916 

@jcmosher @Nastaranlotfi @yashvakilna


toolbox/process/functions/process_remove_spes_artifacts.m
```diff
@@ -0,0 +1,178 @@
...
+eval(macro_method);
+end
+
+%% ===== GET DESCRIPTION =====
+function sProcess = GetDescription() %#ok<DEFNU>
+% Description the process
+sProcess.Comment     = 'Remove SPES artifacts';
+sProcess.Category    = 'Custom';
+sProcess.SubGroup    = 'Stimulation';
+sProcess.Index       = 902;
```
rcassani: Use same group as in #911, and update index accordingly

toolbox/process/functions/process_remove_spes_artifacts.m
```diff
@@ -0,0 +1,178 @@
...
+%% ===== RUN =====
+function OutputFiles = Run(sProcess, sInputs) %#ok<DEFNU>
+    % Initialize output
+    OutputFiles = {};
+
+    % Get process option values
+    StimEvent     = sProcess.options.stimevent.Value;
+    CutoffFreq    = sProcess.options.cutoff.Value{1};
+    TimeArt       = sProcess.options.timeart.Value{1};
+    TimeSpline    = sProcess.options.timespline.Value{1};
```
rcassani: Add checks for valid input values, e.g., frequency cannot be negative

toolbox/process/functions/process_remove_spes_artifacts.m
```diff
@@ -0,0 +1,178 @@
...
+        eventLabels = {DataMat.Events.label};
+        iStimEvent = find(strncmp(eventLabels, StimEvent, length(StimEvent)));
+        if isempty(iStimEvent)
+            bst_report('Error', sProcess, [], ['No ' StimEvent ' event found']);
+            return;
+        end        
+
+        % Convert trigger times to sample indices
+        stimTimes = DataMat.Events(iStimEvent).times;
+        [isMatch, stimSamples] = ismembertol(stimTimes, DataMat.Time, 1e-7);
```
rcassani: You can use `bst_closest`

toolbox/process/functions/process_remove_spes_artifacts.m
```diff
@@ -0,0 +1,178 @@
...
+        for iChan = 1:size(DataMat.F, 1) % for each channel of data
+            for iStimWin = 1:size(stimWin, 2) % for each artifact window
+                DataMat.F(iChan, stimWin(:, iStimWin)) = spline(win(iSpline), DataMat.F(iChan, stimWin(iSpline, iStimWin)), win);
+            end
+        end
+        
+        % Apply EMD based filtering to suppress drift
+        bst_progress('start', 'Process', sprintf('[%d/%d] Processing EMD, rejecting below %.1f Hz...', iFile, length(sInputs), CutoffFreq), 0, 100);
+        % Sampling frequency
+        sampFreq = 1 / mean(diff(DataMat.Time));
```
rcassani: This is already computed as `Fs`

toolbox/process/functions/process_remove_spes_artifacts.m
```diff
@@ -0,0 +1,178 @@
...
+        end
+       
+        % Duplicate original file before writing cleaned data
+        sFile = bst_process('CallProcess', 'process_duplicate', sInputs(iFile).FileName, [], ...
+            'target', 1, ...            % Duplicate data files
+            'tag',    sprintf('_emd'));
+        % Overwrite duplicated file with cleaned signal
+        dataMatNew   = in_bst_data(sFile.FileName, 'F');
+        dataMatNew.F = DataMat.F;
+        bst_save(file_fullpath(sFile.FileName), dataMatNew, 'v7', 1);
```
rcassani: Do not duplicate the file.

1. Create a new file, and set a filename that makes sense
2. Write the in the history of the new file
3. Add that new file to the database 

In fact, this process could be way cleaner using the Category `File`
See `process_extract_time`

rcassani (changes requested):
This PR presents the process `process_remove_spes_artifacts` to remove artifacts from stimulation, and to remove drift. 

There is reason for artifact-removal operations to be in the same process. It would be more versatile to have them in two different process, so Brainstorm users can use them separately.

toolbox/process/functions/process_remove_spes_artifacts.m
chinmaychinara91: Thank you for the suggestion. Slight confusion here.

>Create a new file, and set a filename that makes sense
Write the in the history of the new file
Add that new file to the database

I checked and this is how it is done for the category `File`. e.g. `process_avg_freq`

> In fact, this process could be way cleaner using the Category File
See process_extract_time

Did you mean category `Filter` here, as it is in `process_extract_time`? 

We can go with either categories I believe, `File` or `Filter`. What would you recommend ?

toolbox/process/functions/process_remove_spes_artifacts.m
chinmaychinara91: Fixed [86548e6](https://github.com/brainstorm-tools/brainstorm3/pull/912/commits/86548e6c3e94b19718ee0a20b466c55b7237e4f4)

toolbox/process/functions/process_remove_spes_artifacts.m
chinmaychinara91: Done [0f12ef0](https://github.com/brainstorm-tools/brainstorm3/pull/912/commits/0f12ef0ea4732ee576a2d5cf1b5ecfae64744c6b)

toolbox/process/functions/process_remove_spes_artifacts.m
chinmaychinara91: Done [25e6f9b](https://github.com/brainstorm-tools/brainstorm3/pull/912/commits/25e6f9bddedb85fcf7ba5160f5e9c1441237d74a)

chinmaychinara91:
> There is reason for artifact-removal operations to be in the same process. 

The reason we do stimulation artifact removal before drift correction is to avoid the large transient artifacts that may bias the drift estimate. Hence, we decided to club them in one process here.

> It would be more versatile to have them in two different processes, so Brainstorm users can use them separately.

I agree that the two operations are conceptually independent, and separating them into two independent processes would definitely make the workflow more flexible for users. Thank you for the recommendation.

@jcmosher, any more inputs on this ?

Should I go ahead and create two separate processes? While documenting in the tutorial, we can suggest the recommended order for the users when both steps are applied.

toolbox/process/functions/process_remove_spes_artifacts.m
chinmaychinara91: Check [13317aa](https://github.com/brainstorm-tools/brainstorm3/pull/912/commits/13317aae10d97787610d00ed709d9468a81d2778)

toolbox/process/functions/process_remove_spes_artifacts.m
rcassani: Sorry, yes I meant `Filter`, the process only modifies the data `F` field.

You can read about the details for each category in here:
https://neuroimage.usc.edu/brainstorm/Tutorials/TutUserProcess#Categories_of_process

rcassani:
> Should I go ahead and create two separate processes? While documenting in the tutorial, we can suggest the recommended order for the users when both steps are applied.

Drift removal with EMD can be applied to any other data, that is the reason to have it separated.
Yes, create two processes.

In the tutorial explain what each process do and that will lead the reasoning about the order in which the processes must be used for that tutorial, not a suggestion. As swapping the order is incorrect way to do it for the data of that tutorial.

PR #915, opened 2026-04-28
tess_remesh: add 40962 vertex opption

Edouard2laire:
This is to be more aligned with the list proposed in tess_sphere.  The option to use 40962 vertex in tess_sphere was added here:  https://github.com/brainstorm-tools/brainstorm3/commit/bcbfa71678d90387520ca9665fbf2f692d01f1ab?w=1

Is there a way to increase the max value there? One of my colleagues wanted a 1mm isotropic resolution mesh of the head surface for her project.  but with 40962 it's only 1.5 mm

### DIFF
--- a/toolbox/anatomy/tess_remesh.m
+++ b/toolbox/anatomy/tess_remesh.m
@@ -8,7 +8,7 @@ function [sph_vert, sph_faces] = tess_remesh(vert, nvert, isCenter)
 %    - SurfaceFile : Absolute or relative path to the surface file to remesh
 %    - vert        : [nVertices x 3] matrix of xyz vertices
 %    - nvert       : Number of vertices in the output mesh, possible values
-%                    [12 32 42 92 122 162 273 362 482 642 812 1082 1442 1922 2432 2562 3242 4322 5762 7682 7292 9722 10242 12962]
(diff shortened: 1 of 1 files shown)


rcassani:
Instead of remeshing, the head surface can be recomputed.
Right-click on the MRI, then `MRI segmentation > Generate head surface`, and set desired number of vertices (even higher than 40962).

Edouard2laire:
> Instead of remeshing, the head surface can be recomputed. Right-click on the MRI, then `MRI segmentation > Generate head surface`, and set desired number of vertices (even higher than 40962).

The issue with that is that the mesh is not uniform. Remeshing is mandatory to obtains a nice regular mesh

rcassani:
Try adding the option `122882` `tess_remesh`, and the case for `122882` in `tess_sphere`:

```Matlab
case 122882, n1 = 1;  n2 = 6;
```

This should do the trick, I get a mean_edge_length of 1.2mm in the head surface of the Default anatomy.

If it works well, let me know to add the option on the `master`

rcassani:
Ping @Edouard2laire

Edouard2laire:
tested. it works. Are you going to push directly or should i open a PR ?

rcassani:
I'll push it directly, I just needed your feedback
Done: in a99505d8e

Edouard2laire:
thanks ! :)

PR #917, opened 2026-05-01
[docker] Improve image identification (?)

Edouard2laire:
it seems that the sha returned during pull doesnt correspond to the ID of the docker image that needs to be used later.

### DIFF
--- a/toolbox/core/bst_containers.m
+++ b/toolbox/core/bst_containers.m
@@ -185,6 +185,19 @@ function [errMsg, imageSha] = ImportImage(imageSource, imageTag)
         case 'docker'
             switch imageType
                 case 'reference'
+                    [status, cmdout] = system([' docker manifest inspect ' imageSource]);
+                    if status == 0
+                        image_info = jsondecode(cmdout);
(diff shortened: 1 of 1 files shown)


rcassani:
Thank you for testing the recently added container support.

This feature has been just released and is still undergoing testing on our end. We're not yet certain this fix is the only thing that needs addressing. We'd like to do a more detailed review before accepting patches for it.

For now this PR will be closed, and we will check it back once we've had a chance to fully validate the feature in Apple Silicon.

Edouard2laire:
you could still use this commit: https://github.com/brainstorm-tools/brainstorm3/pull/917/changes/063a1dabbad66510d35ff8c3193ded207ce516aa


Informing user about the size of the docker image they are about to pull still gives good information (espetially since there is no waiting bar during the download)

rcassani:
> you could still use this commit: https://github.com/brainstorm-tools/brainstorm3/commit/063a1dabbad66510d35ff8c3193ded207ce516aa

Yes, a similar message will be included

PR #919, opened 2026-05-09
SEEG: Add infinite homogeneous medium headmodel

Edouard2laire:
Hello, 

This PR adds a very simple head model for SEEG that assumes an infinite medium with uniform conductivity (sigma). 

This is described in https://onlinelibrary.wiley.com/doi/epdf/10.1002/hbm.23127 (equation 4) with a tweak to limit instability close to the electrodes (eq 5).

Although I think we will slowly move toward more complex FEM models, having this simple model in Brainstorm will enable easier comparison of the different models. 

Todo: 

- [x]  Add option for the model: min_distance and conductivity
- [x] Test on volumetric head model


rcassani:
You may want to talk with Takfarinas as he is working in related (and potentially) overlapping) developments

FYI @tmedani

Edouard2laire:
Tested on volume. Works fine as far i know. 

image

@rcassani, I think it's ready for review, most importantly for the user-interface part. Regarding the name, uniform may not be the best. I was thinking  "0-shell" model might work well as an 'extension' to the 1-shell EEG model. (Here, we don't model the skull. What do you think? 


@tmedani, what do you think about this? 


Edouard

tmedani:
Hi @Edouard2laire, @rcassani

This is indeed a good feature for a simplified version of the sEEG LF, similar to the sphere model for EEG/MEG. While we would recommend using BEM/FEM in such cases, this feature can be used for comparison and validation. I think it is nice to have it within BST.

tmedani:
Hi team, 
It is on my list; I'll review it asap.

tmedani:
Hi @Edouard2laire 
The code works fine, needs some cleanup before merging, and probably @rcassani  will make some modifications. 

Have you had a chance to compare the LF of this method with BEM/FEM?

I just did one quickly, FEM vs Inf
image

Looks okay, I'll need to run it in a more refined model + source space.

Also, I haven't checked the paper, but does the formulation assume infinite refElec?

Edouard2laire:
Hi,


> Have you had a chance to compare the LF of this method with BEM/FEM?

No, I didn't. Thanks a lot for having done the comparison. 


> Also, I haven't checked the paper, but does the formulation assume infinite refElec?


In the paper, they used an averaged reference. But based on my understanding of how forward models are stored in Brainstorm, I assumed infinite refElec in the code, as the model can then be re-referenced before being used. If we have an explicit reference as suggested in https://github.com/brainstorm-tools/brainstorm3/pull/893 then it should be relatively easy to add.

rcassani:
@Edouard2laire, @tmedani, minor changes were made:
- Move the interactive parts to `panel_heademodel`
- Improve doc string
- Use functions compatible with older versions of Matlab

Functionally, it looks good to be merged.

Regarding the name, it seems `uniform medium` is quite vague, while `Infinite homogeneous medium` is quite long. 

Would you consider the label `Homogenous medium` as a good trade off to be present in the GUI?

Edouard2laire:
> Would you consider the label Homogenous medium as a good trade off to be present in the GUI?

Yes, that seems reasonable to me. 


What is your take on the reference system? Is it correct to use an infinite reference here, or should we use an average reference or something else?

rcassani:
> Is it correct to use an infinite reference here, or should we use an average reference or something else?

AFAIK we can keep the resulting `G` with the infinite reference. When this `G` is used to compute the inverse model, then the average reference is applied. This is the approach taken for EEG head models

https://github.com/brainstorm-tools/brainstorm3/blob/8f32f9e2cde53f6b0ab42a11bb626c4a2fc010ab/toolbox/forward/bst_headmodeler.m#L683-L687

Edouard2laire:
Tested. it seems to be working fine

toolbox/forward/bst_seeg_homogeneous.m
```diff
@@ -0,0 +1,113 @@
...
+
+        % Compute unit vectors from SEEG contact to source points (u_j)
+        VectorDipolesToSEEG = SEEG_Loc(iContact, :) - GridLoc;
+        DistanceToDipoles = sqrt(sum(VectorDipolesToSEEG.^2,2));
+        VectorDipolesToSEEG = VectorDipolesToSEEG ./ repmat(DistanceToDipoles, 1, 3);
+
+        % Filter short distance
+        iShort = find(DistanceToDipoles < min_distance);
+        if ~isempty(iShort)
+            fprintf(' %d vertex had distance to the cortex smaller than %.2f mm to electrodes %s \n', length(iShort), min_distance*1000, sChannel(iContact).Name);
```
Edouard2laire: Maybe we remove this printf as its not mandatory for the user

rcassani: Agree, addressed in [489b56d](https://github.com/brainstorm-tools/brainstorm3/pull/919/commits/489b56d7f7a27d524786e289e4eec968cf3d0b58)

PR #899, opened 2026-03-12
Add support container-based plugins

rcassani:
This PR has as goal to provide an API to manage containers from Brainstorm, and integrate it in GUI

## General settings
- [x] Select preferred container engine in options
- [x] Save (set) and read (get) preferred container

## Containers API
- Supported engines
  - [x] [Docker Desktop](https://www.docker.com/products/docker-desktop/)
  - [ ] [Podman](https://podman.io/) Left for future PR
  - [ ] [Apptainer](https://apptainer.org/) Left for future PR

## Containers as plugins
- [x] Update plugin descriptor to allow container plugins:
  Add fields for image reference and image SHA
- [x] Update plugin menus

%% Container API
- [x] Pull image (Plugin Install)
- [x] Run container as daemon on (Plugin Load)
- [x] Stop daemon container on (Plugin Unload)
- [x] Remove image (Plugin Uninstall)

## Documentation
- [x] Containers page in Brainstorm website
  https://neuroimage.usc.edu/brainstorm/Tutorials/Containers
- [x] Add information on Plugin tutorial
  https://neuroimage.usc.edu/brainstorm/Tutorials/Plugins#Container_plugins

## Additional info
- Toy example container-based plugin: https://github.com/rcassani/bst-plugin-container
- Related PR: #882

PR #920, opened 2026-05-18
Run resection identification using Docker container

chinmaychinara91:
This PR introduces the execution of the `resection-identification` plugin by transitioning it to run within a Docker container, utilizing the recent container-based plugin support in Brainstorm (https://github.com/brainstorm-tools/brainstorm3/pull/899).

Related forum discussion: https://neuroimage.usc.edu/forums/t/automatic-resection-identification/56589/16

# Usage
- Open Docker Desktop and keep it running in the background. Refer to the **Notes** section in the forum post above for some additional recommended setup.
- Install the `Anatomy > resection-identification` plugin in Brainstorm to load the image and start the container.
- Follow steps as mentioned in the tutorial (https://neuroimage.usc.edu/brainstorm/Tutorials/ResectionIdentification)

# Checklist
- [x] Test on Windows 11 (w/ and w/o NVIDIA GPU)
- [x] Test on Mac (Apple Silicon) - needs more testing (see [discussion](https://github.com/brainstorm-tools/brainstorm3/pull/920#discussion_r3266765639) - kept for future PR)
- [x] Test on Linux (w/ and w/o NVIDIA GPU)
- [x] Update tutorial page with info on the containerization

FYI: @tmedani @ajoshiusc @Edouard2laire


toolbox/core/bst_containers.m
```diff
@@ -383,6 +408,32 @@
...
+                awkPid = '\$1'; % macOS/Linux host: prevent host shell expansion
+            end
+            cmd = ['docker exec ' containerName ' sh -c "ps -eo pid= | awk ''' awkPid ' != 1 {print ' awkPid '}'' | xargs -r kill -9 2>/dev/null || true"'];
+            [status, cmdout] = system(cmd);            
+            if status ~= 0
+                errMsg = strtrim(cmdout);
+            end
+    end
+    bst_progress('stop');
+end
```
chinmaychinara91: This I feel is a good function to have for reasons mentioned in the commit https://github.com/brainstorm-tools/brainstorm3/pull/920/commits/c74c7eab7a388ad1a6593343dcb1486a5f82adda. Another way would be to maybe stop and restart the container but I feel this proposed function is a better way of doing it. Thoughts ?

chinmaychinara91 (commented):
@rcassani I have tested these for Windows and Mac (Apple Silicon) and it works fine. Would need you help on testing it on Linux.

toolbox/core/bst_containers.m
```diff
@@ -70,6 +70,12 @@
                         end
                     end
                 else
+                    if ismac
+                        setenv('PATH', ['/usr/local/bin:' ...
+                               bst_fullfile(getenv('HOME'), '.docker/bin') ':' ...
+                               '/Applications/Docker.app/Contents/Resources/bin:' ...
+                               getenv('PATH')]);
+                    end
```
rcassani: Modifying the PATH, even inside MATLAB, does not seem a good idea
1. We do not do it for other OS
2. It can override users executable 
3. It can have unexpected results later in Brainstorm when looking for other executable

Other approach could be to:
1. Search the existence of the executable docker (and/or other engines) in those paths and keep the full path for the system calls 
2. To require the user properly set the PATH so docker is visible
3. A pop-up to the user about setting docker?

I think we can go with 1, though it will require to update the containers API a bit

FYI @Edouard2laire

toolbox/core/bst_containers.m
rcassani: Can the keepalive approach to leave temporary files and configurations?

The completely stop and clean of the container seems better suited for here.
So if the user interrurts Ctrl-C, the container stops, an error message appears saying that it was interrupted by the user, and the way to start again is to run the process. A new container will be run, thus each run will be consistent

toolbox/core/bst_plugin.m
```diff
@@ -209,15 +209,9 @@
...
     PlugDesc(end).Category       = 'Anatomy';
-    PlugDesc(end).AutoUpdate     = 1;
-    PlugDesc(end).URLzip         = ['https://neuroimage.usc.edu/bst/getupdate.php?d=bst_resection_identification_' OsType '.zip'];
-    PlugDesc(end).TestFile       = 'resection_identification';
-    if strcmp(OsType, 'win64')
-        PlugDesc(end).TestFile   = [PlugDesc(end).TestFile, '.bat'];
-    end
-    PlugDesc(end).URLinfo        = 'https://github.com/ajoshiusc/auto_resection_mask/tree/brainstorm-plugin';
+    PlugDesc(end).URLinfo        = 'https://github.com/ajoshiusc/auto_resection_mask/tree/brainstorm-container';
+    PlugDesc(end).ImageSource    = 'docker.io/chinmaychinara/auto-resection-mask:latest';
```
rcassani: Before merging, let's move it to the Brainstorm account

rcassani:
@chinmaychinara91, did you notice any strange behaviour on using the `bst_containers` API to handle the images and containers in Apple Silicon. 

Specifically, in some tests @Edouard2laire reported the image SHA returned by `ImportImage()` did not match the image SHA obtained with `GetImages()` so Brainstorm was not able to check if the image was properly pulled. Check this PR: https://github.com/brainstorm-tools/brainstorm3/pull/917

FYI @Edouard2laire

toolbox/core/bst_containers.m
chinmaychinara91: > Can the keepalive approach leave temporary files and configurations?

That is a good point. My initial thought was to keep the container alive so that the user does not need to pay the container startup cost again, while still killing any running child processes after an interrupt/error. However, this step does not guarantee that it leaves out temporary files and configurations. That could make the next run inconsistent. Not that I have observed anything as such but it could be an issue.

> The completely stop and clean of the container seems better suited for here.
So if the user interrurts Ctrl-C, the container stops, an error message appears saying that it was interrupted by the user, and the way to start again is to run the process. A new container will be run, thus each run will be consistent

Agreed lets move with this.

toolbox/core/bst_containers.m
chinmaychinara91: I see. Let's move ahead with your suggestion 1.

toolbox/core/bst_plugin.m
chinmaychinara91: You mean moving the Docker image to Brainstorm Docker account ?

chinmaychinara91:
> @chinmaychinara91, did you notice any strange behaviour on using the `bst_containers` API to handle the images and containers in Apple Silicon.
> 
> Specifically, in some tests @Edouard2laire reported the image SHA returned by `ImportImage()` did not match the image SHA obtained with `GetImages()` so Brainstorm was not able to check if the image was properly pulled. Check this PR: #917

It works fine as it is now on Apple Silicon. On pulling, the correct image SHA is obtained. Nothing strange was observed there.

chinmaychinara91:
@Edouard2laire was it a specific scenario where you could not get the correct image SHA ?

Edouard2laire:
> @Edouard2laire was it a specific scenario where you could not get the correct image SHA ?

I just tried with the example image, 

and got this: 

image

with the following plugin:

```
{
  "Name": "cont_plug",
  "Version": "latest",
  "URLzip": "",
  "URLinfo": "https://github.com/rcassani/bst-plugin-container",
  "ImageSource": "docker.io/library/alpine:latest",
  "CompiledStatus": 1,
  "Category": "User defined"
}
```

But the image is downloaded, just with a different sha: 
```
[errMsg, imageList]     = bst_containers('GetImages')
{'alpine:latest'}    {'sha256:2ffb2ff4aab36d06…'}
```

docker pull returns this: 

```
cmdout =

    'latest: Pulling from library/alpine
     Digest: sha256:5b10f432ef3da1b8d4c7eb6c487f2f5a8f096bc91145e68878dd4a5019afde11
     Status: Image is up to date for alpine:latest
     docker.io/library/alpine:latest
     '
```



image


I am using Matlab 2023b; Here are the info of docker:

image

chinmaychinara91:
These are from the Apple Silicon machine we have for the same `cont_plug.json` file. Image SHA looks fine across.

On MATLAB 2024b:
Screenshot 2026-05-19 at 4 12 28 PM


Docker desktop details:
Screenshot 2026-05-19 at 4 11 28 PM
Screenshot 2026-05-19 at 4 09 06 PM

Edouard2laire:
It seems, on my side that the digest is different form the imageID: 

```
[edelaire1@home-imglab01:~]$ docker pull docker.io/library/alpine:latest
latest: Pulling from library/alpine
Digest: sha256:5b10f432ef3da1b8d4c7eb6c487f2f5a8f096bc91145e68878dd4a5019afde11
Status: Image is up to date for alpine:latest
docker.io/library/alpine:latest
[edelaire1@home-imglab01:~]$ docker image ls
REPOSITORY   TAG       IMAGE ID       CREATED       SIZE
alpine       latest    2ffb2ff4aab3   4 weeks ago   8.7MB
[edelaire1@home-imglab01:~]$ docker inspect alpine:latest --format='{{index .RepoDigests 0}}'
alpine@sha256:5b10f432ef3da1b8d4c7eb6c487f2f5a8f096bc91145e68878dd4a5019afde11
```


It seems to be discussed here: https://stackoverflow.com/questions/56364643/whats-the-difference-between-a-docker-images-image-id-and-its-digest

With docker image ls --digests , we can get the link between the two ids: 

```
[edelaire1@home-imglab01:~]$ docker image ls --digests
REPOSITORY   TAG       DIGEST                                                                    IMAGE ID       CREATED       SIZE
alpine       latest    sha256:5b10f432ef3da1b8d4c7eb6c487f2f5a8f096bc91145e68878dd4a5019afde11   2ffb2ff4aab3   4 weeks ago   8.7MB

```

chinmaychinara91:
@rcassani just wanted to gently follow up on the next steps for this PR. Kindly let me know how you would like to proceed.

toolbox/core/bst_containers.m
rcassani: Let's keep this PR with changes only for the resection id.
Remove the search for `docker` in macOS. Let's assumed for now it is already in the Path

toolbox/core/bst_containers.m
rcassani: Ok, this need to be updated

toolbox/core/bst_plugin.m
rcassani: Yes, a Brainstorm Docker account

rcassani:
@chinmaychinara91, 

1. Please resolve the three requested changes:

https://github.com/brainstorm-tools/brainstorm3/pull/920/changes/BASE..a43ae27c42c19f3d583fb54e938c50b1241e9542

2. I'll test on Linux with and without GPU, then it should be good to be merged

toolbox/core/bst_plugin.m
chinmaychinara91: Check [b300560](https://github.com/brainstorm-tools/brainstorm3/pull/920/commits/b3005604a948bc869387e7df11917df7ab433aff)

toolbox/core/bst_containers.m
chinmaychinara91: Check https://github.com/brainstorm-tools/brainstorm3/pull/920/commits/6286877afbda67c2ba7f704c8f7b5f0b55f7fd87

toolbox/core/bst_containers.m
chinmaychinara91: Check https://github.com/brainstorm-tools/brainstorm3/pull/920/commits/45d9bbf6a0cd61c90d5fd3b57042f9e5a7fa92c2

chinmaychinara91:
@rcassani I resolved the requested changes.

One thing I noticed: this error message appears in the command window:
`BST> Error: Could not get online version for plugin: resection-identification`

It looks like `bst_plugin` may need a few additional checks for container-based plugins, for example handling the call to `GetVersionOnline` when the plugin is a container. Since this is not directly in scope for this PR, maybe you could add those changes directly to `master`.

> I'll test on Linux with and without GPU, then it should be good to be merged

Sounds good. I will update the tutorial as well once you give a go ahead.

toolbox/process/functions/process_resection_identification.m
```diff
@@ -130,39 +130,42 @@
...
-    strCall = ['"' ResecExe '"' ' ' '"' preOpNii '"' ' ' '"' postOpNii '"' ' ' '"' TmpDir '"'];
-    disp(['RESEC_ID > System call: ' strCall]);
-    tic;
-    status = system(strCall);
-    if (status ~= 0)
-        errMsg = 'Error during resection-identification, see logs in the command window.';
-        bst_progress('stop');
-        return;
+    % Make dir for processed files
+    mkdir(bst_fullfile(TmpDir, 'temp_dir_resection'));
```
rcassani, line 157: @chinmaychinara91, this is done on the Matlab side, otherwise, if `docker` runs as root (in the case it is the engine alone, not Docker Desktop) Brainstorm will not be able to unzip the resulting anatomies in that dir. When created from Matlab, `docker` respect the current permissions

toolbox/core/bst_containers.m
```diff
@@ -428,6 +446,13 @@
 end
 
 
+%% ===== PROCESS INTERRUPTED =====
+function ProcessInterrupted(containerName)
+    bst_plugin('Unload', strrep(containerName, 'bst_', ''));
```
rcassani: Use regexp to force to replace `bst_` ONLY at the beginning of the string.

(It may not be used, see the commend on the issues with `onCleanup()`)

toolbox/core/bst_containers.m
```diff
@@ -316,7 +328,13 @@
             else
                 commandWrapper = ''''; % Single quote
             end
-            [status, cmdout] = system(['docker exec ' containerName ' sh -c ' commandWrapper cmdStr commandWrapper]);
+            
+            % If MATLAB function errors, or Ctrl+C is pressed
+            cleanupObj = onCleanup(@() ProcessInterrupted(containerName));
```
rcassani: This is not a good idea, the `task` for `cleanupObj` runs either the user presses `Ctrl-C` or the function (`ExecInContainer()`) ends normally. 

In the second case, then the container plugin is Unloaded (removed with its files), before importing back the results.

In newer versions of Matlab there is the option to cancel the task, but that was introduced just in R2025a
https://www.mathworks.com/help/matlab/ref/oncleanup.cancel.html

rcassani (changes requested):
I tested on Linux and things look right, there are some details to address, check the comments for this review.

toolbox/process/functions/process_resection_identification.m
chinmaychinara91: I see sounds good.

toolbox/core/bst_containers.m
chinmaychinara91: Tried a different approach by keeping track of the `exec` process state if it got interrupted or completed normally using a `containers.Map`. Check [53af9cc](https://github.com/brainstorm-tools/brainstorm3/pull/920/commits/53af9ccabbf77042fc897bf4555fbb8e8c057fd1).

The state is set to `false` before `system()` starts and changed to `true` immediately after `system()` returns. This way, on normal completion, including non-zero Docker exit status, does not unload the plugin, while `Ctrl+C` or MATLAB interruption unloads the plugin.

toolbox/core/bst_containers.m
chinmaychinara91: Fixed in [53af9cc](https://github.com/brainstorm-tools/brainstorm3/pull/920/commits/53af9ccabbf77042fc897bf4555fbb8e8c057fd1).

PR #926, opened 2026-07-05
On volume contact sheet, show each dipole only on nearest slice

quraishi:
Previously dipoles were on every slice which was difficult to interpret.

The issue was mentioned here:
https://neuroimage.usc.edu/forums/t/dipole-mri-viewer-contact-sheet-error/47730

Before:
before_axial

After:
after_axial


rcassani:
@quraishi, thank you for your contribution!

I made small changes to place the function in `panel_dipoles` and it as an option to the GUI

Peek 2026-07-07 15-41

PR #930, opened 2026-07-16
Spike detection using spikenet1

Edouard2laire:
Hello, 

This PR add process to call spikenet1 for interictal spike detection from scalp EEG. 
The PR is not ready for merging. 

Todo: 

- [x] Download model weight automatically 
- [x] Thresdhold probabilty and create event automatically 
- [ ] Use GPU when possible
- [ ] Clean code 
- [ ] interface spikenet1 / spikenet2
- [ ] Better inegrate to the interface


tmedani:
Hi @Edouard2laire 

This is great. Thanks for initiating this work and for the PR.

I've heard about SpikeNet, but I've never had a chance to dive into it.

It also seems that there's a SpikeNet2. Is there a significant difference between the two? Have you had a chance to check it out?

Edouard2laire:
hi. 

we wanted to use spikenet2 but were not able to find the weight of the model online.  We tried to contact the author, without success so far :) 


Actually it seems it was uploaded last week: https://github.com/bdsp-core/SpikeNet2/commit/12b1bf52476fd278771e81f48c494cc2ebe6021f

we might give it another try. Maybe i can make the process such that we can choose between the two models

PR #932, opened 2026-07-17
bugfix: correctly identify event that needs channel information

Edouard2laire:
right now, it was adding channel information to event even if the event was present only in one file (MEG for example)


Edouard2laire:
closing. this is the desired behavior

PR #933, opened 2026-07-24
Function signature

Edouard2laire:
Hello, 

This PR doesn't add much to brainstorm but kind of helps when writting script / process (i am always forgetting the exact spelling of the option when calling bst_get or db_template) :  

it allows MATLAB to propose the context name for the function bst_get, when using tab after writing "bst_get" : 
image

Unfortunately, it doesn't seem possible to complete the following arguments nicely depending on the first one; but I guess it's already nice. 

If that's something you think might be useful, we could do the same thing for db_get, db_set, bst_set, db_template.

More information here:  https://www.mathworks.com/help/matlab/matlab_prog/customize-code-suggestions-and-completions.html

Here is what it can look like on "simpler" function: 
image

Edouard


rcassani:
This is a nice-to-have feature

Ideally, we would like to generate the `.json` file automatically on request and for a selected set of files.

Edouard2laire:
So trying to parse the mat file, and extract the call from the comment ?

rcassani:
Parsing the code. Doc strings are not always up to date. That's another place of improvement.

```
VERIFY function has 'eval(macro_method)'
FIND all calls, regexp `'^\s*function.*\s(\w+)\('`
```

1. Does it make sense to add the function signatures in the repo?
  As mentioned, this is more for developers. 

2. The autocompletion is suggested always as string `"TEXT"` rather than char vector `'TEXT'`, right?
  
  In the [scripting tutorial](https://neuroimage.usc.edu/brainstorm/Tutorials/Scripting) we indicate that text data should be given as char vector, as strings and char vectors are not always interchargeable. There have been some reported issues because of that: https://neuroimage.usc.edu/forums/t/org-brainstorm-tree-bstnode-is-not-assignable-to-org-brainstorm-tree-bstnode/33426/5
  
  It would be strange that `bst_get("Subject", 'MrFox')` is valid, but not `bst_get("Subject", "MrFox")`

Edouard2laire:
> Parsing the code. Doc strings are not always up to date. That's another place of improvement.
> 
> ```
> VERIFY function has 'eval(macro_method)'
> FIND all calls, regexp `'^\s*function.*\s(\w+)\('`
> ```

Another nice way is to use localfunctions; so for example: 

```
>> process_absolute('localfunctions')

ans =

  3×1 cell array

    {@GetDescription}
    { @FormatComment}
    {           @Run}

```


>  1. Does it make sense to add the function signatures in the repo?
>    As mentioned, this is more for developers.

Where should we put it ? it seems it needs to be in the same folder as the function


> 2. The autocompletion is suggested always as string `"TEXT"` rather than char vector `'TEXT'`, right?


yeah i wasn't happy about this too. But i just found that if you hit tab, after writing the ', then it completes as char, as expected. So for bst_get, you have to write bst_get(' , then hit tabs, and it completes correctly

PR #934, opened 2026-08-06
[Project sources] Allow to specify target study

Edouard2laire:
Hello, 

This PR is here to allow specifying the target study when using bst_project_source from a script. 

Here is an example script we did today that could benefit from that. A map is imported in a group subject, then projected to another subject using bst_project_source. The modification allows nicely specifying the condition the map appears to.

```matlab
left_map    = 'L.func.gii';
right_map   = 'R.func.gii';

target_subject      = 'PXXXX';
target_condition    = 'hubness-zscore';

DisplayUnits = 'zscore';


% 0. Get Map name
[~, name, ext]  = fileparts(left_map);
name            = strrep(name, 'L.func', 'LR.func');

% Step 1. Import both texture to group
sSubject = bst_get('Subject', 'Group');
FSLR_5k = sSubject.Surface(strcmp({sSubject.Surface.Comment}, 'fsLR-5k.LR.surf'));

iStudies = db_add_condition('Group', sprintf('sub-%s',target_subject));
[OutputFile, errorMsg] = import_sources(iStudies, FSLR_5k.FileName, left_map, right_map, 'GII', name, DisplayUnits);

% Step 2. Project to subject
sSubject = bst_get('Subject', target_subject);
sCortex = sSubject.Surface(strcmp({sSubject.Surface.Comment}, 'mid_8002V'));


iStudies = db_add_condition(target_subject, target_condition);
sMaps = bst_project_sources( {OutputFile}, sCortex.FileName, 0, 0, iStudies);

```

  
  
   
---------------

Hs: it might be nice to be able to do the import directly from GUI. Right now, import_source cannot identify the left-/right maps because micapipe output name them XXX.L.YYY.gii; for example: "fsLR-5k.L.inflated.surf.gii" and "fsLR-5k.R.inflated.surf.gii"

would it be possible to include such naming schemes in https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/io/import_sources.m#L117-L118 ? 

cc: @adascal1

### DIFF
--- a/toolbox/math/bst_project_sources.m
+++ b/toolbox/math/bst_project_sources.m
@@ -38,7 +38,6 @@ end
 if (nargin < 3) || isempty(isAbsoluteValues)
     isAbsoluteValues = 0;
 end
-
 if (nargin < 4) || isempty(destStudyName) || ~ischar(destStudyName)
     destStudyName = '';
 end
--- a/toolbox/math/bst_project_sources.m
+++ b/toolbox/math/bst_project_sources.m
@@ -1,7 +1,7 @@
-function OutputFiles = bst_project_sources( ResultsFile, destSurfFile, isAbsoluteValues, isInteractive )
+function OutputFiles = bst_project_sources( ResultsFile, destSurfFile, isAbsoluteValues, isInteractive, iDestStudy )
 % BST_PROJECT_SOURCES: Project source files on a different surface (currents or timefreq).
 %
-% USAGE:  OutputFiles = bst_project_sources( ResultsFile, DestSurfFile, isAbsoluteValues=0, isInteractive=1 )
+% USAGE:  OutputFiles = bst_project_sources( ResultsFile, DestSurfFile, isAbsoluteValues=0, isInteractive=1, iDestStudy = [auto] )
 % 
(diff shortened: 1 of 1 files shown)


rcassani:
Get/create `destStudy` is done now in `bst_project_sources`.  Commit [24a3fa3](https://github.com/brainstorm-tools/brainstorm3/pull/934/commits/24a3fa3bbff9d384b23f17188a2bd6ce36512a2d)

```
sMaps = bst_project_sources( {OutputFile}, sCortex.FileName, 0, 0, target_condition);
```

The Subject for the targe_condition is obtained from the `destSurfFile`

The modified code now also allows to project to the same subject with a different Study name (surfaces must be different to perform the projection)

rcassani:
> would it be possible to include such naming schemes in https://github.com/brainstorm-tools/brainstorm3/blob/master/toolbox/io/import_sources.m#L117-L118 ?

This is now in master. Commit https://github.com/brainstorm-tools/brainstorm3/commit/8e4a0b898b7f9f2dc18f94e299f355000ffba473

Edouard2laire:
Great ! Thanks a lot :)

PR #935, opened 2026-08-10
Fix GUI warning on matlab2026a

jeanborn:
Fix the following warning in matlab2026a
In gui_brainstorm>@(h,ev)bst_call(@ProcessRun_Callback) (line 301) Warning: 'awtinvoke' is unsupported and might have been changed or removed without notice. With appropriate code changes, use javaMethodEDT instead.
> In awtinvoke (line 53)
In gui_show (line 151)
In db_edit_subject (line 59)
In tree_callbacks>@(h,ev)db_edit_subject (line 524)

### DIFF
--- a/toolbox/gui/gui_show.m
+++ b/toolbox/gui/gui_show.m
@@ -148,7 +148,7 @@ switch (contType)
             end
         end
         % Show window: KEEP IT WITH AWTINVOKE
-        awtinvoke(jWindow, 'setVisible(Z)', 1);
+        javaMethodEDT('setVisible', jWindow, true);
         % Returned variable
         panelContainer.type   = 'JavaWindow';
         panelContainer.handle = {jWindow};


jeanborn:
Actually, this fix doesn't really work.  Any idea on how we could fix it ?

rcassani:
It seems this calls to `awtinvoke` was overseen, as everywhere else we use `java_call()` which uses `javaMethodEDT` if it exists as a builtin function.

About the warning, I'm not able to replicate it I run 2026a on Linux and this does show.
What is the setup you get it? This info can be obtained with `bst_systeminfo()`

```
=== Matlab === 
  Matlab version : R2026a (26.01)
  Java version   : 1.8
 
=== System === 
  OS name   :      Ubuntu 22.04.5 LTS (5.15.0-179-generic)
  OS type   :      linux64
  Mem total :      32023 MiB
  Mem avail :      24510 MiB
```

If possible, report the steps to replicate it

jeanborn:
Here is the result of the command: 
```
=== Brainstorm === 
       Version :        3.260810
       Release :        260810
       Variant :        source
       Plugins :        iso2mesh brainentropy* jsonlab nirstorm* mcxlab-cuda* spm12
      
     === Brainstorm directories === 
     *** Directory paths may contain sensitive information, check before sharing *** 
       Brainstorm :     /NAS/home/jeanborn/Documents/software/brainstorm3
       DataBase   :     /NAS/home/jeanborn/brainstorm_db
       Bst_User   :     /home/jeanborn/.brainstorm
       Temporary  :     /home/jeanborn/.brainstorm/tmp
      
     === Matlab === 
       Matlab version : R2026a (26.01)
       Java version   : 1.8
      
     === System === 
       OS name   :      Ubuntu 24.04.4 LTS (6.8.0-101-generic)
       OS type   :      linux64
       Mem total :      128153 MiB
       Mem avail :      74822 MiB
```

The warning appears when I click on the "RUN" button.

rcassani:
> Actually, this fix doesn't really work. Any idea on how we could fix it ?

@jeanborn, can you confirm if the warning still shows after performing the suggested change in this PR and restart Matlab?

jeanborn:
We'll test that for you tomorrow. 
One thing we noticed before when using the proposed patch was that gui would become unresponsive when doing 'right click on a subject' > 'edit subject'. Maybe, restarting matlab was the thing to do. We'll keep you updated tomorrow.

Thanks, 
Jean-Eudes

jeanborn:
Hello, 
We have been trying to change the function as the initial commit stated and restarting matlab. The 'RUN' interface opens and lets you fill in the parameters but the 'cancel' and 'ok' button becomes unresponsive. The opened interface is also unclosable. 
Also, when trying the 'edit subject' thing, the interface opens empty, does not want to close and creates an error when clicking on 'ok' since all the fields are empty, but it still does not close. 

If that's not clear we can make a small zoom to show you. 

Regards,
Jean-Eudes

rcassani:
@jeanborn, yes we can check it in quick zoom call. Please send the invite for today

rcassani:
Thanks @jeanborn and @Edouard2laire for the detailed info! After the zoom meeting, I made some tests, and I can now reproduced the reported warning with R2026a (Update 4). Before I used R2026a pre-release which did not had the warning.

Following the warning pointed me to th real issue behind the use of `awtinvoke` in this function, and why `javaMethodEDT` "freezes" the GUI. Using `javaMethodEDT` with modal containers (JDialog in this case) blocks the Matlab thread causing the freeze. Fixing this issue requires a deeper fix on how modal dialogs are created in Brainstorm. I'll address it on a different PR in the following days and link that PR in here.

PR #936, opened 2026-08-14
Improve event display

Edouard2laire:
Hello, 

This PR is here to improve the display of the event bar; when reviewing raw files; to not attribuate space to the event that are hidden. This is quite usefull when having a lot of event present but hidden.


rcassani:
@Edouard2laire, I tested adding a small offset to the line of dots for each event as in:

`YData = repmat(.1 + (.9*(iEvt-1)/length(events) + 1/(2*length(events))) * ones(1,nOccur), size(events(iEvt).times, 1), 1);`

This is more visible for a small number of displayed events, and I think it looks better. Any comment?
 
For sake of this PR, the current display is kept

PR #938, opened 2026-08-18
[filter] replace zero padding with constant padding

Edouard2laire:
Hello, 

For a long time, our lab had an issue with a transient effect after our preprocessing pipeline on all-night data (9 hours); I finally took the time to investigate and found it was due to the padding done during filtering. See because we are having such long data; the drift is quite large from the beginning to the end of the night; and therefore; even if the mean of the signal is removed; padding the signal with zero is creating a huge discontinuity at the beginning and end of the signal before applying the filter; cause the presence of a large transient effect (at the beginning and end of the signal). 

By replacing this constant padding of 0 with a constant padding adapted to the first/last value of the signal, we avoid this effect. 

I believe this will have very little effect on other data (as the padding will be close to 0) but will help a lot in our case. 

**Before:** 

image

**After:** 

image

Filter specification: 

image

image

Note: i tried to apply a linear detrend before the filter; but it makes things even worse. So i think this PR is really the best solution :)

Thanks a lot, 
Edouard


rcassani:
To avoid the sensitivity to noise first and last sample, we could compute the constant as an average of few samples in each edge. How many samples, we can use the 99% energy transient as reference. In this example 14 samples each edge. Can you verify this on the data used in this PR?

Edouard2laire:
Something like this ? i tested on the data and it give similar results as using the first/last sample; but i agree that it should be more robust.  


As expected, this PR mainly only affect the edge: 

image


Though, the difference is not 0 during the recording.

PR #940, opened 2026-08-19
process_import_data_time: keep unit display

Edouard2laire:
This is here to keep the file unit when converting a file from raw to epoch.

### DIFF
--- a/toolbox/io/in_fread.m
+++ b/toolbox/io/in_fread.m
@@ -204,6 +204,9 @@ switch (sFile.format)
         F = in_fread_spm(sFile, SamplesBounds, iChannels);
     case 'BST-BIN'
         F = in_fread_bst(sFile, sfid, SamplesBounds, ChannelRange);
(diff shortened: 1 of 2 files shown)


rcassani:
The proposed fix is only partial as it only works for  `process_import_data_time()` process AND `BST-BIN` file format. It will not work as expected with importing time from the popup menu in the raw file, nor for raw files that are not `BST-BIN` (or `BST-DATA`, which was already supported in #449)

I prepared a solution that covers all the import approaches for all the file formats (that have `DisplayUnits`). This approach can be found in branch: https://github.com/rcassani/brainstorm3/tree/keep-display-units

Difference in:  
https://github.com/brainstorm-tools/brainstorm3/compare/master...rcassani:brainstorm3:keep-display-units?expand=1

Please check if those changes accomplish what is aimed with this PR. If so, I can push them directly into `master`.

Edouard2laire:
tested. it is working as expected. Closing this PR so you can merge https://github.com/rcassani/brainstorm3/tree/keep-display-units

rcassani:
FYI: Added in ff2780e66

PR #942, opened 2026-08-22
Bugfix: Replace removed `sun.misc` Base64 classes with `java.util.Base64`

### LINKED ISSUE #941: Anatomy import fails on Matlab R2026a: 'sun.misc.BASE64Decoder' removed in Java 9
### Summary

Brainstorm cannot import anatomy (or read/write GIFTI surfaces) on recent Matlab versions, because it uses the JDK-internal classes `sun.misc.BASE64Decoder` / `sun.misc.BASE64Encoder`, which were **removed in Java 9**.

Matlab R2026a ships with Java 11, so the affected files fail to parse as soon as they are called:

```
Line 298: Error: File: import_label.m Line: 27 Column: 8
Unable to find or import 'sun.misc.BASE64Decoder'. Imported names must end with '.*' or be fully qualified.

Call stack:
>import_anatomy_fs.m at 298
>import_anatomy.m at 78
>bst_call.m at 28
>tree_callbacks.m>@(h,ev)bst_call(@import_anatomy,iSubject,0) at 575
```

### Affected code (current `master`)

| File | Line | Usage |
| --- | --- | --- |
| `toolbox/io/import_label.m` | 27 | `import sun.misc.BASE64Decoder;` (unused, but still fatal) |
| `toolbox/io/in_tess_gii.m` | 33 | `import sun.misc.BASE64Decoder;` (unused, but still fatal) |
| `toolbox/io/in_gii.m` | 24, 59-60 | decoding of `Base64Binary` / `GZipBase64Binary` data arrays |
| `toolbox/io/out_tess_gii.m` | 64-66, 70, 75 | encoding of the vertices and faces |
| `toolbox/process/bst_report.m` | 1044, 1066 | encoding of the PNG snapshots in HTML reports |

Note that the `import` statement is resolved when the function is parsed, so even the two files that never use the decoder crash.

### Impact

- Importing any anatomy folder with FreeSurfer / CAT12 labels (`import_label`)
- Reading and writing GIFTI surfaces (`in_gii`, `in_tess_gii`, `out_tess_gii`)
- Snapshots embedded in HTML reports (`bst_report`)

### Steps to reproduce

1. Matlab R2026a (Java 11), Brainstorm at current `master`
2. Right-click a subject → *Import anatomy folder* → select a FreeSurfer folder
3. The error above appears

### Proposed fix

Replace the `sun.misc` classes with `java.util.Base64` (available since Java 8), wrapped in a small helper `bst_base64('encode'/'decode', ...)` in `toolbox/misc`, so the change is made in a single place. The helper can fall back to `matlab.net.base64decode/encode` (Matlab >= R2016b) and then to the old `sun.misc` classes, so older installations are unaffected.

I have this working locally and would be happy to open a PR if the approach sounds good to you.

### Environment

- Matlab: 26.1.0.3312084 (R2026a) Update 4, Java 11.0.32
- OS: macOS (Darwin 25.2.0)
- Brainstorm: `master` @ 3aad87f7

bdthombre:
Fixes #941

### Problem

`sun.misc.BASE64Decoder` and `sun.misc.BASE64Encoder` are JDK-internal classes that were **removed in Java 9**. Matlab versions shipping a newer JRE (R2026a uses Java 11) cannot even parse the files that import them:

```
Line 298: Error: File: import_label.m Line: 27 Column: 8
Unable to find or import 'sun.misc.BASE64Decoder'. Imported names must end with '.*' or be fully qualified.

Call stack:
>import_anatomy_fs.m at 298
>import_anatomy.m at 78
>bst_call.m at 28
>tree_callbacks.m>@(h,ev)bst_call(@import_anatomy,iSubject,0) at 575
```

Because an `import` is resolved when the function is parsed, even the two files that never use the decoder crash. This breaks importing any anatomy with FreeSurfer/CAT12 labels, reading and writing GIFTI surfaces, and the snapshots in HTML reports.

### Changes

Added `bst_base64()` in `toolbox/misc`, so the Java dependency lives in a single place:

```matlab
bytes = bst_base64('decode', str);    % str  : char array (line breaks allowed)
  str = bst_base64('encode', bytes);  % bytes: int8/uint8 vector (or Java byte[])
```

It uses `java.util.Base64` (Java >= 8), and falls back to `matlab.net.base64decode/encode` (Matlab >= R2016b) and then to the old `sun.misc` classes, so older installations keep working unchanged.

To keep the behaviour identical to the previous code, the helper:
- tolerates line breaks and white spaces in the encoded string (as `decodeBuffer` did),
- returns `int8` when decoding (as `decodeBuffer` did), so the existing `typecast()` and `dunzip()` calls are unchanged,
- strips the line breaks when encoding (this was done by the callers before).

Updated callers:

| File | Change |
| --- | --- |
| `toolbox/io/import_label.m` | removed the unused `import` |
| `toolbox/io/in_tess_gii.m` | removed the unused `import` |
| `toolbox/io/in_gii.m` | decoding of `Base64Binary` / `GZipBase64Binary` data arrays |
| `toolbox/io/out_tess_gii.m` | encoding of the vertices and faces |
| `toolbox/process/bst_report.m` | encoding of the PNG snapshots |

### Testing

Matlab R2026a Update 4 (Java 11.0.32), macOS:

- Base64 round-trip on 5000 random bytes, on `float32` arrays, on known test vectors (`Brainstorm` <-> `QnJhaW5zdG9ybQ==`), and on CRLF-wrapped input: all exact
- `in_gii` / `in_tess_gii` on a `GZipBase64Binary` surface (`external/ImaGIN/private/SEEG_cortex_hip_amy_8196.surf.gii`): 7861 vertices, 15359 faces decoded correctly
- Full `in_tess_gii` -> `out_tess_gii` -> `in_tess_gii` round-trip: faces identical, vertex max difference 0
- FreeSurfer anatomy import through the GUI, which was the original failure, now completes

Happy to adjust the naming or the location of the helper if you prefer something else.

### DIFF
--- a/toolbox/io/import_label.m
+++ b/toolbox/io/import_label.m
@@ -24,8 +24,6 @@ function [sAllAtlas, Messages] = import_label(SurfaceFile, LabelFiles, isNewAtla
 %
 % Authors: Francois Tadel, 2012-2022
 
-import sun.misc.BASE64Decoder;
-
 
 %% ===== GET FILES =====
 sAllAtlas = repmat(db_template('Atlas'), 0);
--- a/toolbox/io/in_gii.m
+++ b/toolbox/io/in_gii.m
@@ -21,8 +21,6 @@ function [sXml, Values] = in_gii(GiiFile)
 %
 % Authors: Francois Tadel, 2017
 
-import sun.misc.BASE64Decoder;
-
 % Read XML file
 sXml = in_xml(GiiFile);
 
@@ -56,8 +54,7 @@ for iArray = 1:nArrays
             
         case {'Base64Binary', 'GZipBase64Binary'}
             % Base64 decoding
-            decoder = BASE64Decoder();
-            Values{iArray} = decoder.decodeBuffer(sXml.GIFTI.DataArray(iArray).Data.text);
+            Values{iArray} = bst_base64('decode', sXml.GIFTI.DataArray(iArray).Data.text);
             % Unpack gzipped stream
             if strcmpi(sXml.GIFTI.DataArray(iArray).Encoding, 'GZipBase64Binary')
                 Values{iArray} = dunzip(Values{iArray});
--- a/toolbox/io/in_tess_gii.m
+++ b/toolbox/io/in_tess_gii.m
@@ -30,8 +30,6 @@ function TessMat = in_tess_gii(TessFile)
 %
 % Authors: Francois Tadel, 2012-2017
 
-import sun.misc.BASE64Decoder;
-
 % Read GII file
 [sXml, Values] = in_gii(TessFile);
 % Initialize matrices
--- a/toolbox/io/out_tess_gii.m
+++ b/toolbox/io/out_tess_gii.m
@@ -61,19 +61,14 @@ if isSPM && ~isempty(TessFile)
 end
 
 % ===== ENCODE DATA =====
-import sun.misc.BASE64Encoder;
-% Initialize Base64 encoder
-encoder = BASE64Encoder();
 % Encode vertices: millimeters
 Vertices = single(TessMat.Vertices' .* 1000);
 Vertices = typecast(Vertices(:), 'uint8');
-Vertices = char(encoder.encodeBuffer(Vertices));
-Vertices((Vertices == 10) | (Vertices == 13)) = [];
+Vertices = bst_base64('encode', Vertices);
 % Encode faces: 0-based array
 Faces = int32(TessMat.Faces' - 1);
 Faces = typecast(Faces(:), 'uint8');
-Faces = char(encoder.encodeBuffer(Faces));
-Faces((Faces == 10) | (Faces == 13)) = [];
+Faces = bst_base64('encode', Faces);
  
 % ===== CREATE XML STRING =====   
 % Create XML file from a template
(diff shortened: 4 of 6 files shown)


rcassani:
@bdthombre thank you for your contribution!

This is issue has been already reported in the forum:

- https://neuroimage.usc.edu/forums/t/49189
- https://neuroimage.usc.edu/forums/t/46084
- https://neuroimage.usc.edu/forums/t/56044

In those cases we pointed the user to install JDK8 and set Matlab to use it.
This was solution was given because Brainstorm supports versions of Matlab as old as 10 years, and  JDK8 is the version that is supported for all of these (including R2026a): https://www.mathworks.com/support/requirements/openjdk.html

The proposed wrapper in this PR can help with this, so the user will not need to manually install and set JDK8. Indeed, I made a proof-of-concept that was not merged: https://github.com/rcassani/brainstorm3/commit/e5dfb01dc4c267ca6622bafc3a68874bb0271450

We will thoroughly test the proposed changes, this may take a bit because there are multiple places where Brainstorm uses the JDK. Moreover, Brainstorm runs in multiple versions of Matlab, different OS and CPU architectures. In the mean while you can setup JDK8 manually. 

---

I think this is an issue only for Matlab macOS, and not only for 2026a, but earlier versions (as per the forum posts). AFAIK in Linux and Windows, Matlab is still shipped with JKD8.

bdthombre:
Thanks. Please do what is best for your project, locally this solution is working without any issues so far since then for me, just to keep you informed. If any issues comes I'll post.

PR #945, opened 2026-08-31
Add process to merge adjacent event into a continious event

Edouard2laire:
As discussed in the forum, this process is here to merge consecutive adjacent events into larger continuous events. 

Additionnaly: 
- Allows users to remove parts of the events affected by artifacts. 
- Allows filtering shorter epochs (events with less than a specified duration)

Here is the GUI: 

image

Do you want me to put the other process to classify events in this PR or submit another one?

### DIFF
--- a/toolbox/process/functions/process_evt_merge_continious.m
+++ b/toolbox/process/functions/process_evt_merge_continious.m
@@ -1,0 +1,201 @@
+function varargout = process_evt_merge_continious( varargin )
+% process_evt_merge_continious:  Merge continious, uninterrupted events
+% into one event. 
+%
+% USAGE:  OutputFiles = process_evt_merge_continious('Run', sProcess, sInputs)
+
+% @=============================================================================
+% This function is part of the Brainstorm software:
+% https://neuroimage.usc.edu/brainstorm
+% 
+% Copyright (c) University of Southern California & McGill University
+% This software is distributed under the terms of the GNU General Public License
+% as published by the Free Software Foundation. Further details on the GPLv3
+% license can be found at http://www.gnu.org/copyleft/gpl.html.
+% 
+% FOR RESEARCH PURPOSES ONLY. THE SOFTWARE IS PROVIDED "AS IS," AND THE
+% UNIVERSITY OF SOUTHERN CALIFORNIA AND ITS COLLABORATORS DO NOT MAKE ANY
+% WARRANTY, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO WARRANTIES OF
+% MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE, NOR DO THEY ASSUME ANY
+% LIABILITY OR RESPONSIBILITY FOR THE USE OF THIS SOFTWARE.
+%
+% For more information type "brainstorm license" at command prompt.
+% =============================================================================@
+%
+% Authors: Edouard Delaire, 2026
+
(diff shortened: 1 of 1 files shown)


rcassani:
Context: https://neuroimage.usc.edu/forums/t/merge-and-combine-events/57939

Which of the two is being solved with this PR?

Edouard2laire:
The first one: 


> Is there a way to merge consecutive events ? Sleep stage in my data is marked in epochs of 30s. So I have events at 30s, 60s, 90s … all with a duration of 30s. Is there a way to generate an event, in this case starting at 30s, lasting 90s (3 epochs) ?



I am currently cleaning the code I have for the second one, Let me know if you want an example file

rcassani:
In the forum, I provided a step-by-step solution for this with the current event processes

If needed, I can add a modification in the current "Merge overlapping extended events" to remove the one-sample issue, e.g., a checkbox to allow merging consecutive, 0-sample gap, extended events

For changes like this, discussing on the forum first lets us agree on the approach before any code

Edouard2laire:
> In the forum, I provided a step-by-step solution for this with the current event processes

The proposed steps are quite complicated, whereas this is only one process which also ensures that there is no bad segment in the continuous event and ensures a minimum duration for the segment. 

The idea is that the user (me in this case) would then segment the file based on the continuous event to then analyse each segment individually. Kind og like for MEG we do 'generate stable head'; but here, with a stable period for each sleep stage. 


> If needed, I can add a modification in the current "Merge overlapping extended events" to remove the one-sample issue, e.g., a checkbox to allow merging consecutive, 0-sample gap, extended events

Could be nice, but is it possible to remove bad period / only merge if the created events is longuer than a desired duration ? 



> For changes like this, discussing on the forum first lets us agree on the approach before any code

Yes i agree. Here i went directly to code as the code was already written from a long time ago; i just cleanned it. If it's too specific for brainstorm; i can always move it to a custom plugin now that it's clean  :) 


The cleaning of the other process involves very similar code so it was nice to start by this one :)

Regars, 
Edouard

rcassani:
The steps I proposed looked complicated, as "duration" was mentioned I assumed the events were already extended. But as they are simple events, the steps can be summarized as:

1. Convert them to extended events with the desired duration: 0 s before, and 30 s after
2. Merge overlapping events (with the new "allow adjacent extended events" option db3731e)

About editing events based on bad events, that is beyond the merging overlapping occurrences of extended event. In the forum, I mention that a process similar to Combine stim/response, but for extended events would be a nice addition. In that process it could compute the intersection (and difference) of extended events, so reaching the function of what is aimed. I.e., given a pair of extended events, A and B, Compute event AnB (intersection of A and B), A-B (A but not B) and B-A (B but not A).

Closing this PR, as the functionality is now in Brainstorm (Commit db3731e)

Edouard2laire:
> The steps I proposed looked complicated, as "duration" was mentioned I assumed the events were already extended. But as they are simple events, the steps can be summarized as:
> 
> Convert them to extended events with the desired duration: 0 s before, and 30 s after
> Merge overlapping events (with the new "allow adjacent extended events" option https://github.com/brainstorm-tools/brainstorm3/commit/db3731e59c3e5481ac45f653c7d1d251f36bd3a7)


Yes, they are extended. But I guess this should work. (all event are 30s). So I just need to do step 2 that you suggested. 

Would there be then a way to do a "difference" of event; so to remove period that are marked as "bad_motion" for example, from other events ?

PR #946, opened 2026-09-07
Update process_ft_sourceanalysis_dics.m

vyoussofzadeh:
This update improves the FieldTrip-based DICS beamformer process by adding user-facing options for source orientation and regularization.

Users can now choose between maximum-power orientation using FieldTrip fixedori and head-model/cortical-normal orientation using Brainstorm GridOrient. The selected orientation, regularization value, ERD/ERS mode, effect type, FOI, taper smoothing, and time windows are saved in the result metadata and shown in the result comment.

The update also simplifies result registration using db_add_data and excludes MEG reference channels to avoid possible channel mismatch issues.

### DIFF
--- a/toolbox/process/functions/process_ft_sourceanalysis_dics.m
+++ b/toolbox/process/functions/process_ft_sourceanalysis_dics.m
@@ -28,7 +28,7 @@ end
 %% ===== GET DESCRIPTION =====
 function sProcess = GetDescription() 
     % Description the process
-    sProcess.Comment     = 'FieldTrip: DICS beamformer - clean';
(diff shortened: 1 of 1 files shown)


vyoussofzadeh (commented):
Typo corrected

vyoussofzadeh:
I also have a Brainstorm-native version of the DICS beamformer that does not depend on FieldTrip. I did not include it in this pull request to keep the update focused. If useful, I can submit it separately, or we could consider integrating it later as an optional backend under the same DICS process to avoid having two confusing user-facing DICS tools.

PR #947, opened 2026-09-13
Help with CTF head points: warn and offer MEG alignment, auto-select best .pos (#944)

### LINKED ISSUE #944: CTF coordinate system & adding headpoints
There's a small issue when dealing with CTF files (perhaps others) and importing head points from a .pos file (perhaps other formats), outside the common case where a single head points file is loaded automatically with the raw data initially.  This happens for example in these cases:
1. There are two (or more) .pos files, e.g. one for EEG electrodes and one for head points (recorded before putting the EEG cap on).
2. There is an issue in the .pos file, and the points need to be removed and re-imported after fixing it, e.g. left-right coils were swapped, or the units were inches instead of cm (which can happen if the system is not initialized correctly).

In case 1, the behavior is to warn the user and not import either of the pos files initially.  This means there are no digitized anatomical points to create the SCS coordinates, and in this case the MEG side stays in "native" coordinates, defined by the head coils.  The "Native=>Brainstorm/CTF" transformation is missing.
In case 2, a "Native=>Brainstorm/CTF" transformation is created when first importing the data, from the "bad" pos file.  This transformation can be removed by resetting the channel file with process_adjust_coordinates (in the pipeline editor: Import > Channel File > Adjust Coordinate System).

If a user then manually adds head points through the GUI, the user is offered to either apply a transformation from the MRI to SCS coordinates (which in this case would be wrong), or to use the digitized anatomical points it contains (which is right).  However, the MEG coordinate system is not updated, and "Native=>Brainstorm/CTF" is still missing.  This results in the MEG head coils being aligned with the digitized anatomical points instead of the digitized head coils.  

There is currently no option to fix this, other than reimporting that data with a single/fixed .pos file present next to it.

Proposed actions:
1. Minimal: add a warning when adding head points that the current MEG coordinate system will not be changed based on the digitized fiducials, and likely wrong.  Suggest to rather import the data again with a single .pos file next to it.
2. Check the current MEG transformations and fiducials present in the pos file.  If the MEG is in native coordinates and coils and anatomical points are present, offer to adjust the MEG coordinate system to SCS.  If the MEG is already in SCS, and the digitized fiducials are present, warn that the new points may define a different SCS for the MEG (perhaps check if that's the case), and offer to keep the previous SCS coordinates on the MEG side, or use the new one (likely the correct choice).
3. Maybe when there are multiple .pos files, warn and prompt the user to select one to import, or if running non-interactively (e.g. importing a large BIDS dataset like OMEGA), have selection criteria to pick one.  For the default choice, we can look for one that has all digitized fiducials and the most head points.  That's a little arbitrary, but could still save at worse 50% of cases to have to redo it manually.  Whatever is decided, having a clear warning of what was done would be important.

I wrote all this here because I don't think I'll have time to work on this myself soon. But I think the minimal change 1 above is important to add.

sonoAESS:
Fixes https://github.com/brainstorm-tools/brainstorm3/issues/944

## Summary
- Add head points: warn when the MEG is in "Native" coordinates without the
  "Native=>Brainstorm/CTF" transformation and offer to align the MEG sensors in
  SCS using the digitized anatomical fiducials (NAS/LPA/RPA). Aligning is only
  proposed when it is safe (MEG still in Native); when the MEG is already in SCS
  we only warn, to avoid applying the transformation twice.
- CTF import: when several .pos files are present in/next to a .ds folder, no
  longer abort with "Two Polhemus files in the same folder." Instead the best
  candidate is selected deterministically by score (measured fiducials > head
  coils > digitized points), confirmed interactively in the GUI.

## Changes
- `toolbox/process/functions/process_headpoints_add.m`: warning + SCS alignment
  offer (interactive only), new `GetMegAlignWarning` / `GetAnatomicalFiducials` /
  `AlignMegToScs` helpers, `Run()` reports the message as a warning.
- `toolbox/io/select_pos_file.m` (new): deterministic .pos selection/scoring.
- `toolbox/io/ctf_get_files.m`: delegate multi-.pos selection to `select_pos_file`.
- `toolbox/script/test_headpoints_ctf.m` (new): automated tests
  (GetMegAlignWarning matrix, ScorePosFile, SelectPosFile, AlignMegToScs).

## Validation
- Automated tests pass locally and in headless/server mode.
- Manually validated with the CTF sample phantom dataset (ds with 2 .pos files,
  ds without .pos, ds already aligned in SCS).

## AI usage disclosure
This pull request was developed with the assistance of an AI coding assistant
(OpenCode). All changes were reviewed and understood by the author before submission.


toolbox/io/ctf_get_files.m
```diff
@@ -131,7 +131,16 @@
 if (length(posDir) == 1)
     pos_file = bst_fullfile(ds_directory, posDir(1).name);
 elseif (length(posDir) > 1)
-    error('Two Polhemus files in the same folder.');
+    % Multiple Polhemus files in the same folder: select the one with all the anatomical
+    % fiducials and the most digitized points (see select_pos_file for details).
+    % In an interactive session (Brainstorm GUI running), ask the user to confirm the automatic
+    % selection; in batch/headless mode the selection is automatic and deterministic.
+    global GlobalData;
+    isInteractive = (verbose >= 1) && ~isempty(GlobalData) && isstruct(GlobalData) && ...
```
Moo-Marc: We might want this option to be a function input. Sometimes we import a large dataset using the GUI, and we still don't want the process to stop midway with a prompt.  E.g. from process_import_bids (OPTIONS.isInteractive) > in_fopen (ImportOptions.DisplayMessages) > ...  This is not currently further propagated to in_fopen_ctf and here in ctf_get_files, but it is for other file formats and could be added for CTF too (in a backwards compatible way).

toolbox/io/ctf_get_files.m
```diff
@@ -131,7 +131,16 @@
 if (length(posDir) == 1)
     pos_file = bst_fullfile(ds_directory, posDir(1).name);
 elseif (length(posDir) > 1)
```
Moo-Marc, line 143: This modification should apply to the other cases below as well, where the pos file is not "inside" the .ds directory.  This is actually the standard now with BIDS, the pos file should be next to the .ds folders, and named sub-subid_ses-sessid_headshape.ds.  So this function is a little outdated and pos files will be found under "attempt 3" below.  So this should be done after this if clause (line 167).

toolbox/io/select_pos_file.m
```diff
@@ -0,0 +1,125 @@
...
+    % Report the selection
+    if verbose
+        disp(['CTF> Multiple .pos files found, selected: ' pos_file ' (score: ' num2str(bestScore) ')']);
+        disp('BST> Warning: Please verify that the selected .pos file is the correct one.');
+    end
+end
+
+
+%% ===== SCORE POS FILE =====
+function score = ScorePosFile(ChannelMat)
```
Moo-Marc, line 118: This scoring is not ideal I think.  A meg coil is not "equivalent" to 10 head points.  And we don't necessarily want to score higher when there are more repeated measures of the fiducials.  In practice, the most common case and the only one where we'd trust an automatic selection, is when there's one file for electrodes and one for head points.  Any other case is when something unexpected happened and a human will have to investigate to know which file to use for each recording.  Still, I think the desired logic here is: look for file(s) with all 3 types: anat fids, meg coils, and head points, ignoring electrodes. Otherwise, if some types are missing, there isn't really a "best" way to select, but I'd prioritize meg coils + points, then anat fids + points, then coils + anat without any points.  (If a file has only one type out of these 3, it can't be used.)  Only if there are multiple files with the same "best" combination of types of points should we pick the one with the highest number of head points among them.

toolbox/process/functions/process_headpoints_add.m
```diff
@@ -182,16 +190,138 @@
         ChannelMat.HeadPoints = HeadPoints;
     end
 
+    % ===== CHECK MEG COORDINATE SYSTEM =====
+    % If the MEG sensors are not aligned in SCS, and the imported head points contain anatomical
+    % fiducials (NAS/LPA/RPA), warn the user and offer to update the MEG coordinate system.
+    % This only applies when the file was selected manually (interactive session), not when the
+    % process is running as part of a pipeline.
+    [strWarn, isOfferAlign] = process_headpoints_add('GetMegAlignWarning', ChannelMat, HeadPoints);
```
Moo-Marc: Local functions should be called locally:
[strWarn, isOfferAlign] = GetMegAlignWarning(ChannelMat, HeadPoints);
and lower:
ChannelMat = AlignMegToScs(ChannelMat, HeadPoints);

toolbox/process/functions/process_headpoints_add.m
```diff
@@ -182,16 +190,138 @@
...
+        return;
+    end
+    % Does the MEG have the "Native=>Brainstorm/CTF" transformation?
+    if isfield(ChannelMat, 'TransfMegLabels') && iscell(ChannelMat.TransfMegLabels) && ~isempty(ChannelMat.TransfMegLabels)
+        hasCtfTransf = ismember('Native=>Brainstorm/CTF', ChannelMat.TransfMegLabels);
+    else
+        hasCtfTransf = 0;
+    end
+    % Detect the anatomical fiducials (NAS/LPA/RPA) in the imported head points
+    [iNas, iLpa, iRpa] = GetAnatomicalFiducials(HeadPoints);
```
Moo-Marc, line 253: It's not quite as simple as this. Normally, we expect both anatomical fiducials which Brainstorm uses to define its SCS, and MEG head coils which define the native coordinate system of the data files.  Both are required to generate the transformation usually.  However, we have older data where we only had one set which were the head coils. These were named as the anatomical fiducials are now, or as the new standard "hpi-n/l/r" labels.  Either way, when only one set is present we assume they are head coils because these are required to do anything useful with MEG data.  In that case, the user must use the head coil positions on the MRI as well and the SCS will actually correspond to the native coordinate system.  see in_channel_pos and in_fopen_ctf for how this is dealt with.
So this logic should be compatible with checks here as well, and not rely only on anatomical fiducials.

Moo-Marc (commented):
Thanks for your contribution.  Here are some comments that require modifications.  (I did not have time to check the testing script.)

toolbox/io/ctf_get_files.m
sonoAESS: Done in ffa1df32. `ctf_get_files(ds_directory, verbose, isInteractive)` now takes an optional third input... when omitted it falls back to the previous GUI-state heuristic, so existing callers keep working. `in_fopen` passes `ImportOptions` to `in_fopen_ctf`, which forwards `ImportOptions.DisplayMessages`. The BIDS chain already mapped `OPTIONS.isInteractive` to `DisplayMessages (process_import_bids.m:639)`, so it now reaches `ctf_get_files`: batch imports with `isInteractive=0` never stop on a prompt. `in_channel_ctf` (no ImportOptions) keeps the heuristic.

toolbox/io/ctf_get_files.m
sonoAESS: Done in 10c124ee. Attempts #1–#3 now accumulate all candidate .pos files (attempt #2 required exactly one match before, attempt #3 with ≥2 only warned and imported nothing), and the single selection step runs after the if/else for every origin...covering `sub-*_headshape.pos` + `sub-*_electrodes.pos` next to the `.ds`. Plus bff01cb6: attempt #2 only runs when the `*_meg.ds` to `*_*.pos` pattern actually changed, so a legacy `.ds` without `_meg.ds` doesn't list its whole folder as candidates.

toolbox/process/functions/process_headpoints_add.m
sonoAESS: Done in 806a168f...

toolbox/io/select_pos_file.m
sonoAESS: Rewritten in 2a2705f6 as suggested: `score = combination × 1e6 + nHeadPoints`, with combination 4 = fiducials+coils+points, 3 = coils+points, 2 = fiducials+points, 1 = fiducials+coils, and -1 for a single type (unusable). Ties break only on head-point count within the same combination; electrodes stay ignored (read as EEG channels). Non-interactive with all candidates at -1: nothing imported + warning... interactive: dialog still shown, first file preselected with no recommendation.

toolbox/process/functions/process_headpoints_add.m
sonoAESS: Done in 64c651b2. `GetAnatomicalFiducials` returns indices only when both complete sets are present (NAS/LPA/RPA + HPI-N/L/R)... a single set is treated as head coils, matching `in_channel_pos / in_fopen_ctf`. No anatomical alignment is offered or applied in that case, and the warning message explains it.

sonoAESS:
The test script was updated for all of the above: new scoring expectations (2a–2h), both-set fixtures, single-set cases 1f/1g/4i/4j, and a fixture layout fix (85820f4b). All 4 blocks pass headless, plus an end-to-end scenario check (auto-selection, offer/no-offer cases).

